PR #178, opened 2017-04-19
Updated QR code for adding users. Fixed bug when updating user data.

Elland:


Token/Models/User.swift
```diff
@@ -57,10 +57,14 @@ public class User: NSObject, JSONDataSerialization {
 
     private var _avatar: UIImage?
 
-    private(set) public var username: String
+    public var username: String {
+        return "@\(self.name)"
```
faymontage: @Elland "@" is not part of the username at the model level, it is just a rendering/display stylisation

Elland: This var is for UI/QR code generation only. Instead of having this code replicated in every view. Should probably rename this stuff better now.

faymontage: You don't get this state?

img_3178
img_3179

Elland: Fixing that as well!

Elland: Naming got confusing, stuff passed me by. Should be clearer now.

Elland:
It's up.

PR #179, opened 2017-04-19
Fixes to Signal signout. 

Elland:
Some code cleanup for apps.


Elland:
@faymontage Try reproducing the ghost user/notifications now. 👍

PR #183, opened 2017-04-20
Fixed missing title for messages view when recipient is not a contact.

Elland:


Elland:
@roberthein let me know if there are any issues here when you have the chance to look at it.

PR #185, opened 2017-04-20
Update app structure and layout. Fix visual glitches. Restore auto-correct on send.

Elland:


Elland:
Will probably wait to merge this once we get the new assets and finish the whole redesign.

Cartfile
```diff
@@ -9,7 +9,7 @@ github "marketplacer/keychain-swift" ~> 7.0
 github "UseSweet/SweetFoundation" ~> 1.2.0
 github "UseSweet/SweetSwift" "master"
 github "tokenbrowser/AwesomeCache" "master"
-github "hyperoslo/ImagePicker" ~> 2.0
-github "UseSweet/SweetUIKit" ~> 1.0
```
3lvis: It's better if we point to the major versions instead of master or nothing, in that way we avoid breaking the app on dependency updates. So I suggest we keep this as it was.

```
github "hyperoslo/ImagePicker" ~> 2.0
 github "UseSweet/SweetUIKit" ~> 1.0
```

Elland: Than we should add new tags, because latest tag is outdated. Or we can point to commit hash.

3lvis: Are you talking about `ImagePicker `? SweetUIKit is up to date. Running `carthage update` here will update any 1.X.X versions https://github.com/UseSweet/SweetUIKit/releases/tag/1.12.1

Elland: I'll look into it. 👌

Elland:
Considering just merging this and we fix anything that I might have missed as we go. Too big a change. Ideas?

Token/Extensions/TSThread+Additions.h
```diff
@@ -20,7 +20,7 @@
 
 @interface TSThread (Additions)
 
-@property (nonatomic, readonly) NSArray<TSIncomingMessage *> *visibleIncomingInteractions;
+@property (nonatomic, readonly) NSArray<TSMessage *> *messages;
```
yuliaveres: don't we need nullable specifiers ?

Token/Models/ContactsManager.h
```diff
@@ -34,9 +34,9 @@
 
 - (NSArray<Contact *> * _Nonnull)signalContacts;
 
-- (NSArray<TokenContact *> * _Nonnull)tokenContacts;
+- (NSArray<TokenUser *> * _Nonnull)tokenContacts;
```
yuliaveres: also, i know it is not related to this change, but let's decide on a way we indicate nullability,  
cause as for me`- (nonnull NSArray<TokenUser *> *)tokenContacts` looks better

Token/Controllers/Contacts/FavoritesNavigationController.swift
```diff
@@ -28,7 +28,8 @@ public class ContactsNavigationController: UINavigationController {
     public override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
         super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
 
-        self.tabBarItem = UITabBarItem(title: "Contacts", image: #imageLiteral(resourceName: "contacts"), tag: 1)
+        self.tabBarItem = UITabBarItem(title: "Favorites", image: #imageLiteral(resourceName: "favourites"), tag: 1)
```
Elland: They're not optional, right?

Token/Controllers/Contacts/ContactsController.swift
```diff
@@ -88,7 +73,7 @@ open class ContactsController: SweetTableController {
             self.mappings.update(with: transaction)
         }
 
-        self.title = "Contacts"
+        self.title = "Favorites"
```
Elland: @yuliaveres we're not localising yet, but might be worth starting now.

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    open override func viewDidLoad() {
+        super.viewDidLoad()
+
+        self.collectionView.showsVerticalScrollIndicator = true
+        self.collectionView.alwaysBounceVertical = true
+        self.collectionView.backgroundColor = Theme.viewBackgroundColor
+
+        self.searchBar.delegate = self
```
Elland: Aren't they? 💭

Token/Controllers/Contacts/ContactsController.swift
roberthein: 👍

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -0,0 +1,143 @@
...
+        let searchField = self.searchBar.value(forKey: "searchField") as? UITextField
+        searchField?.backgroundColor = Theme.inputFieldBackgroundColor
+
+        self.collectionView.register(AppCell.self)
+
+        self.title = "Explore"
+        self.appsAPIClient.getFeaturedApps { apps, error in
+            if let error = error {
+                let alertController = UIAlertController.errorAlert(error as NSError)
+                self.present(alertController, animated: true, completion: nil)
```
Elland: Always. 🚀

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            self.recommendedApps = apps
+        }
+    }
+}
+
+extension BrowseController {
```
Elland: We're inheriting from `SearchableCollectionController`, it's already its own datasource/delegate. Adding that back here would give us a "protocol already declared" error.

Token/Controllers/Messaging/ChatsFloatingHeaderView.swift
```diff
@@ -16,13 +16,13 @@
...
-    func messagesFloatingView(_ messagesFloatingView: MessagesFloatingView, didPressRequestButton button: UIButton)
-    func messagesFloatingView(_ messagesFloatingView: MessagesFloatingView, didPressPayButton button: UIButton)
+protocol ChatsFloatingHeaderViewDelegate: class {
+    func messagesFloatingView(_ messagesFloatingView: ChatsFloatingHeaderView, didPressRequestButton button: UIButton)
+    func messagesFloatingView(_ messagesFloatingView: ChatsFloatingHeaderView, didPressPayButton button: UIButton)
 }
 
-class MessagesFloatingView: UIView {
-    weak var delegate: MessagesFloatingViewDelegate?
+class ChatsFloatingHeaderView: UIView {
```
Elland: Are there any advantages? Only I could find was performance, might it be worth measuring?

Token/Extensions/TSThread+Additions.h
Elland: I thought not, since they all default to nullable.

Token/Models/ContactsManager.h
Elland: I'm fine with either. ObjC kept changing that, I lost track of the difference between them. Are there any?

Elland: Meaning: is that no longer applicable? http://stackoverflow.com/a/33682230

PR #196, opened 2017-04-25
Refactor and extract common logic from User and TokenContact

yuliaveres:
Got rid of numerous duplications inside User and TokenContact models, 
considered subclassing from common class instead of conforming to protocol due to several limitations.


Token/Models/TokenContact.swift
```diff
@@ -33,70 +33,40 @@ public class TokenContact: NSObject, JSONDataSerialization, NSCoding {
...
-
-    public var JSONData: Data {
+    
+    override public var asDict: [String: Any] {
         var imageDataString = ""
         if let image = self.avatar, let data = (UIImagePNGRepresentation(image) ?? UIImageJPEGRepresentation(image, 1.0)) {
             imageDataString = data.hexadecimalString
         }
-
+
```
3lvis: Make sure to use SwiftFormat, will clean all of these things for your plus will make sure that we keep the project with a consistent style.

https://github.com/bakkenbaeck/iOS-playbook/blob/master/BEST_PRACTICES.md#linting

Token/APIClients/idAPIClient.swift
```diff
@@ -147,8 +147,8 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
                 case .success(let json, _):
                     guard let userDict = json?.dictionary else { completion(false); return }
 
-                    User.current?.update(avatar: avatar, avatarPath: userDict["avatar"] as! String)
-
+                    TokenUser.current?.update(avatar: avatar, avatarPath: userDict["avatar"] as! String)
+
```
Elland: oops :)

Token/Controllers/Contacts/FavoritesController.swift
```diff
@@ -104,7 +104,7 @@ open class FavoritesController: SweetTableController {
         if let address = UserDefaults.standard.string(forKey: self.selectedContactKey) {
             // This doesn't restore a contact if they are not our contact, but a search result
             DispatchQueue.main.asyncAfter(seconds: 0.0) {
-                guard let contact = self.contact(withAddress: address) else { return }
+                guard let contact = self.contact(with: address) else { return }
```
Elland: 👏

3lvis:
👏

Token/Controllers/Contacts/FavoritesController.swift
3lvis: 💥

Token/Models/TokenUser.swift
```diff
@@ -0,0 +1,224 @@
+//
+//  TokenUser.swift
+//  Token
+//
+//  Created by Yuliia Veresklia on 24/04/2017.
+//  Copyright © 2017 Bakken&Bæck. All rights reserved.
```
3lvis: Could you delete the attribution header?

Token/Views/ContactCell.swift
```diff
@@ -30,11 +30,6 @@ class ContactCell: UITableViewCell {
 
                 if let avatar = contact.avatar {
                     self.avatarImageView.image = avatar
-                } else {
-                    IDAPIClient.shared.downloadAvatar(path: contact.avatarPath) { image in
```
3lvis: What's the reason for removing this piece of code?

3lvis: Nevermind, had a second look.

Token/Models/TokenUser.swift
yuliaveres: oh sorry i need to get used to this :/

Elland: Not delete, but add the one we use for this project.

Elland: swiftformat will delete it for you, normally. But for this project we use one.

3lvis: I guess is this one refactor/user-and-token_contact-common-logic

yuliaveres: done

3lvis:
This looks good to me.

Elland:
Same here. Gonna run it now and look for buggie bugs! 🐛

Elland:
I think we can delete `User.swift`, right?

yuliaveres:
isn't it ? :O

Elland:
No, still used some places too. Wanna remove it all? 🚀

yuliaveres:
can't find any occurrence of old User, i was removing all of them

Token/Models/TokenUser.swift
```diff
@@ -0,0 +1,231 @@
...
+            }
+            
+            return [
+                Constants.Keys.address        : address,
+                Constants.Keys.paymentAddress : paymentAddress,
+                Constants.Keys.username       : username,
+                Constants.Keys.about          : about,
+                Constants.Keys.location       : location,
+                Constants.Keys.name           : name,
+                Constants.Keys.avatarDataHex  : avatarPath,
```
Elland: Should probably be `keys.avatar`. Duplicated key causes a crash. Also missing all the `self`s. 😄 Also, we don't indent inside lines. Checking the guidelines to be sure it's there now. 👍

Token/Models/TokenUser.swift
```diff
@@ -0,0 +1,231 @@
...
+}
+
+extension Notification.Name {
+    public static let CurrentUserDidUpdateAvatarNotification = Notification.Name(rawValue: "CurrentUserDidUpdateAvatarNotification")
+    public static let TokenContactDidUpdateAvatarNotification = Notification.Name(rawValue: "TokenContactDidUpdateAvatarNotification")
+}
+
+public class TokenUser: NSObject,  JSONDataSerialization, NSCoding {
+    
+    struct Constants {
```
Elland: Hmm, what is the reason for a structure inside a structure?

Token/Models/TokenUser.swift
```diff
@@ -0,0 +1,231 @@
...
+                    self.avatar = image
+                }
+            } else {
+                IDAPIClient.shared.downloadAvatar(path: self.avatarPath) { image in
+                    self.avatar = image
+                }
+            }
+        }
+        
+        if isApp {
```
Elland: This doesn't look right to me. I think we need those notifications for both apps and contacts, right? I think they're only different notifications between contacts/apps (app is a subset of contact) and the current user.

yuliaveres: okay

Token/Models/TokenUser.swift
yuliaveres: Used to separate different constants types, string keys, or numeric values etc.

Elland: But we don't have any, do we? They're all string. A bit early optimisation maybe? 😄

yuliaveres: i can remove nested one

Elland: Just realised, is there a reason why this isn't an `enum`? 🤔

yuliaveres: nope, i think both are ok. 
i used structs previously for already mentioned grouping reasons (type, purpose of constants inside)
which one do u think is preferred ?

Elland: I think enums seem more fitting, but I think I'm fine either way, since we are not using exhaustive switches.

PR #197, opened 2017-04-25
Run swiftformat 0.28.3

3lvis:


Token/Controllers/Apps/Views/ReputationView.swift
```diff
@@ -100,7 +100,7 @@ class ReputationView: UIView {
 
     func addSubviewsAndConstraints() {
 
-        for guide in guides {
+        for guide in self.guides {
```
yuliaveres: so in the end we tend to use self-s everywhere ? 😢

3lvis: I think we agreed that you'll use whatever you want, selfs or not, but at the end of the day we'll run `swiftformat` and it will fix it for us.

Elland: Actually we can't do that, because `swiftformat` can only add selfs for stuff declared inside the same file.

3lvis: That's a shame.

Elland: That's why I looked into swift lint, as well. Where they have a RP open that never gets finished.

3lvis:
Can we merge this now?

PR #198, opened 2017-04-25
Add SwiftFormat to README

3lvis:


Elland:
👌 Grand. We should look into a CONTRIBUTING manifest, since we're getting outside issues now.

PR #202, opened 2017-04-27
Updated SignalServiceKit and SignalProtocolKit.

Elland:
Please consider doing some live testing messaging other devices.


Token.xcodeproj/project.pbxproj
```diff
@@ -1317,6 +1321,7 @@
 				9FB36EA81E7C27A300DB6F86 /* menuButton.m4a in Resources */,
 				149F9AD21E72E29A00FB74AA /* SF-UI-Text-Medium.otf in Resources */,
 				149F9AD31E72E29A00FB74AA /* SF-UI-Text-Bold.otf in Resources */,
+				9F3FF7F61EB3341F002D0293 /* uk.svg in Resources */,
```
yuliaveres: what is that image for ? :)

3lvis:
Looks fine to me, gonna do some testing.

Token/Controllers/Messaging/MessagesViewController.swift
```diff
@@ -699,7 +711,7 @@ extension MessagesViewController: ImagePickerDelegate {
                 let timestamp = NSDate.ows_millisecondsSince1970(for: Date())
                 let outgoingMessage = TSOutgoingMessage(timestamp: timestamp, in: self.thread, messageBody: "")
 
-                self.messageSender.sendAttachmentData(imageData, contentType: "image/jpeg", filename: "", in: outgoingMessage, success: {
+                self.messageSender.sendAttachmentData(imageData, contentType: "image/jpeg", filename: "image.jpeg", in: outgoingMessage, success: {
```
yuliaveres: the image name not the best one :/ but dnt have better idea now

Token/Models/Message.swift
```diff
@@ -32,8 +32,15 @@ public class Message: NSObject, NOCChatItem {
                 let attachment = TSAttachment.fetch(withUniqueID: attachmentId)!
                 if attachment is TSAttachmentPointer {
                     return [#imageLiteral(resourceName: "placeholder")]
-                } else if let stream = attachment as? TSAttachmentStream, let image = stream.image() {
-                    return [image]
+                } else if let stream = attachment as? TSAttachmentStream {
+                    if stream.isVideo(), let thumbnail = stream.image() {
+                        return [thumbnail]
```
yuliaveres: are we okay with multiple returns in a method ? 😢

Token/Models/Message.swift
```diff
@@ -32,8 +32,15 @@ public class Message: NSObject, NOCChatItem {
...
+                    } else if stream.isImage(), let image = stream.image() {
```
yuliaveres: also, both checks look similar, only namings are different, can we use common logic for video / image stream's image() ?

Token/Models/Message.swift
```diff
@@ -32,8 +32,15 @@ public class Message: NSObject, NOCChatItem {
...
+                        return [image]
+                    } else if let url = stream.mediaURL() {
+                        print(url)
```
yuliaveres: do we need print here ? if not, should remember to remove it 😉

Carthage/Build/iOS/ImagePicker.framework/Headers/ImagePicker-Swift.h
```diff
@@ -198,6 +198,7 @@ SWIFT_CLASS("_TtC11ImagePicker21ImagePickerController")
 @property (nonatomic) NSInteger imageLimit;
 @property (nonatomic) BOOL startOnFrontCamera;
 @property (nonatomic, copy) NSString * _Nullable doneButtonTitle;
+- (nonnull instancetype)initWithNibName:(NSString * _Nullable)nibNameOrNil bundle:(NSBundle * _Nullable)nibBundleOrNil OBJC_DESIGNATED_INITIALIZER;
```
yuliaveres: we should keep consistency between nullability specifiers nullable / _Nullable
(as for me nullable looks more readable.)

Elland: That's not our code tho. 👍

3lvis:
On this update, the "sending" indicator keeps spinning meanwhile in `master` it works fine.

image uploaded from ios 2

Token/Models/Message.swift
Elland: They're only the same for now, because we didn't get the design yet. 👍

Token/Models/Message.swift
Elland: Instead of?

Token/Controllers/Messaging/MessagesViewController.swift
Elland: We could probably extract something from the attachment, once we get there.

Token.xcodeproj/project.pbxproj
Elland: Testing. Gonna delete it!

Token/Models/Message.swift
3lvis: Something like this:

```swift
public var images: [UIImage] {  
    var result = [UIImage]

    if self.signalMessage.hasAttachments() {
        if let attachmentId = (signalMessage.attachmentIds as? [String])?.first {
            let attachment = TSAttachment.fetch(withUniqueID: attachmentId)!
            if attachment is TSAttachmentPointer {
                result = [#imageLiteral(resourceName: "placeholder")]
            } else if let stream = attachment as? TSAttachmentStream {
                if stream.isVideo(), let thumbnail = stream.image() {
                    result = [thumbnail]
                } else if stream.isImage(), let image = stream.image() {
                    result = [image]
                } else if let url = stream.mediaURL() {
                    result = [#imageLiteral(resourceName: "placeholder")]
                }
            }
        }
    }

    return result
}
```

Token/Models/Message.swift
yuliaveres: ok

Token/Models/Message.swift
Elland: Sure, but why? 💭

Token/Models/Message.swift
```diff
@@ -127,5 +127,17 @@ public class Message: NSObject, NOCChatItem {
         self.isActionable = shouldProcess && !isOutgoing && (sofaWrapper?.type == .paymentRequest)
 
         super.init()
+
+        if (signalMessage as? TSOutgoingMessage) != nil {
+            self.signalMessage.addObserver(self, forKeyPath: "self.messageState", options: [.new, .old], context: nil)
```
yuliaveres: don't we want to keep keypaths as constants ?

Token/Models/Message.swift
```diff
@@ -127,5 +127,17 @@ public class Message: NSObject, NOCChatItem {
...
+        }
+    }
+
+    public override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey : Any]?, context: UnsafeMutableRawPointer?) {
+        if keyPath == "messageState" {
```
yuliaveres: and here use the constants would be better, wdyt ?

Token/Models/Message.swift
Elland: Trying to get this to work, nothing so far. 😭 Ideas?

Elland:
@3lvis known issue. Whole morning looking into it (message status update).

Token/Models/Message.swift
```diff
@@ -127,5 +127,17 @@ public class Message: NSObject, NOCChatItem {
         self.isActionable = shouldProcess && !isOutgoing && (sofaWrapper?.type == .paymentRequest)
 
         super.init()
+
+        if (signalMessage as? TSOutgoingMessage) != nil {
```
yuliaveres: also wouldn't that be better `if let signalMessage = signalMessage as TSOutgoingMessage? { ` ...

Token/Models/Message.swift
yuliaveres: that makes understanding code and debugging much easier, with one return place

Token/Models/Message.swift
yuliaveres: please, try `self.signalMessage.addObserver(self, forKeyPath: #keyPath(TSOutgoingMessage.messageState),` ...

Elland:
State tracking should be fixed now. 🚀

PR #205, opened 2017-05-02
Improved nullability annotations, deleted unused code.

Elland:


Token/Controllers/Settings/Backup Phrase/BackupPhraseEnableController.swift
```diff
@@ -94,10 +94,6 @@ class BackupPhraseEnableController: UIViewController {
         ])
     }
 
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
-    }
-
```
roberthein: ✨

Token/Models/ContactsManager.m
```diff
@@ -16,11 +16,11 @@ - (Yap *)yap {
     return [Yap sharedInstance];
 }
 
-+ (BOOL)name:(NSString * _Nonnull)nameString matchesQuery:(NSString * _Nonnull)queryString {
++ (BOOL)name:(nonnull NSString *)nameString matchesQuery:(nonnull NSString *)queryString {
```
yuliaveres: i think those are required only in headers, aren't they ? :)
but not a mistake anyway

Elland: Is that so? I thought they would give me a warning b/c of mismatched nullability indicators?

Token/Supporting Files/AppDelegate.m
```diff
@@ -326,7 +326,7 @@ - (void)updateRemoteNotificationCredentials {
             }
         }];
 
-    } failure:^(NSError * _Nonnull error) {
+    } failure:^(NSError * error) {
```
yuliaveres: i am not sure about the style guide here, maybe *error ?

yuliaveres:
looks good to me too 👍

Token/Supporting Files/AppDelegate.m
Elland: Strangely enough, we don't have one for ObjC 👍 But yeah, I usually do `Class *name`. Adjusting it.

Token/Models/ContactsManager.m
yuliaveres: no, it shouldn't ( as far as I remember)

PR #208, opened 2017-05-03
Added new Me view, using a static tableview.

Elland:
screen shot 2017-05-03 at 14 37 22


Token/Controllers/Settings/SettingsController.swift
```diff
@@ -86,23 +98,20 @@ open class SettingsController: SweetTableController {
...
-        self.tableView.delegate = self
-        self.tableView.separatorStyle = .none
-        self.tableView.contentInset = UIEdgeInsets(top: 20, left: 0, bottom: 0, right: 0)
+    static func instantiateFromNib() -> SettingsController {
+        return UIStoryboard(name: "Settings", bundle: nil).instantiateInitialViewController() as! SettingsController
+    }
 
-        for type in self.cellTypes {
-            self.tableView.register(type)
+    func updateVerificationStatus(_ notification: Notification) {
```
yuliaveres: private ?

Token/Models/EthereumConverter.swift
```diff
@@ -61,11 +61,26 @@ struct EthereumConverter {
         return "\(self.fiatValueString(forWei: balance)) \(Locale(identifier: self.forcedLocale).currencyCode!)"
     }
 
+    public static func balanceSparseAttributedString(forWei balance: NSDecimalNumber, width: CGFloat) -> NSAttributedString {
+        let attributedString: NSMutableAttributedString = self.balanceAttributedString(forWei: balance).mutableCopy() as! NSMutableAttributedString
+        let range = NSRange(location: 0, length: attributedString.length)
+
+        let paragraph = NSMutableParagraphStyle()
+        paragraph.alignment = .left
+        let t = NSTextTab(textAlignment: .right, location: width, options: [:])
```
yuliaveres: really, let's give it better name ? 🙏

Token/Models/EthereumConverter.swift
```diff
@@ -61,11 +61,26 @@ struct EthereumConverter {
...
+        paragraph.tabStops = [t]
+
+        attributedString.addAttribute(NSParagraphStyleAttributeName, value: paragraph, range: range)
+        print(attributedString)
```
yuliaveres: needed ?

Token/Models/EthereumConverter.swift
```diff
@@ -61,11 +61,26 @@ struct EthereumConverter {
         return "\(self.fiatValueString(forWei: balance)) \(Locale(identifier: self.forcedLocale).currencyCode!)"
     }
 
+    public static func balanceSparseAttributedString(forWei balance: NSDecimalNumber, width: CGFloat) -> NSAttributedString {
```
yuliaveres: let's stick to the following:
https://swift.org/documentation/api-design-guidelines/
then I would write `public static func balanceSparseAttributedString(for wei: NSDecimalNumber, width: CGFloat) -> NSAttributedString {`

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -38,42 +38,54 @@ open class SettingsController: SweetTableController {
...
-        NotificationCenter.default.addObserver(self, selector: #selector(self.updateVerificationStatus(_:)), name: SettingsController.verificationStatusChanged, object: nil)
+    public var chatAPIClient: ChatAPIClient {
+        return ChatAPIClient.shared
     }
 
-    func updateVerificationStatus(_ notification: Notification) {
-        if let verificationStatus = notification.object as? VerificationStatus {
-            self.verificationStatus = verificationStatus
-        }
+    public var idAPIClient: IDAPIClient {
```
yuliaveres: do we need it public ? in other controllers it is not, and seems like not used outside :/ 
please check, maybe possible private ?

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -148,80 +157,72 @@ open class SettingsController: SweetTableController {
...
 
-                    if self.didVerifyBackupPhrase == true, cell.checkbox.checked == false {
-                        cell.checkbox.bounce()
-                    }
+        case 2:
+            switch indexPath.row {
+            case 0:
+                // passphrase backup
+                self.navigationController?.pushViewController(BackupPhraseEnableController(), animated: true)
+            case 1:
```
yuliaveres: why we need this case if default breaks?

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -148,80 +157,72 @@ open class SettingsController: SweetTableController {
...
+                // go to user profile
+                self.navigationController?.pushViewController(ProfileController(), animated: true)
+            case 1:
+                break // go to qr code
+            default:
+                break
+            }
+        case 1:
+            switch indexPath.row {
+            case 1:
```
yuliaveres: same as in second section

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -148,80 +157,72 @@ open class SettingsController: SweetTableController {
...
+                break // trusted frieds
+            default:
+                break
+            }
 
-                    cell.checkbox.checked = self.didVerifyBackupPhrase
-                }
+        case 3:
+            switch indexPath.row {
+            case 0:
```
yuliaveres: and here

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -148,80 +157,72 @@ open class SettingsController: SweetTableController {
...
+    open override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        self.performAction(for: indexPath)
     }
 
-    public func tableView(_: UITableView, estimatedHeightForRowAt _: IndexPath) -> CGFloat {
-        return UITableViewAutomaticDimension
-    }
+    open override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        if section == 2 {
+            let view = SettingsSectionHeader(title: "Security", error: "Your account is at risk")
```
yuliaveres: i think if we target for locales in future, lets already use NSLocalized.. while adding new code ?

yuliaveres:
Also, doesn't it look like only "My QR Code" gets aligned with sections titles, but rest cells not ? ;(

Token/Controllers/Settings/SettingsController.swift
Elland: If they're private, ObjectiveC runtime can't see it. This is a notification handler, gotta be internal at least.

Token/Controllers/Settings/SettingsController.swift
Elland: These are placeholders for the actions we'll implement later. I like having this already scaffolded here, so I don't have to double check the indices again later. 👍

Token/Controllers/Settings/SettingsController.swift
Elland: There was a previous discussion about how to improve on the localisation thing. We can discuss it tomorrow all in person!

Token/Models/EthereumConverter.swift
Elland: Then we can't differentiate between `for wei` and `for ether`. They'll conflict.

Token/Models/EthereumConverter.swift
Elland: Hoopla, forgot to rename this. Thanks. :)

Token/Controllers/Settings/SettingsController.swift
roberthein: Ok. 👌 But I think in general it's never a good idea to use hardcoded strings.

Elland: Agreed!

Elland:
Approved? 🤖

Token/Controllers/Settings/SettingsController.swift
yuliaveres: u can mark @objc private func

Token/Models/EthereumConverter.swift
yuliaveres: sad :(

Elland:
Alright. All set?

PR #209, opened 2017-05-04
Added basic user blocking and reporting. 

Elland:


Elland:
Please let me now post hoc.

Issue #220, opened 2017-05-04
Content overlaps searchBar on Favourites


yuliaveres:
While started searching someone and got results, scrolling table makes content go over search.
![img_5799](https://cloud.githubusercontent.com/assets/27722554/25705954/4cb6a5c2-30df-11e7-8e81-012a13c039f0.PNG)

Issue #218, opened 2017-05-04
Browse search bar wrong top offset


yuliaveres:
When started typing into searchbar on Browse page,
and trying to scroll up-down, 
- searchBar is getting cropped;
- top offset is not ok
![img_5797](https://cloud.githubusercontent.com/assets/27722554/25705731/9a87ae1e-30de-11e7-8203-718e9edc3e9d.PNG)
![img_5798](https://cloud.githubusercontent.com/assets/27722554/25705779/c25b98a6-30de-11e7-98d0-a620e0775cca.PNG)

marijnschilling:
Maybe it's totally unrelated, but I had a similar (yet opposite) problem this morning. Where the scrollView used an offset of 20px on the top while it shouldn't 🙈 : https://github.com/bakkenbaeck/Lift/pull/17

I needed to set `automaticallyAdjustsScrollViewInsets` on the viewController to false.

PR #225, opened 2017-05-09
Fixed some layout glitches.

Elland:
Including: 
- #220 
- #219 
- #218


Token/Controllers/Settings/ProfileEditController.swift
```diff
@@ -40,21 +40,14 @@ open class ProfileEditController: UIViewController {
...
     }
 
-    lazy var toolbar: UIToolbar = {
-        let view = UIToolbar(withAutoLayout: true)
-        view.barTintColor = Theme.viewBackgroundColor
-        view.tintColor = Theme.tintColor
-        view.delegate = self
-
-        let space = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
+    lazy var navbar: UINavigationBar = {
```
yuliaveres: private ?

Token/Controllers/Settings/ProfileEditController.swift
```diff
@@ -172,6 +172,12 @@ open class ProfileEditController: UIViewController {
             }
         }
     }
+
+    @objc private func didTapView(sender: UITapGestureRecognizer) {
+        if sender.state == .recognized {
+            self.becomeFirstResponder()
```
yuliaveres: what is the purpose of the gesture ? not resigning ?

Elland: To resign, I need to know exactly which table view cell's text field is the active one. This way, I don't have to care, keeping everything decoupled. 🚀

Elland:
@yuliaveres ready for review again. 🚀 

Doesn't fix every issue, but improves a lot.

Token/Controllers/Settings/ProfileEditController.swift
```diff
@@ -72,56 +68,59 @@ open class ProfileEditController: UIViewController {
...
         super.viewDidLoad()
 
-        self.view.backgroundColor = Theme.viewBackgroundColor
+        self.view.backgroundColor = Theme.navigationBarColor
         self.addSubviewsAndConstraints()
 
         guard let user = TokenUser.current else { return }
         self.avatarImageView.image = user.avatar
+
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(ProfileEditController.didTapView))
```
yuliaveres: just curious, what happens in case just -> #selector(didTapView)

Elland: I think it calls our method even if a subclass overloads it. But it's mostly a reflex, used to doing that in ObjC.

PR #237, opened 2017-05-16
New user storing system.

Elland:
1. Fixed and improved a few issues with contact handling, like not being able to access their stored data unless they were a favourite, or missing avatars for non-favourited chat windows.
2. Changed how we store private keys. It's now in the Keychain, that means it can persist across installs and even if our local database file is corrupted or tampered with, it's there. Safer, more resilient.

As a side-effect, we now persist accounts across installs. So even if you accidentally delete the app without securing the account, it will restored it. On the other hand, if you've already secured it, it will force you to go through the account securing method again. Not yet sure how we'd avoid that.


Elland:
Cancelled.

PR #238, opened 2017-05-16
Add web login functionality

tristan:
Allows login to https://admin.service.tokenbrowser.com/ 

note: admin page expects "live" env users, so dev env users wont work, and users are also whitelisted. If you want to test with a live user, ping me on slack with the username and i'll add it to the whitelist.

i've tested this with a local copy of the admin page which i configured to use dev env users.


Elland:
Will check this shortly! 👏🏽

PR #239, opened 2017-05-17
Layout and UX updates.

Elland:


Elland:
Sorry for the massive PR, but with a redesign, it's hard to keep changes small.

PR #243, opened 2017-05-19
Added bot greeting, fixed a bug with contact setup.

Elland:
demo


Token/APIClients/idAPIClient.swift
```diff
@@ -276,17 +276,8 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
 
                     success(contact, self.cacheExpiry)
                 case .failure(_, let response, let error):
-                    if response.statusCode == 404 {
```
yuliaveres: what happened to this scenario ?

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -204,6 +208,10 @@ class ChatController: MessagesCollectionViewController {
         self.thread.markAllAsRead()
         SignalNotificationManager.updateApplicationBadgeNumber()
         self.title = self.thread.cachedContactIdentifier
+
+        if self.contact.isApp && self.messages.isEmpty {
+            self.sendMessage(sofaWrapper: SofaMessage(body: ""))
```
yuliaveres: empty ? 😄

Elland: Yeah, I should add a comment here, just a moment.

yuliaveres: - is there possibility to delete chats?
- do we need this bit each time we start conversation with new contact ?

Elland: You can `archive` it. So it's not visible in the UI, but if you restart chat with that contact, history was preserved. At least for now.

We only send this the first time you open a window with a bot, so you're greeted by them, instead of having to initiate the conversation.

yuliaveres:
ok, good to merge

PR #244, opened 2017-05-19
Added SOFA webview support.

faymontage:
Here's a rough go of it


Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -16,6 +16,48 @@
 import UIKit
 import SweetUIKit
 
+class LaunchAppButton: UIControl {
+    lazy var titleLabel: UILabel = {
```
yuliaveres: don't we want to keep set option only inside LaunchAppButton ? private(set)

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -16,6 +16,48 @@
...
+        let view = UILabel(withAutoLayout: true)
+
+        return view
+    }()
+
+    lazy var chevron: UIImageView = {
```
yuliaveres: same here

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -76,16 +144,50 @@ class BrowseController: SearchableCollectionController {
...
+        UIView.animate(withDuration: 0.25) {
+            self.collectionView.layoutIfNeeded()
+            self.openURLButton.setAttributedTitle(nil)
+        }
+    }
+
+    func didTapOpenURLButton() {
+        let sofaController = SOFAWebController()
+
+        sofaController.load(url: self.searchController.searchBar.text!)
```
yuliaveres: please, don't do force unwrap here.
u can't be sure, 
it is enough to clear searchbar and press url button simultaneously -> boom

yuliaveres: nil_crash

Token/Controllers/Contacts/FavoritesController.swift
```diff
@@ -142,6 +142,15 @@ open class FavoritesController: SweetTableController {
         super.viewDidDisappear(animated)
     }
 
+    open override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        for view in self.searchController.searchBar.subviews {
+            view.clipsToBounds = false
```
yuliaveres: i don't know what u prefer, for me looks better:
`self.searchController.searchBar.subviews.forEach{ $0.clipsToBounds = false }`

Token/Controllers/Contacts/FavoritesController.swift
```diff
@@ -370,6 +379,8 @@ extension FavoritesController: UISearchBarDelegate {
         searchBar.text = nil
         self.displayContacts()
     }
+
+
```
yuliaveres: line

Token/Extensions/String+isURL.swift
```diff
@@ -0,0 +1,30 @@
...
+
+import Foundation
+
+extension String {
+    public var isURL: Bool {
+        let fullRange = NSRange(location: 0, length: self.length)
+        let detector = try! NSDataDetector(types: NSTextCheckingAllSystemTypes)
+        for match in detector.matches(in: self, options: [], range: fullRange) {
+            if match.resultType == .link && match.range.location == fullRange.location && match.range.length == fullRange.length {
+                return true
```
yuliaveres: please, keep single return 😢

Token/SOFAWebview/SOFAWebview.swift
```diff
@@ -0,0 +1,173 @@
...
+                js += try String(contentsOfFile: filepath)
+                print("Loaded sofa.js")
+            } catch {
+                print("Failed to load sofa.js")
+            }
+        } else {
+            print("Sofa.js not found in bundle")
+        }
+
+        var userScript:WKUserScript =  WKUserScript(source: js, injectionTime: WKUserScriptInjectionTime.atDocumentStart, forMainFrameOnly: false)
```
yuliaveres: :_ WKUserScript

Token/SOFAWebview/SOFAWebview.swift
```diff
@@ -0,0 +1,173 @@
...
+    }()
+
+    fileprivate lazy var forwardBarButtonItem: UIBarButtonItem = {
+        UIBarButtonItem(customView: self.forwardButton)
+    }()
+
+    fileprivate lazy var toolbar: UIToolbar = {
+        let view = UIToolbar(withAutoLayout: true)
+
+        // view.tintColor = Theme.tintColor
```
yuliaveres: needed ?

Token/SOFAWebview/SOFAWebview.swift
```diff
@@ -0,0 +1,173 @@
...
+        self.toolbar.set(height: 44)
+        self.hidesBottomBarWhenPushed = true
+        self.webView.topAnchor.constraint(equalTo: self.topLayoutGuide.bottomAnchor).isActive = true
+        self.webView.fillSuperview()
+        self.webView.bottomAnchor.constraint(equalTo: self.toolbar.topAnchor).isActive = true
+
+    }
+
+    public func load(url: String) {
+        let request = URLRequest(url: URL(string: url)!)
```
yuliaveres: same here, what if string is broken?

docs: ` "Returns `nil` if a `URL` cannot be formed with the string (for example, if the string contains characters that are illegal in a URL, or is an empty string)." `

check if url is here, i know now you call it with url button title, but it might be called from anywhere with not secure string :/

Token/SOFAWebview/SOFAWebview.swift
```diff
@@ -0,0 +1,173 @@
...
+
+extension SOFAWebController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        self.backBarButtonItem.isEnabled = webView.canGoBack
+        self.forwardBarButtonItem.isEnabled = webView.canGoForward
+    }
+}
+
+extension SOFAWebController: WKScriptMessageHandler {
+    func jsCallback(callbackId: String, payload: String) {
```
yuliaveres: private ?

Token.xcodeproj/project.pbxproj
```diff
@@ -702,6 +711,15 @@
 			name = Frameworks;
 			sourceTree = "<group>";
 		};
+		84C1952A1ECF056800B9512F /* SOFAWebview */ = {
+			isa = PBXGroup;
+			children = (
+				84C1952E1ECF05C100B9512F /* sofa-web3.js */,
+				84C1952B1ECF057E00B9512F /* SOFAWebview.swift */,
```
Elland: Forgot to rename file!

Token/Controllers/Apps/Controllers/BrowseController.swift
Elland: I don't think we want set at all. :D `fileprivate`?

Token/Controllers/Apps/Controllers/BrowseController.swift
Elland: this one has the image being set!

Token/Controllers/Contacts/FavoritesController.swift
Elland: I don't feel strongly one way or the other, except for the `$0`. I stopped use positional arguments in the 90s, 😂.

Token/Extensions/String+isURL.swift
Elland: Hmm, I think this is worse: 
```swift
public var isURL: Bool {
        var isURL = false

        let fullRange = NSRange(location: 0, length: self.length)
        let detector = try! NSDataDetector(types: NSTextCheckingAllSystemTypes)
        for match in detector.matches(in: self, options: [], range: fullRange) {
            if match.resultType == .link && match.range.location == fullRange.location && match.range.length == fullRange.length {
                isURL = true
                break
            }
        }

        return isURL
    }
```

Token/Controllers/Apps/Controllers/BrowseController.swift
yuliaveres: private(set)

Token/Controllers/Contacts/FavoritesController.swift
yuliaveres: with swift i started 😃

Elland: I think this is one of those awful awful design decisions that shouldn't even be an option. 😅 Character space is free on a 1TB drive, hehe. ¯\_(ツ)_/¯ but what can I do.

Token/SOFAWebview/SOFAWebview.swift
Elland: spacing, you mean? Fixed all the spacing with `swiftformat` now, forgot to run it. 🤦🏽‍♂️

Token/SOFAWebview/SOFAWebview.swift
Elland: This shouldn't even be a string, but a `URL`, right? Fixed it 🚀

Token/SOFAWebview/SOFAWebview.swift
Elland: can it be private inside an extension or fileprivate?

yuliaveres: yep

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -76,16 +141,52 @@ class BrowseController: SearchableCollectionController {
...
+
+        self.openURLButton.isHidden = true
+        self.openURLButtonTopAnchor.constant = -self.searchBar.frame.maxY
+        UIView.animate(withDuration: 0.25) {
+            self.collectionView.layoutIfNeeded()
+            self.openURLButton.setAttributedTitle(nil)
+        }
+    }
+
+    func didTapOpenURLButton() {
```
yuliaveres: private as well

Token/Extensions/String+isURL.swift
yuliaveres: i dnt think so :/

Elland: Well, you have a variable, so there's state handling, and a break. With two returns, everything is static. Easier to follow. Also quite sure the compiler will optimise that variable out to multiple returns. 😅 I really don't see the advantage of a single return here. Also, multiple returns is kinda baked into swift (b/c of `guard`). 🤔

yuliaveres: that is why i am using guard as less as possible;
whatever - as YOU wish

Elland: Oh, you're avoiding `guard`?  😱 What do you do instead?

yuliaveres: optional chaining if let..

yuliaveres: not always, but prefer

Elland: But `guard` is so much safer and cleaner. ☹️

Token/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -16,6 +16,46 @@
 import UIKit
 import SweetUIKit
 
+class LaunchAppButton: UIControl {
+    fileprivate lazy var titleLabel: UILabel = {
+        UILabel(withAutoLayout: true)
```
yuliaveres: does that work ? where you lost return ? :)

Elland: Yeah, implicit returns for one-line blocks. I don't like it, adding them back. :D

PR #258, opened 2017-05-29
Added new attachment picker.

yuliaveres:


Elland:
@yulia needs to update to latest message UI before we can merge it, 🚀

Elland:
@yuliaveres check this out when you can. 🚀 It points to 1.1.4 now.

PR #259, opened 2017-05-29
Migrated payment address.

Elland:
Payment address now comes from the new derivation path, same as Metamask's.


Token/Models/Cereal.swift
```diff
@@ -81,8 +81,8 @@ public class Cereal: NSObject {
         let idPrivateKey = idKeychain.key.privateKey.hexadecimalString()
         self.idCereal = EtherealCereal(privateKey: idPrivateKey)
 
-        // wallet path: 0H/0/0
-        let walletKeychain = self.mnemonic.keychain.derivedKeychain(at: 0, hardened: true).derivedKeychain(at: 0).derivedKeychain(at: 0)
+        // wallet path: 44H/60H/0H/0
+        let walletKeychain = self.mnemonic.keychain.derivedKeychain(at: 44, hardened: true).derivedKeychain(at: 60, hardened: true).derivedKeychain(at: 0, hardened: true).derivedKeychain(at: 0).derivedKeychain(at: 0)
```
yuliaveres: why .derivedKeychain(at: 0) twice ?

yuliaveres: comment says 44H/60H/0H/0

Elland: Because we need the first key in that path. Should probably update comment.

Elland:
Updated.

PR #260, opened 2017-05-29
Fix optionality for signing callback, use hex encoded value string

yuliaveres:


Elland:
Looks good to me now. But it's 🌞hot as hell so I could be wrong.

yuliaveres:
@faymontage , please check this

PR #265, opened 2017-05-31
Splash and signin flow

yuliaveres:


Token/Controllers/Sign In/SplashViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+    
+    @objc private func signinPressed(_ sender: UIButton) {
+        let controller = SignInController()
+        self.navigationController?.pushViewController(controller, animated: true)
+    }
+    
+    @objc private func newAccountPressed(_ sender: UIButton) {
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
+        appDelegate.createNewUser()
```
Elland: Dismissing here and inside the `SignInController` should work. 🚀

yuliaveres: yep, inside Signin as well ! thanks

PR #266, opened 2017-05-31
Added basic id server migration for payment address.

Elland:
To test, install from 042fc32 or earlier, creating a new user, and then run from this branch. User should have new payment address both locally and on ID service.


Elland:
@yuliaveres thoughts?

PR #267, opened 2017-05-31
Fix alert message being dismissed automatically

yuliaveres:


Token/Controllers/TabBarController.swift
```diff
@@ -106,7 +106,7 @@ open class TabBarController: UITabBarController {
     private func presentAddressChangeAlertIfNeeded() {
         guard UserDefaults.standard.bool(forKey: AddressChangeAlertShown) == false else { return }
 
-        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
Elland: I think we should remove this delay and remove the VC being dismissed inside `didCreateUser`.

yuliaveres: you think it is better to show alert without delay at all ?

yuliaveres: i think it feel better with delay 😄

Elland: I am assuming this only shows after you've signed in/signed up, right?

yuliaveres: i dnt see point of removing dismissal there, as i described, with new flow we will need to dismiss sign in scene :/

Elland: But that is not dismissing sign in, that is dismissing anything that is being currently presented. 😜

yuliaveres: in this case i need to make this change in sign in flow pr . are you ok with that ?

Elland: Show me what you have in mind.

yuliaveres: https://github.com/tokenbrowser/token-ios-client/pull/265/files#diff-73a30b83c4db1a4cc6e681e79bb192a5R137

here i will dismiss and only then will remove this line

yuliaveres: regarding showing only after  signed in/signed up - nope, oppositely, it shows if you are using app already

Elland: 👌🏽

Elland:
@yuliaveres ready for merging?

yuliaveres:
@Elland  Yes !

PR #273, opened 2017-06-02
Empty states

yuliaveres:


Token/Controllers/Contacts/FavoritesController.swift
```diff
@@ -99,7 +114,17 @@ open class FavoritesController: SweetTableController {
 
     open override func viewDidLoad() {
         super.viewDidLoad()
-
+        
+        self.view.addSubview(emptyStateContainerView)
```
Elland: Don't we have an `addSubviewsAndConstraints` method for this VC?

yuliaveres: nope, will add !

Elland:
Gonna try and test after lunch so we can ship this!

yuliaveres:
@Elland  making last changes, will let you know !

yuliaveres:
@ella, @roberthein , please test and check this !  i think we may ship this asap

Token/Controllers/EmptyStateView.swift
```diff
@@ -0,0 +1,129 @@
...
+    }
+    
+    func actionButton() -> UIButton {
+        let button = UIButton(withAutoLayout: true)
+        button.backgroundColor = Theme.tintColor
+        button.titleLabel?.font = Theme.regular(size: 16.0)
+        button.layer.cornerRadius = 5.0
+        button.contentEdgeInsets = UIEdgeInsets(top: 0.0, left: 30.0, bottom: 0.0, right: 30.0)
+        
+        button.block_setAction { sender in
```
roberthein: Why did you create `block_setAction` instead of using the native `addTarget`?

Token/Controllers/EmptyStateView.swift
```diff
@@ -0,0 +1,129 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+import UIKit
+import SweetUIKit
+
+import ObjectiveC
+
+protocol Emptiable: class {
```
roberthein: I think this file should be named `Emptiable` not `EmptyStateView`.

Token/Controllers/EmptyStateView.swift
```diff
@@ -0,0 +1,129 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+import UIKit
+import SweetUIKit
+
+import ObjectiveC
```
roberthein: You don't need this I think.

Token/Controllers/Messaging/ChatsController.swift
```diff
@@ -185,6 +218,33 @@ open class ChatsController: SweetTableController {
 
         return thread
     }
+    
+    func emptyStateTitle() -> String {
```
roberthein: Maybe it's nicer to implement the `Emptiable` conformance into a extension of `ChatsController`?

Token/Extensions/Button+Action.swift
```diff
@@ -0,0 +1,39 @@
...
+    var block : BlockButtonActionBlock
+    init(block: @escaping BlockButtonActionBlock) {
+        self.block = block
+    }
+}
+
+extension UIButton {
+    func block_setAction(block: @escaping BlockButtonActionBlock) {
+        objc_setAssociatedObject(self, &ActionBlockKey, ActionBlockWrapper(block: block), objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+        addTarget(self, action: #selector(block_handleAction(sender:)), for: .touchUpInside)
```
roberthein: It's good avoid `AssociatedObject` whenever you can, do you really need it here?

Token/Controllers/Messaging/ChatsController.swift
Elland: Always. 🚀

Token/Extensions/Button+Action.swift
```diff
@@ -0,0 +1,39 @@
...
+class ActionBlockWrapper : NSObject {
+    var block : BlockButtonActionBlock
+    init(block: @escaping BlockButtonActionBlock) {
+        self.block = block
+    }
+}
+
+extension UIButton {
+    func block_setAction(block: @escaping BlockButtonActionBlock) {
+        objc_setAssociatedObject(self, &ActionBlockKey, ActionBlockWrapper(block: block), objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
```
roberthein: Why not just use camel-case here?

Token/Extensions/Button+Action.swift
Elland: This looks and feels completely out of place to me as well.

Token/Controllers/EmptyStateView.swift
yuliaveres: the idea is to use protocol for UIViewController which supports emptyState instead of separate empty state controller;
having common configuration code inside protocol extension - is the reason. i need to add target here; it requires selector method @objc exposure; that is not possible inside protocol extension

Token/Extensions/Button+Action.swift
yuliaveres: yes i do

roberthein: Maybe you can do something like `var buttonPressed: ((Void) -> Void)?` instead of `func buttonPressed(sender: AnyObject)`

Token/Extensions/Button+Action.swift
yuliaveres: you mean `objc_setAssociatedObject` ? it is runtime method

roberthein: No I meant `setAction` instead of `block_setAction`. But it would be better if we could get rid of `Button+Action.swift` completely.

yuliaveres: how can I get rid of it if I can't use addTarget api ?

Token/Controllers/EmptyStateView.swift
roberthein: `var buttonPressed: (() -> ())?`
Can you not use something like this instead?

Token/Extensions/Button+Action.swift
Elland: Why can't you use `addTarget`?

Token/Extensions/Button+Action.swift
yuliaveres: how are you going to bing this block with button action @roberthein 
@Elland  , better ideas ?

Elland: I still don't get why you need it, instead of using `addTarget…`.

Token/Extensions/Button+Action.swift
yuliaveres: just see my previous reply

yuliaveres: addTarget requires selector method @objc exposure; that is not possible inside protocol extension

yuliaveres: @objc can only be used with members of classes, @objc protocols, and concrete extensions OF CLASSES

roberthein: You can use this instead: `var buttonPressed: Selector { get }`
Implementation would the be:  
```
var buttonPressed: Selector {
        return #selector(buttonPressed(sender:))
    }
```
Now you can remove `Button+Action.swift` :)

roberthein: And then you can add it as a target to the button like this:
`button.addTarget(self, action: self.buttonPressed, for: .touchUpInside)`

Token/Controllers/EmptyStateView.swift
```diff
@@ -0,0 +1,125 @@
...
+        scrollView.isScrollEnabled = self.isScrollable()
+        scrollView.contentSize = self.sourceView().frame.size
+        
+        return scrollView
+    }
+    
+    func actionButton() -> UIButton {
+        let button = UIButton(withAutoLayout: true)
+        button.backgroundColor = Theme.tintColor
+        button.titleLabel?.font = Theme.regular(size: 16.0)
```
roberthein: Font weight for the button should me `medium`.

Token/Controllers/EmptyStateView.swift
```diff
@@ -0,0 +1,125 @@
...
+    
+    func adjustEmptyView() {
+        let sourceView = self.sourceView()
+        
+        let containerView = UIView(withAutoLayout: true)
+        containerView.setContentHuggingPriority(UILayoutPriorityRequired, for: .vertical)
+        containerView.setContentCompressionResistancePriority(UILayoutPriorityRequired, for: .vertical)
+        
+        let titleLabel = self.label(of: 20.0)
+        let descriptionLabel = self.label(of: 16.0)
```
roberthein: The `descriptionLabel` need some more line spacing.

Token/Controllers/EmptyStateView.swift
```diff
@@ -0,0 +1,125 @@
...
+        button.addTarget(self, action: self.buttonPressed, for: .touchUpInside)
+        
+        return button
+    }
+    
+    func label(of fontSize: CGFloat) -> UILabel {
+        let label = UILabel(withAutoLayout: true)
+        label.numberOfLines = 0
+        label.textAlignment = .center
+        label.font = Theme.regular(size: fontSize)
```
yuliaveres: it is regular in Sketch

roberthein: Latest sketch file on `master`:
screen shot 2017-06-07 at 13 37 48

yuliaveres: yes, but mine shows me : 
screenshot at jun 07 13-41-40

yuliaveres: but yeah, in abstract it is obviously medium 👍

roberthein: Your screenshot also shows a size of 16, that's not the title.

yuliaveres: screenshot at jun 07 13-46-08

roberthein: Strange, do you have the latest versions of sketch and abstract? Is your master up to date?

yuliaveres: master is updated;
abstract Version 0.47.9 (0.47.9.4306)
Sketch 43.2

roberthein: 🤔 Abstract is a little behind but I'm not sure if updating will change anything.

yuliaveres: whatever ! as i told it is obvious on Abstract the title is medium ! Thanks for this good catch 😄

roberthein:
@yuliaveres Looks great! But can you still rename `EmptyStateView.swift` to `Emptiable.swift`?

yuliaveres:
sure

roberthein:
Ship it! 🚀

PR #275, opened 2017-06-06
[WIP] Support contract creation in signing UI

faymontage:
When a dapp is attempting to prompt the user to deploy a contract, the tx will have no "to" parameter. Regardless, we are delegating validation/verification of the tx, and should not treat any field as "required" here for the time being.


Token/SOFAWebview/SOFAWebController.swift
```diff
@@ -219,43 +219,30 @@ extension SOFAWebController: WKScriptMessageHandler {
...
-                
-                var userInfo = UserInfo(address: to, avatar: nil, name: nil, username: to, isLocal: false)
-
-                if let user = user as TokenUser? {
-                    let avatar = user.avatar != nil ? user.avatar : UIImage(color: UIColor.lightGray)
-                    userInfo.avatar = avatar
-                    userInfo.username = user.username
-                    userInfo.name = user.name
-                    userInfo.isLocal = true
+            var parameters: [String: Any] = [:]
```
yuliaveres: in this case better to instantiate parameters immediately and keep it let / constant ?

Token/SOFAWebview/SOFAWebController.swift
```diff
@@ -219,43 +219,30 @@ extension SOFAWebController: WKScriptMessageHandler {
...
+                        userInfo.username = user.username
+                        userInfo.name = user.name
+                        userInfo.isLocal = true
+                    }
+
+                    self.displayPaymentConfirmation(userInfo: userInfo, parameters: parameters, callbackId: callbackId)
                 }
-
+            } else {
+                let userInfo = UserInfo(address: "", avatar: nil, name: "New Contract", username: "", isLocal: false)
```
yuliaveres: make `userInfo` outside the scope and call `displayConfirmation` once ?

faymontage: The first `displayConfirmation` call is in an async closure, so there's a little more to do than just moving `userInfo` outside the scope - will take a look time permitting

yuliaveres: ohh ! okay just leave it ! sorry

PR #277, opened 2017-06-06
Updated chats controller design.

roberthein:
img_1377
img_1376


Token/Views/ChatCell.swift
```diff
@@ -24,16 +24,27 @@ class ChatCell: UITableViewCell {
         didSet {
             // last visible message
             if let message = self.thread?.messages.last, let messageBody = message.body {
+                
+                let paragraphStyle = NSMutableParagraphStyle()
```
Elland: extract attributes into a property 👌🏽

Token/Views/ChatCell.swift
```diff
@@ -68,52 +81,64 @@ class ChatCell: UITableViewCell {
...
         }
     }
-
-    lazy var unreadLabel: UILabel = {
-        let view = UILabel(withAutoLayout: true)
-        view.textAlignment = .center
-        view.font = Theme.medium(size: 12)
-        view.textColor = Theme.lightTextColor
+    
+    lazy var unreadView: UIView = {
```
Elland: Why inside a UIView? Code comment with explanation? 👍 Also is that why I don't see an ellipsis (`…`) in truncated messages?

Token/Views/ChatCell.swift
```diff
@@ -125,8 +150,14 @@ class ChatCell: UITableViewCell {
 
     override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
-
-        self.contentView.addSubview(self.unreadLabel)
+        
+        self.accessoryType = .disclosureIndicator
+        
+        let selectedBackgroundView = UIView()
+        selectedBackgroundView.backgroundColor = UIColor(white: 0.95, alpha: 1)
```
Elland: Where is this colour coming from? Also being called directly instead of extracted to Theme.

roberthein:
`self.thread?.lastMessageDate()` always returns a date, even before any message is sent.

Elland:
Yes. It defaults to Thread creation date.

Token/Views/ChatCell.swift
```diff
@@ -54,6 +54,8 @@ class ChatCell: UITableViewCell {
                 } else {
                     self.lastMessageDateLabel.text = DateTimeFormatter.dateFormatter.string(from: date)
                 }
+            } else {
+                self.lastMessageDateLabel.text = nil
```
Elland: That never happens. :)

roberthein: but it should :)

Elland: Nothing we can do.

roberthein:
So we just show the date always?

Token/Views/ChatCell.swift
roberthein: @Elland So we just show the date always?

Elland: Correct.

roberthein:
🚀 it!

roberthein:
@Elland Can we merge this?

Elland:
As soon as feedback is implemented 😉 You missed a couple. GH is marking them as outdated for some reason. 👀

roberthein:
I see..

roberthein:
Fixed.

Token/Views/ChatCell.swift
Elland: I still don't understand why not use the `UILabel` directly, is it because of content insets?

roberthein: Yes ofcourse, it only looked good before because you only saw it with single digits, if you had seen it with "20", or "200" (shown above), you could see it was broken.

Elland:
💪🏽ship it!

PR #280, opened 2017-06-07
Hotfixes for 1.1.5.

Elland:


Token/Controllers/Contacts/ContactController.swift
```diff
@@ -214,6 +214,13 @@ public class ContactController: UIViewController {
         super.viewDidAppear(animated)
     }
 
+    public override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        guard let scrollView = self.view as? UIScrollView else { return }
+        scrollView.contentSize.height = self.bottomSeparatorView.frame.maxY
```
yuliaveres: do we need to set this on EACH viewDidLayoutSubviews ? :/

yuliaveres: it is being invoked very often, doesn't setting this after `self.view.layoutIfNeeded` do the job ?

Token/Controllers/PaymentController.swift
```diff
@@ -94,6 +94,12 @@ class PaymentController: UIViewController {
         self.shadowTextField.becomeFirstResponder()
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        self.shadowTextField.becomeFirstResponder()
```
yuliaveres: if that is need here, please remove it from `viewWillAppear`

Elland: This is intended to be redundant, to avoid issues with review in the simulator.

Token/Controllers/Contacts/ContactController.swift
Elland: If size doesn't change, this call has no effect. For a hotfix like this I'd rather be safe than sorry. 👍

yuliaveres: ohh 😢

Token/Controllers/PaymentController.swift
yuliaveres: 😢

Token/Controllers/Contacts/ContactController.swift
Elland: Yeah, I know. It is a hotfix tho, so we'll be cleaning this up shortly. 🚀

Elland:
Sorry about the rushed PR this time, but kinda running against time so we can resubmit ASAP as possible. :dolan:

PR #281, opened 2017-06-07
Fix/messages image gallery

roberthein:
messages-image-gallery-transition


Token/Controllers/Messaging/ChatController.swift
```diff
@@ -595,6 +595,13 @@ class ChatController: MessagesCollectionViewController {
 
 extension ChatController: MessageCellDelegate {
 
+    func didTapImage(in cell: MessageCell) {
+        guard let indexPath = cell.indexPath else { return }
+
+        let controller = ImagesViewController(messages: messageModels, initialIndexPath: indexPath)
+        present(controller, animated: true, completion: nil)
```
Elland: 😉

Elland:
Pretty good, looking forward to having it ready 🚀

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -98,22 +98,18 @@ class ImagesViewController: UIViewController {
...
-        if let currentIndexPath = self.currentIndexPath {
-            self.prepareBeforeDismiss?(currentIndexPath)
-        }
+        self.prepareBeforeDismiss?(currentIndexPath)
 
         self.dismiss(animated: true, completion: nil)
     }
 
-    var currentIndexPath: IndexPath? {
+    var currentIndexPath: IndexPath {
```
Elland: Do you mind keeping the code organised with properties on top and functions at the bottom?

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
```diff
@@ -50,24 +50,40 @@ class ImagesViewControllerTransition: NSObject, UIViewControllerAnimatedTransiti
     }
 
     func thumbnailImageView(for messagesViewController: MessagesViewController, _ imagesViewController: ImagesViewController) -> UIImageView? {
-        let indexPath = imagesViewController.currentIndexPath ?? imagesViewController.initialIndexPath!
-        guard let cell = messagesViewController.collectionView.cellForItem(at: indexPath) as? MessageCell else { return nil }
+        messagesViewController.view.setNeedsLayout()
+        messagesViewController.view.layoutIfNeeded()
+        
+        if !isPresenting {
```
Elland: Leaving to add `self` later will just make code review more annoying and risk us missing spots. 👍

Elland:
Don't forget to merge back from `1.1.4` regularly to avoid conflicts piling up. 👍

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
roberthein: Writing `self` explicitly is not in 'my system' nor will it be in the future. The only way to solve this mismatch in coding-style is to not write `self` explicitly from now on. Like we agreed upon.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: 👍

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
Elland: Not for new projects tho. 

How did you code before swift? 🤔

roberthein: What not for new projects?
And I know, old habits die hard, but it is an old habit that most of us have let go.

roberthein: I wouldn't mind making one clean PR that remove's all explicitly used `self` where we can. To settle this issue once and for all.

roberthein: Else we will have to rely on formatting tools and what we can catch in the reviews.

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -129,7 +131,9 @@ extension ImagesViewController: UICollectionViewDataSource {
     }
 
     func scrollViewDidScroll(_: UIScrollView) {
-        self.prepareBeforeDismiss?(currentIndexPath)
+        if !isInitialScroll {
+            self.prepareBeforeDismiss?(currentIndexPath)
```
Elland: Is this an optional function?

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -365,37 +365,37 @@ class ChatController: MessagesCollectionViewController {
...
+        //                self.messageSender.sendMessage(fromKeyError: (errorMessage as! TSInvalidIdentityKeySendingErrorMessage), success: { () -> Void in
+        //                    print("Got it!")
+        //                }, failure: { (_ error: Error) -> Void in
+        //                    print(error)
+        //                })
+        //            }
+        //        }
+        //        actionSheetController.addAction(acceptSafetyNumberAction)
+        //
+        //        self.present(actionSheetController, animated: true, completion: nil)
```
Elland: Where's this whitespacing coming from? swiftformat should've removed it, right?

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -632,10 +632,25 @@ class ChatController: MessagesCollectionViewController {
             self.navigationController?.pushViewController(contactController, animated: true)
         }
     }
+
+    var scrollToIndex: ImagesViewControllerDismiss {
```
Elland: What's `ImagesViewControllerDismiss`? Also where's `scrollToIndex` used? Also properties got at the top.

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -632,10 +632,25 @@ class ChatController: MessagesCollectionViewController {
...
+        return { indexPath in
```
Elland: I don't know what is going on here. You're returning a closure?

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -632,10 +632,25 @@ class ChatController: MessagesCollectionViewController {
...
+            self.collectionView.scrollToItem(at: indexPath, at: .centeredVertically, animated: false)
+        }
+    }
 }
 
 extension ChatController: MessageCellDelegate {
 
+    func didTapImage(in cell: MessageCell) {
+        guard let indexPath = cell.indexPath else { return }
```
Elland: Cell's shouldn't be aware of their index path. That's the controller's job, no?

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -632,10 +632,25 @@ class ChatController: MessagesCollectionViewController {
...
+
+        let controller = ImagesViewController(messages: messageModels, initialIndexPath: indexPath)
```
Elland: Why is this created with an index path? Looks like `ImagesViewController` (please rename it to follow convention and be more descriptive, btw) knows too much about the collection view controller presenting it. Should probably use a delegate system. So the gallery controller can ask how many image messages it should display, which one is the previous and next one, etc.

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -632,10 +632,25 @@ class ChatController: MessagesCollectionViewController {
...
+        controller.prepareBeforeDismiss = self.scrollToIndex
```
Elland: Not sure what this is doing.

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -632,10 +632,25 @@ class ChatController: MessagesCollectionViewController {
...
+        controller.title = title
+        present(controller, animated: true, completion: nil)
```
Elland: `self`

Token/MessagesUI/Images/ImageCell.swift
```diff
@@ -20,7 +20,7 @@ class ImageCell: UICollectionViewCell {
     override init(frame: CGRect) {
         super.init(frame: frame)
 
-        contentView.backgroundColor = .white
+        contentView.backgroundColor = .clear
```
Elland: Theme.

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -8,6 +8,7 @@ class ImagesViewController: UIViewController {
     var messages: [MessageModel] = []
     var initialIndexPath: IndexPath!
     var prepareBeforeDismiss: ImagesViewControllerDismiss?
+    var isInitialScroll: Bool = true
```
Elland: What is this used for?

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: It is.

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -37,13 +38,25 @@ class ImagesViewController: UIViewController {
         return view
     }()
 
+    lazy var underline: UIView = {
```
Elland: If this is a separator, please follow the convention and call it a `separatorView` as well. 👍

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -37,13 +38,25 @@ class ImagesViewController: UIViewController {
...
+
     private lazy var doneButton: UIBarButtonItem = {
         let view = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(done(_:)))
         view.tintColor = Theme.tintColor
 
         return view
     }()
 
+    var currentIndexPath: IndexPath {
+        let indexPath = self.collectionView.indexPathForItem(at: CGPoint(x: self.collectionView.contentOffset.x + (self.collectionView.bounds.width / 2), y: self.collectionView.bounds.height / 2))
```
Elland: please extract the values into named constants.

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -37,13 +38,25 @@ class ImagesViewController: UIViewController {
...
+        return indexPath ?? self.initialIndexPath
```
Elland: What is the initial index path?

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -62,13 +75,19 @@ class ImagesViewController: UIViewController {
         self.navigationBar.right(to: self.view)
         self.navigationBar.height(64)
 
+        self.navigationBar.addSubview(self.underline)
+        self.underline.bottom(to: self.navigationBar)
+        self.underline.left(to: self.navigationBar)
+        self.underline.right(to: self.navigationBar)
+        self.underline.height(Theme.borderHeight)
```
Elland: Why adding another border to the navigation bar? Doesn't it already come with one?

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -62,13 +75,19 @@ class ImagesViewController: UIViewController {
...
+
         self.view.addSubview(self.collectionView)
         self.collectionView.topToBottom(of: self.navigationBar)
         self.collectionView.left(to: self.view)
         self.collectionView.bottom(to: self.view)
         self.collectionView.right(to: self.view)
 
-        self.navigationBar.setItems([UINavigationItem(title: self.title!)], animated: false)
+        self.navigationBar.setItems([UINavigationItem(title: self.title ?? "")], animated: false)
```
Elland: Why a navigation bar if there's no navigation? `UIToolbar` maybe?

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -80,35 +99,27 @@ class ImagesViewController: UIViewController {
...
+        self.isInitialScroll = false
     }
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
-        if let currentIndexPath = self.currentIndexPath {
-            self.prepareBeforeDismiss?(currentIndexPath)
-        }
+        self.prepareBeforeDismiss?(self.currentIndexPath)
```
Elland: Why an optional function?

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -80,35 +99,27 @@ class ImagesViewController: UIViewController {
...
     }
 
     func done(_: UIBarButtonItem) {
```
Elland: Please follow naming convention for target actions. 👍

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -80,35 +99,27 @@ class ImagesViewController: UIViewController {
...
 }
 
 extension ImagesViewController: UICollectionViewDataSource {
 
-    func collectionView(_: UICollectionView, willDisplay cell: UICollectionViewCell, forItemAt _: IndexPath) {
-        guard cell is ImageCell else { return }
-        // cell.imageUrl = messages[indexPath.row].imageUrl
+    func collectionView(_: UICollectionView, willDisplay cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {
+        guard let cell = cell as? ImageCell else { return }
+        cell.imageView.image = messages[indexPath.row].image
```
Elland: `self`

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -118,6 +129,12 @@ extension ImagesViewController: UICollectionViewDataSource {
     func collectionView(_: UICollectionView, numberOfItemsInSection _: Int) -> Int {
         return self.messages.count
     }
+
+    func scrollViewDidScroll(_: UIScrollView) {
+        if !isInitialScroll {
```
Elland: `self`

Token/MessagesUI/Images/ImagesViewControllerPresentationController.swift
```diff
@@ -3,23 +3,38 @@ import UIKit
 class ImagesViewControllerPresentationController: UIPresentationController {
 
     override func presentationTransitionWillBegin() {
-        guard let controller = self.presentedViewController as? ImagesViewController else { return }
-        guard let containerView = self.containerView, let presentedView = self.presentedView else { return }
 
+        guard let presented = self.presentedViewController as? ImagesViewController else { return }
```
Elland: Please more descriptive names for variables and constants.

Token/MessagesUI/Images/ImagesViewControllerPresentationController.swift
```diff
@@ -3,23 +3,38 @@ import UIKit
 class ImagesViewControllerPresentationController: UIPresentationController {
```
Elland: I'm not sure what `UIPresentationController` is being used for, here.

Token/MessagesUI/Images/ImagesViewControllerPresentationController.swift
```diff
@@ -3,23 +3,38 @@ import UIKit
...
 
     override func presentationTransitionWillBegin() {
-        guard let controller = self.presentedViewController as? ImagesViewController else { return }
-        guard let containerView = self.containerView, let presentedView = self.presentedView else { return }
 
+        guard let presented = self.presentedViewController as? ImagesViewController else { return }
+        presented.collectionView.backgroundColor = nil
+
+        presented.beginAppearanceTransition(true, animated: true)
```
Elland: This looks like a method that will transition the controllers, but it's being called in a `will do something` notification-style method.

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
```diff
@@ -29,40 +33,126 @@ class ImagesViewControllerTransition: NSObject, UIViewControllerAnimatedTransiti
     }
 
     func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
-        switch self.operation {
-        case .present:
-            self.present(with: transitionContext)
-        case .dismiss:
-            self.dismiss(with: transitionContext)
+
+        if isPresenting {
```
Elland: Hard to read what's going on here. You might want to just use one class for each transition.

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
```diff
@@ -29,40 +33,126 @@ class ImagesViewControllerTransition: NSObject, UIViewControllerAnimatedTransiti
...
+        guard let toCell = imagesViewController.collectionView.visibleCells.flatMap({ cell in cell as? ImageCell }).filter({ imageCell in imageCell.frame.width != 0 }).first else { return nil }
 
-        UIView.animate(withDuration: self.duration, delay: 0, usingSpringWithDamping: 1, initialSpringVelocity: 0, options: .easeIn, animations: {
-            controller.collectionView.transform = CGAffineTransform(scaleX: 0.5, y: 0.5)
-            controller.collectionView.alpha = 0
-            controller.navigationBar.alpha = 0
+        return toCell.imageView
+    }
+
+    func animate(with context: UIViewControllerContextTransitioning, _ messagesViewController: MessagesViewController, _ imagesViewController: ImagesViewController) {
```
Elland: Why unnamed parameters?

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -1,16 +1,13 @@
 import UIKit
 
-typealias MessageCellImageTap = (IndexPath) -> Void
-
 protocol MessageCellDelegate {
+    func didTapImage(in cell: MessageCell)
```
Elland: 👌🏽

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -66,11 +63,11 @@ class MessageCell: UICollectionViewCell {
         return view
     }()
 
-    private lazy var imageView: MessageImageView = {
```
Elland: Why no longer private?

Token/MessagesUI/Messages/MessagesViewController.swift
```diff
@@ -35,7 +35,7 @@ class MessagesViewController: OverlayController {
 
     // We need to add this margin to our collection view layout to prevent seeing
     // cells being reused bacause of the layout's misplacement.
-    let margin: CGFloat = 200
+    let margin: CGFloat = 0
```
Elland: If it's 0 now, do we still need this?

Token/MessagesUI/Objects/Extensions.swift
```diff
@@ -78,4 +78,37 @@ extension UIViewAnimationOptions {
     static var easeOut: UIViewAnimationOptions {
         return [.curveEaseOut, .beginFromCurrentState, .allowUserInteraction]
     }
+
+    static var easeInOut: UIViewAnimationOptions {
```
Elland: Please make the name clearer. If it has more behaviour, the name should reflect that.

Token/MessagesUI/Objects/Extensions.swift
```diff
@@ -78,4 +78,37 @@ extension UIViewAnimationOptions {
...
+        return [.curveEaseInOut, .beginFromCurrentState, .allowUserInteraction]
+    }
+}
+
+extension UIImageView {
+
+    func duplicate() -> UIImageView? {
```
Elland: Why convert the image to jpeg instead of assigning it directly?

Elland: Also not sure why it's needed.

Token/MessagesUI/Objects/Extensions.swift
```diff
@@ -78,4 +78,37 @@ extension UIViewAnimationOptions {
...
+    }
+
+    func contentModeAwareImageSize() -> CGSize? {
+        guard let image = image else { return nil }
+        let actualSize = image.size
+
+        switch contentMode {
+        case .scaleAspectFill:
+
+            let scale = max(frame.size.width / actualSize.width, frame.size.height / actualSize.height)
```
Elland: `self`

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Why tho?

Token/MessagesUI/Images/ImagesViewController.swift
```diff
@@ -113,6 +122,54 @@ class ImagesViewController: UIViewController {
...
+    }
+}
+
+extension ImagesViewController: UIGestureRecognizerDelegate {
+
+    func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+
+        if gestureRecognizer == panGestureRecognizer {
+            let translation = panGestureRecognizer.translation(in: panGestureRecognizer.view)
+            return translation.y > translation.x
```
Elland: What is this doind?

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
roberthein: Then we would have a lot of duplicate code that has to stay in sync. Not a good idea.

Elland: Why? It's a simple transition going and another coming back, no? Should be fairly straight forward.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: Added.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: Added.

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
roberthein: It is straight forward. And presenting and dismissing is the same animation but in opposite direction. So this should ideally be written once, not twice.

Token/MessagesUI/Messages/MessageCell.swift
roberthein: I need to read it in the transition.

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
Elland: I've read it three times and I'm still not sure what is going one, what's the difference between the two methods and why animate has no parameter labels. 😕 So I wouldn't say it's as straight forward as it could be.

Elland: Maybe removing `UITransitionContextViewControllerKey` and using the dot-notation only might clear this up.

roberthein: Cleared up! 👍

Token/MessagesUI/Images/ImagesViewControllerPresentationController.swift
roberthein: Removed the whole file due to refactoring.

Token/MessagesUI/Images/ImagesViewControllerPresentationController.swift
roberthein: Removed.

Token/MessagesUI/Images/ImagesViewControllerPresentationController.swift
roberthein: Removed it.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: What would be the convention here?
I can't find what you mean in:
https://github.com/bakkenbaeck/iOS-handbook/blob/master/BEST_PRACTICES.md#naming

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: Refactored!

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: The idea here was that it should look the same as the navigationBar in chat.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Same as Apple's: `didPress…`. 👍

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: The initialIndexPath is the indexPath that is passed to the initializer to start the collectionView at right cell.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: I don't see why it should. Feels light we're fighting the APIs here. :)

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: 👍

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Hmm, so I guess it goes with the other more general comment: this view controller shouldn't be aware of the inner works of its parent.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: 👌

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: To prevent scrolling when the collectionView is being put in place for the first time.

Elland: No sure I follow. Do you have a quick demo/video of what is going on?

Token/MessagesUI/Images/ImageCell.swift
roberthein: Is it really theming to make something transparent? I see it more as a technical solution. If you don't agree, then how would you name it?

Token/Controllers/Messaging/ChatController.swift
roberthein: This is replaced by a delegate conformation.

Token/MessagesUI/Images/ImageCell.swift
Elland: 🤔 good question. `Theme.clearBackgroundColor` maybe?

Token/Controllers/Messaging/ChatController.swift
Elland: > the images view controller should be responsible for displaying the images.

Exactly. But with better isolation. Right now it knows too much about how its parent works, if we change the parent we have to change this as well. With delegation, anything that conforms to the protocol can control the image presentation. Code becomes clearer and less coupled. 👍

Token/MessagesUI/Images/ImageCell.swift
roberthein: But why would you theme that though?

Elland: I guess it's fine for this for now.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: Maybe a better solution would be to wrap the viewController in a navigationController? But recreating a navigationBar in a UIToolBar seems like the wrong way to go for me.

Token/MessagesUI/Objects/Extensions.swift
roberthein: It does not have more behaviour just a different animation curve. I think the naming is consistent with the behaviour and consistent with how apple named it.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Why is it going to look like a navigation bar at all, tho? It's a toolbar, should look like on.

Token/MessagesUI/Objects/Extensions.swift
roberthein: I don't want the same instance, I need to copy it. But you are right `UIImageJPEGRepresentation` is not the way to go. Now using `copy()`.

Token/MessagesUI/Objects/Extensions.swift
Elland: Of course it does. `[.curveEaseInOut, .beginFromCurrentState, .allowUserInteraction]` that's three options combined in one, but the name makes it look like it's just an `easeInOut` curve. Making it both misleading and redundant, since there's already a `curveEaseInOut` option. 😉

Token/MessagesUI/Objects/Extensions.swift
roberthein: Added

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: Returning `true` if it's panning more vertical than horizontal.

Token/MessagesUI/Objects/Extensions.swift
Elland: Why a full on copy tho? If this is for a transition, have you tried using the snapshot API?

UIView's `func snapshotView(afterScreenUpdates afterUpdates: Bool) -> UIView?`.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Not sure that's the clearest or cleanest way to do that. I think what we want to trigger it is a "pull down", so checking if vertical translation is moving down and locking horizontal translation should work. Unless we need to be able to drag the image around for some reason?

Token/MessagesUI/Objects/Extensions.swift
roberthein: So you want to rename all three of these?

Token/MessagesUI/Objects/Extensions.swift
roberthein: I did, but that doesn't do the trick.

Token/MessagesUI/Objects/Extensions.swift
Elland: Yes.

Token/MessagesUI/Objects/Extensions.swift
Elland: Why? What happened?

roberthein: It's about transitioning between two imageViews with different content modes. I can't manipulate a view created with `func snapshotView(afterScreenUpdates afterUpdates: Bool)` the way I need to.

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: I don't see how that would be better.

Elland: Code is clearer, so is functionality. Did I miss something? 🤔

Token/Controllers/Messaging/ChatController.swift
roberthein: This is what my swiftformat did.

Token/Controllers/Messaging/ChatController.swift
roberthein: Maybe good practice for next time?

Token/Controllers/Messaging/ChatController.swift
Elland: Ok then.

Token/Controllers/Messaging/ChatController.swift
Elland: Not sure what you mean.

roberthein: I think what you are proposing is a big change that is not necessary. But I'll keep your idea in mind for the next something like this comes along. 👍

Token/MessagesUI/Images/ImagesViewController.swift
roberthein: I believe this implementation is very clear, I can add some clarity to `translation.y > translation.x` but connecting the dismissal to the scrolling of the collectionView is not a good idea I think.

Token/Controllers/Messaging/ChatController.swift
Elland: It might be a somewhat bigger change than what we'd like, but I do think it's necessary. Catching architectural issues and bugs is one of the reasons we do reviews. :)

roberthein: Agreed, of course. But this is not an architectural issues.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Not sure what you mean with connecting dismissal to scrolling. 

what I had in mind
```swift
if translation.y == pulling down {
    trigger
}
```

Token/Controllers/Messaging/ChatController.swift
Elland: We have a unclear tightly couple view controller implementation. What would you call it if not architectural?

Token/Controllers/Contacts/FavoritesController.swift
```diff
@@ -291,6 +291,8 @@ open class FavoritesController: SweetTableController {
         }
 
         self.tableView.endUpdates()
+
+        self.showOrHideEmptyState()
```
yuliaveres: i think not correct, on master it was removed

Token/Controllers/Messaging/ChatController.swift
```diff
@@ -365,37 +365,37 @@ class ChatController: MessagesCollectionViewController {
...
-//                    print("Got it!")
-//                }, failure: { (_ error: Error) -> Void in
-//                    print(error)
-//                })
-//            }
-//        }
-//        actionSheetController.addAction(acceptSafetyNumberAction)
-//
-//        self.present(actionSheetController, animated: true, completion: nil)
+        //        let keyOwner = self.contactsManager.displayName(forPhoneIdentifier: errorMessage.theirSignalId())
```
yuliaveres: o no , why that commented code ?

roberthein: I don't know who put it there.

Token/Controllers/Contacts/FavoritesController.swift
yuliaveres: self.showOrHideEmptyState

roberthein: ✅

Token/Controllers/Messaging/ChatController.swift
faymontage: looking into it

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
```diff
@@ -29,41 +33,129 @@ class ImagesViewControllerTransition: NSObject, UIViewControllerAnimatedTransiti
...
+        if !self.isPresenting {
+            messagesViewController.collectionView.scrollToItem(at: imagesViewController.currentIndexPath, at: .centeredVertically, animated: false)
         }
 
-        UIView.animate(withDuration: self.duration, delay: 0, usingSpringWithDamping: 0.9, initialSpringVelocity: 10, options: .easeOut, animations: {
-            controller.collectionView.transform = .identity
-        }, completion: nil)
+        guard let cell = messagesViewController.collectionView.cellForItem(at: imagesViewController.currentIndexPath) as? MessageCell else { return nil }
+
+        return cell.imageView.imageView
```
yuliaveres: cell.imageView.imageView

Token/Controllers/Messaging/ChatController.swift
faymontage: This code will need to be re-enabled, but in a new PR as the only thing happening here is whitespace changing.

Token/Controllers/Contacts/FavoritesController.swift
roberthein: Removed

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
roberthein: Fixed!

Token/Controllers/Messaging/ChatController.swift
Elland: Remember that sometimes Github will consider a comment outdated even tho it isn't. For instance this is still missing calls to `self`. Please check where else this might be true. 💪🏽

Token/MessagesUI/Images/ImagesViewControllerTransition.swift
Elland: Still missing here as well.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Still missing.

Token/MessagesUI/Images/ImagesViewController.swift
Elland: Same here.

PR #286, opened 2017-06-12
Prefetch current user

yuliaveres:


Token/Controllers/Settings/ProfileController.swift
```diff
@@ -278,7 +278,9 @@ open class ProfileController: UIViewController {
     }
 
     fileprivate func updateReputation() {
-        RatingsClient.shared.scores(for: TokenUser.current!.address) { ratingScore in
```
Elland: 👏🏽

Token/Models/TokenUser.swift
```diff
@@ -146,6 +137,21 @@ public class TokenUser: NSObject, NSCoding {
         ]
     }
 
+    public static func retrieveCurrentUser() {
+        if let userData = (Yap.sharedInstance.retrieveObject(for: TokenUser.storedUserKey) as? Data), self._current == nil,
```
Elland: I think it's clearer if you remove this from the `if let` call and do a `guard` beforehand. 

```swift
guard self._current == nil else { return }
if let ... {}
```

Token/Supporting Files/AppDelegate.m
```diff
@@ -62,6 +62,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(createOrRestoreNewUser) name:@"CreateNewUser" object:nil];
 
     [self configureAndPresentWindow];
+    [TokenUser retrieveCurrentUser];
```
Elland: 👌🏽

Elland:
Looks solid on my end. Merging.

Token/Models/TokenUser.swift
yuliaveres: oh, sorry, this is odd at all, we check inside current getter the same, so not needed here

PR #289, opened 2017-06-13
Prefetch current user

yuliaveres:


yuliaveres:
Successfully merged into correct release branch

PR #290, opened 2017-06-13
Fetch 32 featured apps instead of default 10

faymontage:


Token/APIClients/AppsAPIClient.swift
```diff
@@ -30,7 +30,7 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
     }
 
     func getFeaturedApps(completion: @escaping (_ apps: [TokenUser], _ error: Error?) -> Void) {
-        self.teapot.get("/v1/apps/featured") { (result: NetworkResult) in
+        self.teapot.get("/v1/apps/featured?limit=32") { (result: NetworkResult) in
```
roberthein: Before on Slack you mentioned that this should be a multiple of 3. That's not the case anymore?

faymontage: I thought the rows were 3 items (was I dreaming?) - they seem to be 2.

Elland: I changed it to two a while ago, to give it more spacing and avoid clipping a lot of bot names. I think @roberthein will start today on the new browser view layout?

roberthein: My plan is to make the transition interactive first and after that process @Elland's feedback on the PR.

faymontage: For the purposes of this PR, 32 is the correct number yes.

PR #292, opened 2017-06-13
Present splash if no user is logged into

yuliaveres:


Elland:
Mind explaining the change, how it was broken before and how this fixes it? 😅

yuliaveres:
Sure

yuliaveres:
So I misunderstood the naming of RequiresSignIn flag; and used it for showing splash;
of course, by default it is NO, so now splash is never shown. it is set to YES only after user is created.

but requires sign in is " not verified ", and has nothing to do with signin + splash flow.

Now if no user logged into, meaning no current user - we show sign in.

Elland:
Hmm, I think this is still the wrong model. 🤔 I'll try to elaborate.

Elland:
Got it. 🚀

PR #294, opened 2017-06-14
Quick fixes and performance improvements.

Elland:
We had a some performance issues when going in and out of a chat, I started noticing after new picker was merged. It was being caused by avatar image size, it was too large and converting it to a hex string was taking too long.


yuliaveres:
Merge it

yuliaveres:
Why u say "after new picker was merged"  @Elland  ?
an image was saved in the very same way as it was previously. we just had `image.resized(toHeight: 320)`

Elland:
@yuliaveres that's when I started noticing it. Not sure if there's something different going on between native picker and our new one that could account for the change. 😅 But works better for me now.

Elland:
@yuliaveres that's when I started noticing it. Not sure if there's something different going on between native picker and our new one that could account for the change. 😅 But works better for me now.

PR #296, opened 2017-06-15
Fix images aspect ratio.

roberthein:
images-aspect-ratio img_1533


Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -266,6 +271,27 @@ class MessageCell: UICollectionViewCell {
...
+                    container.backgroundColor = nil
+                }
+
+                setNeedsLayout()
+                layoutIfNeeded()
+            } else {
+                imageView.widthConstraint?.isActive = false
+
+                setNeedsLayout()
+                layoutIfNeeded()
```
yuliaveres: can you tell this once ?

roberthein: Yes. Fixed.

csdodd:
This got merged whilst I was typing so it's too late now, but leaving it here anyway.

You should consider squashing all these commits into one. As the last three commits are really just fixes for the first commit. You should also read up on the git command `--fixup` which makes it easier to squash later.

PR #297, opened 2017-06-15
Fix for small images when received.

roberthein:
I removed the fallback to image size when the image is really small. This fixes the wrong calculation of the cell heights.


csdodd:
These three commits are related; and can (should) be squashed.

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -272,9 +272,9 @@ class MessageCell: UICollectionViewCell {
                 layoutIfNeeded()
```
Elland: Can you add all the missing `self` calls to this file, while you're at it? 💪🏽

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -272,9 +272,9 @@ class MessageCell: UICollectionViewCell {
...
             }
 
-            if let image = message.image {
+            if message.image != nil {
                 let maxWidth = UIScreen.main.bounds.width - totalHorizontalMargin
-                let maxHeight = min(200, image.size.height)
+                let maxHeight: CGFloat = 200
```
Elland: Thinking if this should be device metric specific. What do you think? It might look to small on the plus and too large on the SE.

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -429,8 +431,8 @@ class MessageCell: UICollectionViewCell {
         var totalHeight: CGFloat = 0
         var totalMargin: CGFloat = 0
 
-        if let image = message.image {
-            let maxHeight = min(200, image.size.height)
+        if message.image != nil {
+            let maxHeight: CGFloat = 200
```
Elland: Used twice, sounds like a good candidate for a constant or property.

Token/MessagesUI/Messages/MessageImageView.swift
```diff
@@ -27,7 +27,7 @@ class MessageImageView: UIImageView {
         super.init(frame: frame)
 
         self.isUserInteractionEnabled = true
-        self.contentMode = .scaleAspectFit
+        self.contentMode = .scaleAspectFill
```
Elland: I thought we didn't want to crop the contents.

Elland:
@csdodd all PRs are squashed automatically before being merged back into master. 💪🏽

PR #299, opened 2017-06-16
Restored link detector, adding deep links for users.

Elland:
1. Restored link detection for chat buttons.
2. Added the severely missed ability to select and copy text.
3. Deep link to users (WIP, needs a link spec, leaving it in as an undocumented nicety for now, unless there are objections.).

demo


Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -78,15 +88,9 @@ class MessageCell: UICollectionViewCell {
         return view
     }()
 
-    private lazy var container: UIView = {
-        UIView()
-    }()
-
-    private lazy var avatar: UIImageView = {
-        let view = UIImageView()
+    private let container = UIView()
```
yuliaveres: why not lazy ?

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -78,15 +88,9 @@ class MessageCell: UICollectionViewCell {
...
 
-        return view
-    }()
+    private let avatar = UIImageView()
```
yuliaveres: the same.

Token/MessagesUI/Messages/MessageCell.swift
```diff
@@ -297,34 +306,34 @@ class MessageCell: UICollectionViewCell {
     override init(frame: CGRect) {
         super.init(frame: frame)
 
-        contentView.isOpaque = false
+        self.contentView.isOpaque = false
 
-        contentView.addSubview(self.statusLabel)
-        self.statusLabel.left(to: contentView)
-        self.statusLabel.right(to: contentView)
+        self.contentView.addSubview(self.statusLabel)
```
yuliaveres: can you put it inside addSubviewsAndConstraints ?

Token/Supporting Files/AppDelegate.m
```diff
@@ -321,6 +321,17 @@ - (void)verifyDBKeysAvailableBeforeBackgroundLaunch {
     }
 }
 
+- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {
+    if ([url.scheme isEqualToString:@"toshi"]) {
+        TabBarController *controller = (TabBarController *)self.window.rootViewController;
+        [controller openDeepLinkURL:url];
+
+        return YES;
+    } else {
```
yuliaveres: no need to keep it inside Else :/

Token/MessagesUI/Messages/MessageCell.swift
Elland: Because it's never changed and it's not setup during creation. My thinking here is that properties should be constant > lazy > optional > force-unwrapped, from most desirable to least desirable.

yuliaveres: why u want to load it immediately, why const most desirable ?

Elland: I feel safer knowing the property can't be re-assigned at runtime. 👍

PR #300, opened 2017-06-19
Restored fingerprint check.

Elland:


Elland:
We have this
```swift
        if let interaction = interaction as? TSInvalidIdentityKeySendingErrorMessage {
            DispatchQueue.main.async {
                self.handleInvalidKeyError(interaction)
            }

            return Message(sofaWrapper: nil, signalMessage: interaction, date: interaction.dateForSorting(), isOutgoing: false)
        }
```

But if it's never being called, then changes run deeper than I first thought. Will need to dig in deeper.

Elland:
Changes are up! 🤞🏽

Elland:
Pointing to correct branch now.

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -366,45 +366,42 @@ class ChatController: MessagesCollectionViewController {
...
+        let actionSheetController = UIAlertController(title: titleText, message: nil, preferredStyle: .actionSheet)
+
+        let dismissAction = UIAlertAction(title: "Cancel", style: .cancel, handler: nil)
+        actionSheetController.addAction(dismissAction)
+
+//        let showSafteyNumberAction = UIAlertAction(title: NSLocalizedString("Compare fingerprints.", comment: "Action sheet item"), style: .default) { (_: UIAlertAction) -> Void in
+//
+//            print("Should display fingerprint comparison UI.")
+//        }
+//        actionSheetController.addAction(showSafteyNumberAction)
```
csdodd: Is there a reason for committing commenting out code? I really advise against doing it. There's no mention of this under iOS best practices but I recommend you add something.

From Android handbook:
```
Don't commit commented out code

 - If the code is so important, why is it commented out?
 - It creates unease amongst other developers -- what is this code?
 - It goes against the purpose of source control
 - It creates noise and adds no value
```

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -366,45 +366,42 @@ class ChatController: MessagesCollectionViewController {
...
-        //                }, failure: { (_ error: Error) -> Void in
-        //                    print(error)
-        //                })
-        //            }
-        //        }
-        //        actionSheetController.addAction(acceptSafetyNumberAction)
-        //
-        //        self.present(actionSheetController, animated: true, completion: nil)
+        let keyOwner = self.contactsManager!.displayName(forPhoneIdentifier: recipientId)
+        let titleText = "Your safety number with \(keyOwner) has changed. You may wish to verify it."
```
csdodd: Looks like this title won't be translated to users local language, is that the case?
What is the standard way of handling strings so that they get translated?

Toshi/Controllers/Messaging/ChatController.swift
Elland: I think we added it now. 👍

Toshi/Controllers/Messaging/ChatController.swift
Elland: Usually using `NSLocalizedString`. 👍

faymontage:
jul-05-2017 13-55-03

Works! We'll have to investigate the new style for this in the future, but this issue can be closed.

Elland:
✨

Elland:
@csdodd removed the commented out code. Didn't add the localised string yet, b/c @roberthein is working on simplifying that for the whole app. Please add ✅ since you're now a reviewer for this PR. 😅

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -1118,64 +1116,64 @@ extension ChatController: PaymentSendControllerDelegate {
...
-
+    
     func animationController(forDismissed dismissed: UIViewController) -> UIViewControllerAnimatedTransitioning? {
         return dismissed is ImagesViewController ? ImagesViewControllerTransition(operation: .dismiss) : nil
     }
-
+    
     func interactionControllerForDismissal(using _: UIViewControllerAnimatedTransitioning) -> UIViewControllerInteractiveTransitioning? {
-
+
```
yuliaveres: @Elland  what about this ?

Elland: @yuliaveres running swift format here changes 13 files. Could you open a PR today that is just a swift format run?

yuliaveres: sure

yuliaveres: 💥 💥 💥

PR #302, opened 2017-06-19
Fix signal being setup too early

yuliaveres:


Elland:
Looks good to me. 💪🏽 Shipping!

PR #303, opened 2017-06-19
Fixes for app store validation errors

faymontage:
* include LaunchScreen.storyboard in distribution
* add NSMotionUsageDescription


Elland:
💪🏽

PR #307, opened 2017-06-20
Get rid of numerous force unwraps

yuliaveres:


faymontage:
I've observed a crash at https://github.com/tokenbrowser/token-ios-client/blob/master/Token/Controllers/Messaging/ChatsController.swift#L213

I don't have any reproduction for how I got into that state - but looks like a good candidate for this effort.

yuliaveres:
yes, a lot of it, and not needed fatal errors

PR #308, opened 2017-06-20
Remove print(response) on successful API calls

faymontage:
IMO it's not interesting to print successful results and would love to see this gone


Elland:
🙏🏽📿👼🏽

roberthein:
@faymontage Good idea, but why not all of them?

faymontage:
Could do - I just removed the ones that spam really hard

PR #311, opened 2017-06-21
Allow user to store backup phrase even when verified

faymontage:


Elland:
If it's already verified, it shouldn't re-verify it. Maybe change it to just show you the passphrase instead?

faymontage:
@Elland I disagree - if my original backup is lost or unavailable, I should be able to store my phrase again, with verification steps and all.

Elland:
🤔 good point!

yuliaveres:
Great !

faymontage:
@yuliaveres As the sole reviewer, please merge this.

PR #312, opened 2017-06-21
Post notification when user is fetched / created / updated

yuliaveres:
Instead of retrieving with completion handler


Token/Controllers/Settings/SettingsController.swift
```diff
@@ -106,6 +108,10 @@ open class SettingsController: UITableViewController {
 
         self.fetchAndUpdateBalance()
     }
+    
+    deinit {
+        NotificationCenter.default.removeObserver(self, name: .currentUserUpdated, object: nil)
```
Elland: This is automatic now in Swift. 💪🏽

Token/Models/TokenUser.swift
```diff
@@ -141,13 +146,13 @@ public class TokenUser: NSObject, NSCoding {
             let deserialised = (try? JSONSerialization.jsonObject(with: userData, options: [])),
             var json = deserialised as? [String: Any] {
 
-            // Because of payment address migration, we have to override the stored payment address.
+            // Because of payment address pmigration, we have to override the stored payment address.
```
Elland: Typo? 👍

Token/Controllers/Settings/SettingsController.swift
yuliaveres: yeah

yuliaveres: i got used to this a lot :D removing

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -97,6 +97,8 @@ open class SettingsController: UITableViewController {
 
     open override func viewDidLoad() {
         super.viewDidLoad()
+        
+        NotificationCenter.default.addObserver(self, selector: #selector(updateUI), name: .currentUserUpdated, object: nil)
```
Elland: `currentUserDidUpdate` would be more idiomatic but this is fine. 👍

Token/Controllers/Settings/SettingsController.swift
```diff
@@ -115,6 +117,13 @@ open class SettingsController: UITableViewController {
         self.usernameLabel.text = TokenUser.current?.displayUsername
         self.userAvatarImageVIew.image = TokenUser.current?.avatar
     }
+    
+    @objc private func updateUI() {
```
Elland: Function name could be clearer, but this works.

Elland:
Looks fine here. 💥✨

PR #314, opened 2017-06-23
Improve performance

yuliaveres:


Token/APIClients/ChatAPIClient.swift
```diff
@@ -34,18 +34,24 @@ public class ChatAPIClient: NSObject {
...
-                print(response)
-                print(json ?? "")
+        DispatchQueue.global(qos: .userInitiated).async {
+            
+            self.teapot.get("/v1/accounts/bootstrap/") { (result: NetworkResult) in
+                
+                DispatchQueue.main.async {
+                    switch result {
+                    case .success(let json, let response):
+                        guard response.statusCode == 200 else { fatalError("Could not retrieve timestamp from chat service.") }
```
Elland: I think we can move timestamp server from crashing the app to returning an error to the user, asking them to try again? And even adding a retry call ourselves first?

Token/APIClients/ChatAPIClient.swift
```diff
@@ -60,25 +66,31 @@ public class ChatAPIClient: NSObject {
             let hashedPayload = cereal.sha3WithID(string: payloadString)
             let message = "PUT\n\(path)\n\(timestamp)\n\(hashedPayload)"
             let signature = "0x\(cereal.signWithID(message: message))"
-
+
```
Elland: Lint! 👍

Token/APIClients/ChatAPIClient.swift
```diff
@@ -60,25 +66,31 @@ public class ChatAPIClient: NSObject {
...
+            
+            DispatchQueue.global(qos: .userInitiated).async {
+                
+                self.teapot.put(path, parameters: requestParameter, headerFields: fields) { result in
+                    
+                    DispatchQueue.main.async {
+                        switch result {
+                        case .success(_, let response):
+                            guard response.statusCode == 204 else {
+                                print("Could not register user. Status code \(response.statusCode)")
```
Elland: Would be great if we could get this reported back to us, too!

Token/Models/ContactsManager.h
```diff
@@ -28,14 +28,13 @@
 
 @interface ContactsManager : NSObject <ContactsManagerProtocol>
 
+@property (nonatomic, copy, readonly, nonnull) NSArray<TokenUser *> *tokenContacts;
```
Elland: And protocol compliance worked? Nice!

Token/AvatarManager.swift
```diff
@@ -0,0 +1,113 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+import Teapot
+import AwesomeCache
+
+final class AvatarManager: NSObject, CacheExpiryDefault {
```
Elland: 😍

Token/Models/ContactsManager.h
yuliaveres: why might not ?

Elland: Property v. method. I thought it wouldn't but ObjctiveC has it's quirks I guess. Also apparently I commented on the wrong line, supposed to be the one below, for the signal stuff.

Token/APIClients/ChatAPIClient.swift
faymontage: (yes, but in a different PR)

Elland: @faymontage we can do something like that? ✨nice!

Token/APIClients/ChatAPIClient.swift
yuliaveres: i think that is also not related

Elland: Should I add a Trello card?

yuliaveres: yes, backlog

Token/APIClients/AppsAPIClient.swift
```diff
@@ -30,43 +30,56 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
...
-                    return
-                }
-
-                let appsJSON = json["results"] as! [[String: Any]]
-                let apps = appsJSON.map { json -> TokenUser in
-                    let app = TokenUser(json: json)
-
-                    return app
+        
+        DispatchQueue.global(qos: .userInitiated).async {
```
yuliaveres: we may try this

yuliaveres: but is that actually needed ?

Token/APIClients/AppsAPIClient.swift
```diff
@@ -30,43 +30,56 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
...
+            self.teapot.get("/v1/apps/featured?limit=32") { (result: NetworkResult) in
+                
+                DispatchQueue.main.async {
```
yuliaveres: oh, in that case no, sure

yuliaveres: thanks Teapot documentation!
`- completion: The completion block, called with a NetworkResult once the request completes.`

Token/APIClients/AppsAPIClient.swift
Elland: I think it results in cleaner calls, not sure if there's any measurable performance improvement to be had tho. More of an organisation thing.

Token/APIClients/AppsAPIClient.swift
Elland: Fixing it there now as well.

Token/Models/ContactsManager.m
```diff
@@ -119,44 +118,40 @@ - (nullable TokenUser *)tokenContactForAddress:(nullable NSString *)address
 
 - (NSArray<SignalAccount *> *)signalAccounts
 {
-    if (!_signalAccounts) {
```
Elland: I missed this. You shouldn't cache it! 👌🏽

Elland:
No assignee designated here, btw.

PR #315, opened 2017-06-23
Refetch contacts while payment

yuliaveres:


Token/Models/ContactsManager.h
```diff
@@ -39,4 +39,6 @@
 
 - (nullable UIImage *)imageForPhoneIdentifier:(nullable NSString *)phoneNumber;
 
+- (void)refreshContacts;
```
Elland: This seems redundant to me. Shouldn't `tokenContacts` still be a computed property?

yuliaveres: not redundant, definitely not, we access them a lot, while looking for a contact f.e.

yuliaveres: we should refetch only when needed

Elland: So we're caching results until we know we need to update it? 🤔 Good call. But then shouldn't this be called in more places or just for payments?

yuliaveres: So in ither cases I listen for db changes, like adding / removing contacts works perfect

yuliaveres: but i am not sure for what changes can i listen in case of payment

yuliaveres: it would be the best

yuliaveres: ok let me try 🤞

Elland: After a lot of digging and experimenting, this looks like the best solution for now.

PR #319, opened 2017-06-26
Remove rest of unwraps

yuliaveres:


Token/APIClients/EthereumAPIClient.swift
```diff
@@ -219,7 +219,7 @@ public class EthereumAPIClient: NSObject {
         self.timestamp { timestamp in
             let cereal = Cereal.shared
             let address = cereal.paymentAddress
-            let paymentAddress = TokenUser.current!.paymentAddress
+            let paymentAddress = TokenUser.current?.paymentAddress
```
Elland: This shouldn't be optional. If we don't have a payment address here we need to deal it in some way. Alert and retry?

Token/Controllers/Settings/Backup Phrase/Views/BackupPhraseView.swift
```diff
@@ -65,7 +65,7 @@ class BackupPhraseView: UIView {
                     self.shake()
                 }
             } else {
-                TokenUser.current!.verified = self.verificationStatus == .correct
+                TokenUser.current?.verified = self.verificationStatus == .correct
```
Elland: Parenthesis to clarify behaviour? 👍

Token/Controllers/Settings/ProfileEditController.swift
```diff
@@ -348,7 +348,7 @@ open class ProfileEditController: OverlayController, Editable {
         for item in self.dataSource.items {
             if item.fieldName == "username" {
                 if item.validate() {
-                    username = item.value as? String ?? TokenUser.current!.username
+                    username = item.value as? String ?? TokenUser.current?.username ?? ""
```
Elland: Triple coalescing is a bit hard to read, maybe some parenthesis?

yuliaveres: what is hard to read here ? ..

Token/APIClients/EthereumAPIClient.swift
yuliaveres: this should be optional + Maybe alert BUT why it should be missing at all

Elland: Well, current user is an optional. That's why it used a force-unwrap, the code assumes user here will never be nil, if it is, we have bigger problems. Of course, for the users, having a hard crash because of an internal error is very annoying, but not sure if an alert is better. Also we don't get reports from alerts? 🤔

Token/Controllers/Settings/ProfileEditController.swift
Elland: What is going on. Too much noise, with an optional cast, and then two calls to coalesce some values. Too much happening at once.

PR #322, opened 2017-06-26
Validate payment number string before multiplying

yuliaveres:


Token/Controllers/PaymentController.swift
```diff
@@ -135,6 +135,9 @@ extension PaymentController: UITextFieldDelegate {
             if let currencyString = (textField.text as NSString?)?.replacingCharacters(in: range, with: string) {
                 if let currencyValue = self.inputNumberFormatter.number(from: currencyString) {
                     let ether = EthereumConverter.localFiatToEther(forFiat: currencyValue)
+                    
+                    guard ether != NSDecimalNumber.notANumber else { return }
```
Elland: Is the `NSDecimalNumber` necessary or can we just say `ether != .notANumber`? Also I find the double negative here a bit confusing to follow, but I guess no easy way around that.

yuliaveres: checking

Token/Controllers/PaymentController.swift
```diff
@@ -136,10 +136,10 @@ extension PaymentController: UITextFieldDelegate {
                 if let currencyValue = self.inputNumberFormatter.number(from: currencyString) {
                     let ether = EthereumConverter.localFiatToEther(forFiat: currencyValue)
                     
-                    guard ether != NSDecimalNumber.notANumber else { return }
-                    
-                    self.valueInWei = ether.multiplying(byPowerOf10: EthereumConverter.weisToEtherPowerOf10Constant)
-                    self.etherAmountLabel.text = EthereumConverter.ethereumValueString(forEther: ether)
+                    if ether != .notANumber {
```
Elland: Maybe just an `if ether == .notANumber { return }` instead? Then there's no double negative.

yuliaveres: No return inside defer 😢

Toshi/Controllers/PaymentController.swift
```diff
@@ -136,7 +136,7 @@ extension PaymentController: UITextFieldDelegate {
                 if let currencyValue = self.inputNumberFormatter.number(from: currencyString) {
                     let ether = EthereumConverter.localFiatToEther(forFiat: currencyValue)
                     
-                    if ether != .notANumber {
+                    if ether.isANumber == true {
```
Elland: if `isANumber` is a bool, do we need to check against `true`? Just realised we were doing that.

Toshi/Controllers/PaymentController.swift
```diff
@@ -136,7 +136,7 @@ extension PaymentController: UITextFieldDelegate {
                 if let currencyValue = self.inputNumberFormatter.number(from: currencyString) {
                     let ether = EthereumConverter.localFiatToEther(forFiat: currencyValue)
                     
-                    if ether != .notANumber {
+                    if ether.isANumber {
```
Elland: 💪🏽

Toshi/Extensions/NSDecimalNumber+Additions.swift
```diff
@@ -28,6 +28,10 @@ public extension NSDecimalNumber {
     public var toHexString: String {
         return "0x\(BaseConverter.decToHex(self.toDecimalString).lowercased())"
     }
+    
+    public var isANumber: Bool {
+        return self != .notANumber
```
Elland: Same here. `return !.notANumber`. 👍

yuliaveres: come on
what does that mean?

yuliaveres: and you can't do this
`Type 'Bool' has no member 'notANumber'`

Elland: Simpler than `return self != self.notANumber`, no? That feels super redundant to me. Even with the implicit `self` there.

Elland: And it being hard to read either way, I guess this is not avoidable, it's the reason we extended the class anyway? 😅

yuliaveres: i think we achieved syntactical improvements :) i am fine with this, not sure about other people

PR #326, opened 2017-06-27
Fix hardcoded reputation bar layout; Resize according to screen

yuliaveres:


yuliaveres:
not_nice
nice

Toshi/Controllers/Apps/Views/ReputationView.swift
```diff
@@ -259,6 +259,16 @@ class ReputationBarView: UIView {
         ])
     }
 
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        
+        let barRightMargin = 11
```
Elland: Shouldn't this be happening through auto-layout? 🤔

yuliaveres: what this ? :)

Elland: Layout adjustments. 👍

yuliaveres: no, we should do this.

Elland:
Before we merge, can you check why testing failed? This usually means CI couldn't build, since we don't have any local tests.

PR #327, opened 2017-06-27
Refactor app navigation

yuliaveres:


yuliaveres:
- By using app navigator for viewControllers pushing and presentation, we ensure that app won't crash in case it is happening twice for some reason;

- Alert presentation is successful independent on current top controller, as previously if controller which presented alert is not in the view hierarchy anymore, alert was not shown and we received debugger warning:
`Warning: Attempt to present <UIAlertController: 0x103befb80> on <Toshi.ChatController: 0x104913800> whose view is not in the window hierarchy!`

- Navigator provides with easy access to app root controller, tab bar controller, current top controller and top non modal controller.

yuliaveres:
Also, Navigator prevent from the following crash:
https://fabric.io/bakken--bck2/ios/apps/com.bakkenbaeck.token/issues/594eda52be077a4dcc934d93?time=last-seven-days

Toshi/Controllers/TabBarController.swift
```diff
@@ -126,21 +124,6 @@ open class TabBarController: UITabBarController {
         }
     }
 
-    private func presentAddressChangeAlertIfNeeded() {
```
Elland: We're not longer showing this?

Toshi/Models/Navigator.swift
```diff
@@ -16,8 +16,76 @@
...
+        guard viewController.presentingViewController == nil else { return }
+        
+        if parentViewController?.presentedViewController != nil {
+            parentViewController?.presentedViewController?.dismiss(animated: animated, completion: completion)
+        }
+        
+        parentViewController?.present(viewController, animated: true, completion: nil)
+    }
+    
+    public static func presentAddressChangeAlertIfNeeded() {
```
Elland: Oh ok. Nice. 👍

Toshi/Protocols/PaymentPresentable.swift
```diff
@@ -72,17 +72,17 @@ extension PaymentPresentable where Self: UIViewController {
...
 
     func presentSuccessAlert(with actionBlock: ((UIAlertAction) -> Swift.Void)?) {
-        let alertControlelr = UIAlertController.init(title: "Done", message: "Payment succeeded", preferredStyle: .alert)
+        let alertController = UIAlertController.init(title: "Done", message: "Payment succeeded", preferredStyle: .alert)
         let action = UIAlertAction.init(title: "OK", style: .default) { action in
             actionBlock?(action)
         }
 
-        alertControlelr.addAction(action)
+        alertController.addAction(action)
```
Elland: 🕵🏽‍♂️

Toshi/Controllers/TabBarController.swift
yuliaveres: we do, from other place

PR #328, opened 2017-06-27
Alert users if sign out fails.

Elland:
Better than silent failure. Currently investigating how to avoid it altogether.


Toshi/Supporting Files/AppDelegate.m
```diff
@@ -141,8 +141,8 @@ - (void)userDidSignOut {
 
         exit(0);
     } failure:^(NSError *error) {
-        // alert user
-        NSLog(@"Error attempting to unregister text secure.");
```
yuliaveres: i think we should leave log no ?

Elland: Why? 🤔

yuliaveres: it is handy to see important issues logs in debugger, as for me

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -141,8 +141,8 @@ - (void)userDidSignOut {
...
+        UIAlertController *alert = [UIAlertController alertControllerWithTitle:@"Could not sign out" message:@"Error attempting to unregister from chat service. Ours engineers are looking into it." preferredStyle:UIAlertControllerStyleAlert];
```
yuliaveres: don't you need any action here ? 😄

yuliaveres: it is not dismissable prolly 😄

Elland: 🤦🏽‍♂️ used to the SweetUIKit one!

yuliaveres: btw, you can use handy UIAlertController dismissableAlert ...

Elland:
Changes are up. 🤞🏽

PR #331, opened 2017-06-27
Check if indexPath is valid when accessing chat buttons

yuliaveres:
Fixes invalid indexPath crash
https://fabric.io/bakken--bck2/ios/apps/com.bakkenbaeck.token/issues/59516ed8be077a4dcca7033d?time=last-twenty-four-hours


yuliaveres:
@roberthein , i didn't but in general it is good practise to always check indexPath 😉

Toshi/Extensions/ControlsCollectionViewDelegateDatasource.swift
```diff
@@ -59,10 +59,10 @@ class SubcontrolsViewDelegateDatasource: NSObject, UICollectionViewDataSource, U
     }
 
     func didTapButton(for cell: ControlCell) {
-        guard let indexPath = self.subcontrolsCollectionView?.indexPath(for: cell) else { return }
+        guard let indexPath = self.subcontrolsCollectionView?.indexPath(for: cell), indexPath.item != NSNotFound, indexPath.item < self.items.count else { return }
```
Elland: Should probably abstract this. Extension method?

yuliaveres: Extension for indexPath ?

Elland: To check for validity. `func isValid(for count: Int) -> Bool` or something.

Elland: What do you think? Since it's used in more than one place.

yuliaveres: yeah, 🤔  good idea i think

PR #336, opened 2017-06-29
Clean Contact cell on reuse, prevent from displaying false data

yuliaveres:


Toshi/Views/ContactCell.swift
```diff
@@ -74,6 +74,15 @@ class ContactCell: UITableViewCell {
 
         return view
     }()
+    
+    override func prepareForReuse() {
+        super.prepareForReuse()
```
Elland: And there's no async call to set the image that could cause it to override the right one with a stale call to update it?

yuliaveres: there is, but before image is downloaded, cell display previous user's at this index

PR #339, opened 2017-06-29
Fix inconsistent presentation of alerts while scanning qr codes

yuliaveres:


Elland:
Description, screenshots or gif or video? 👍

yuliaveres:
just made 😄

yuliaveres:
previous
now

yuliaveres:
yep , basically payment confirmation was not dismissed completely when Navigator started to present alert, thus topController was still PaymentConfirmation.
as a result confirmation was dismissed shortly and error alert was never shown 😢

PR #340, opened 2017-06-29
Send an empty SOFA Message along with attachments

tristan:
The currently android build depends on attachment messages including a SOFA Message for it to store the attachments in it's history. If no SOFA Message is present, then the attachment gets lost next time the android client draws the chat.

This is probably more an android bug to fix, or a deeper issue with the sofa spec/signal mix that needs to be sorted out, but this quick fix keeps things sane while we sort out the deeper issues.

@yuliaveres Since i'm such a swift newbie, I just put the SOFA message in directly as a string. I don't know if there's a more standard way to generate a SOFA:Message here, but if there is feel free to modify this as you see fit.
Also, if you know of any other places where empty messages are sent with other attachments, we might need to change them there too.


Elland:
Good stuff you two! @tristan @yuliaveres

PR #346, opened 2017-07-04
Disable autocapitalisation on the seed input field

tristan:
The seed is always all lowercase letters. It's a pain in the arse to have to toggle the first letter to force it to be lower case.


Toshi/Controllers/Sign In/SignInController.swift
```diff
@@ -45,7 +45,9 @@ open class SignInController: UIViewController {
     }()
 
     private lazy var passwordField: InputField = {
-        InputField(type: .password)
+        let field = InputField(type: .password)
+        field.textField.autocapitalizationType = UITextAutocapitalizationType.none
```
Elland: `UITextAutocapitalizationType` is redundant. You can just say:

```swift
field.textField.autocapitalizationType = .none
```
✨

We do a newline before returns, for easier scanning. 👍  

(no worries, we can change this. thanks for the improvement 🚀)

yuliaveres: pushing this @Elland

tristan:
Along the same lines, i considered adding
```
field.textField.autocorrectionType = .no
```
to disable the auto correction. 
My concern there is that it might be that after you've typed your seed in enough times the next word suggestion thing would learn your seed and people would be able to extract it using the suggestions.
But I think since they're all normal words it is kinda nice to have, and I think the barrier to finding the first word is high enough that it's probably ok.

Toshi/Controllers/Sign In/SignInController.swift
tristan: i still have the code open, i can fix that

yuliaveres: and cause you commented, you also need to approve :/

PR #349, opened 2017-07-04
Add existing attributes to EthereumConverter Attributed string method

marijnschilling:
You can add the existing attributes of the label as an optional parameter to the `balanceSparseAttributedString` function.

And add tests for it, and make the test target work 😅


Toshi/Controllers/Settings/Settings.storyboard
```diff
@@ -42,7 +42,7 @@
                                                     </constraints>
                                                 </imageView>
                                                 <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="@robertv" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="qXq-f8-Bm1">
-                                                    <rect key="frame" x="91" y="48" width="79.5" height="20"/>
```
yuliaveres: hmm, do we need those changes in this PR ? and why ?

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -203,9 +206,9 @@ open class SettingsController: UITableViewController {
     // MARK: TableView methods
 
     /// This handles the actions for cell selection.
-    ///
+    ///
```
yuliaveres: nope 😄

Toshi/Views/ChatCell.swift
```diff
@@ -135,6 +135,7 @@ class ChatCell: UITableViewCell {
     lazy var lastMessageLabel: UILabel = {
         let view = UILabel()
         view.numberOfLines = 2
+        view.font = Theme.regular(size: 15)
```
yuliaveres: why u added this change here ? is that related to the PR?

yuliaveres:
Okay, is that UI improvements pr ?

Toshi/Views/ChatCell.swift
marijnschilling: Yeah i did this list of 'small' changes, but one simple font change needed this bigger change of the EthereumConverter, and then it needed tests, and then it needed to be able to run the test target.

yuliaveres: @marijnschilling , maybe just rename pr then 😉

marijnschilling: Well I might want to remove the other font changes, that would make more sense. But that would involve a lot of git magic that i'm really not up for after all this merge conflict drama 🙈  There's way more (hopefully) small UI tweaks that I need to fix as well after this!

Toshi/Controllers/Settings/SettingsController.swift
marijnschilling: This is swiftFormat, don't you have the same?

Toshi/Controllers/Settings/Settings.storyboard
marijnschilling: I did not change anything here, just the font being used. I will revert it back to the old values!

PR #350, opened 2017-07-04
Adds attributes of label to the attributed String method of EthereumConverter

marijnschilling:
You can give the attributes that are existing on a UILabel as an optional parameter on this method, so that they are applied to the attributed string returned from the `balanceSparseAttributedString` method.


Toshi/Controllers/Settings/Settings.storyboard
```diff
@@ -42,7 +42,7 @@
                                                     </constraints>
                                                 </imageView>
                                                 <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="@robertv" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="qXq-f8-Bm1">
-                                                    <rect key="frame" x="91" y="48" width="79.5" height="20"/>
+                                                    <rect key="frame" x="91" y="48" width="83" height="20"/>
```
yuliaveres: are those changes relevant here ?

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -203,9 +206,9 @@ open class SettingsController: UITableViewController {
     // MARK: TableView methods
 
     /// This handles the actions for cell selection.
-    ///
+    /// 
     /// There's unfortunatelly way to directly add IBActions for touching a cell.
-    ///
+    ///
```
yuliaveres: remove whitespaces, please, in all places

Toshi/Models/EthereumConverter.swift
```diff
@@ -89,20 +89,21 @@ struct EthereumConverter {
     /// Complete formatted string value for a given wei, with fiat aligned left and eth aligned right.
     ///    "$4.99 USD                        0.0050 ETH"
     ///    Fiat is black, and eth value is light grey.
-    ////
+    /// /
```
yuliaveres: ? 😄

Toshi/Models/EthereumConverter.swift
```diff
@@ -111,20 +112,23 @@ struct EthereumConverter {
     /// Complete formatted string value for a given wei, fully left aligned.
     ///    "$4.99 USD    0.0050 ETH"
     ///    Fiat is black, and eth value is light grey.
-    ////
+    /// /
```
yuliaveres: 😟

Toshi/Controllers/Settings/Settings.storyboard
```diff
@@ -115,10 +115,15 @@
...
                                             <autoresizingMask key="autoresizingMask"/>
                                             <subviews>
-                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="$20.00 USD" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="x2z-Rp-bD6">
-                                                    <rect key="frame" x="20" y="11" width="335" height="21"/>
-                                                    <fontDescription key="fontDescription" type="system" pointSize="17"/>
-                                                    <nil key="textColor"/>
+                                                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" usesAttributedText="YES" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="x2z-Rp-bD6">
```
yuliaveres: i dnt think you need to make it attributed here in storyboard 🤔  ?

marijnschilling: No you are right! Don't need it. I thought we would need it since I take the attributes from this label and put the on the attr string. But apparently just a custom string works as well ✌️

Toshi/Models/EthereumConverter.swift
marijnschilling: 🤔

PR #351, opened 2017-07-05
UI tweaks

marijnschilling:
Updates a bunch of UI things fromt his card: https://trello.com/c/FgP0GhTz/70-ui-alignment-texts-and-fonts


Toshi/Controllers/PaymentConfirmationController.swift
```diff
@@ -32,7 +32,7 @@ class PaymentConfirmationController: AlertController {
     }()
 
     lazy var textLabel: UILabel = {
-        let view = TextLabel("How would you rate your experience with this app?")
```
yuliaveres: ohh, that was copy pasted here, you can just remove `textLabel` from this controller.
please do this

Toshi/Controllers/PaymentConfirmationController.swift
```diff
@@ -32,7 +32,7 @@ class PaymentConfirmationController: AlertController {
...
+        let view = TextLabel(String(format: NSLocalizedString("How would you rate %@", comment: ""), self.userInfo.username ?? "your experience"))
```
yuliaveres: maybe we should use displayName here instead of username ?

Toshi/Controllers/Rate Users/RateUserController.swift
```diff
@@ -45,7 +45,7 @@ class RateUserController: ModalPresentable {
     }()
 
     lazy var textLabel: UILabel = {
-        let view = TextLabel("How would you rate your experience with this app?")
+        let view = TextLabel(String(format: NSLocalizedString("How would you rate %@", comment: ""), self.user.username))
```
yuliaveres: maybe we should use displayName here instead of username ?

Toshi/Controllers/PaymentConfirmationController.swift
marijnschilling: Oke so just use a UILabel with this text and set the font and stuff?

yuliaveres: no no , just remove this label at all from the class 😉

Toshi/Controllers/Settings/Settings.storyboard
```diff
@@ -42,7 +42,7 @@
                                                     </constraints>
                                                 </imageView>
                                                 <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="@robertv" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="qXq-f8-Bm1">
-                                                    <rect key="frame" x="91" y="48" width="79.5" height="20"/>
```
yuliaveres: 😩

PR #356, opened 2017-07-05
Unify background blur

marijnschilling:
✨  Blurs are unified in a BlurView ✨


faymontage:
Could you add instructions on how to trigger/see this?

marijnschilling:
img_0487
Blur 1: Chats navigation controller
Blur 2: Chats floating header view (one displaying your balance)
Blur 3: Chat input text panel

Toshi/Views/BlurView.swift
```diff
@@ -0,0 +1,42 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+class BlurView: UIVisualEffectView {
+
+    lazy var dimming: UIView = {
```
yuliaveres: why not private(set) ?

Toshi/Controllers/Messaging/ChatsNavigationController.swift
```diff
@@ -35,17 +35,11 @@ public class ChatsNavigationController: UINavigationController {
...
         view.translatesAutoresizingMaskIntoConstraints = false
         view.isUserInteractionEnabled = false
-
-        let dimming = UIView()
-        dimming.backgroundColor = Theme.messageViewBackgroundColor.withAlphaComponent(0.6)
-        dimming.isUserInteractionEnabled = false
-        view.addSubview(dimming)
-
-        dimming.edges(to: view)
+        view.dimming.isUserInteractionEnabled = false
```
yuliaveres: we need to set this on both view and dimmingView ?

Toshi/Views/BlurView.swift
yuliaveres: and lets call it dimmingView ?

Toshi/Controllers/Messaging/ChatsNavigationController.swift
```diff
@@ -35,17 +35,11 @@ public class ChatsNavigationController: UINavigationController {
         self.tabBarItem.titlePositionAdjustment.vertical = TabBarItemTitleOffset
     }
 
-    lazy var backgroundBlur: UIVisualEffectView = {
-        let view = UIVisualEffectView(effect: UIBlurEffect(style: .extraLight))
+    lazy var backgroundBlur: BlurView = {
+        let view = BlurView()
         view.translatesAutoresizingMaskIntoConstraints = false
```
yuliaveres: u don't need it here anymore since u set it inside the initialiser

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
```diff
@@ -77,16 +77,10 @@ class ChatsFloatingHeaderView: UIView {
         return view
     }()
 
-    lazy var backgroundBlur: UIVisualEffectView = {
-        let view = UIVisualEffectView(effect: UIBlurEffect(style: .extraLight))
+    lazy var backgroundBlur: BlurView = {
+        let view = BlurView()
         view.translatesAutoresizingMaskIntoConstraints = false
```
yuliaveres: same here , not needed

Toshi/Views/BlurView.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    init() {
+        super.init(effect: UIBlurEffect(style: .extraLight))
+
+        translatesAutoresizingMaskIntoConstraints = false
+        addSubviewsAndConstraints()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
yuliaveres: I encourage to implement this since someone might want to use it in xib ?

Toshi/Controllers/Messaging/ChatsNavigationController.swift
marijnschilling: Hmmm yeah I wonder though what would be the best for that, maybe remove it from inside the DimingView init itself would be better, than when you use it you are more in control?

Toshi/Controllers/Messaging/ChatsNavigationController.swift
marijnschilling: I don't really know why this `isUserInteractionEnabled = false` is necessary in the first place. But it was like this in the original implementation on the Chats navigation controller, not sure what it's trying to solve or do.

yuliaveres: maybe now we have a chance to double check and remove odd one. since this property is enough to set on parent view right ?

Toshi/Controllers/Messaging/ChatsNavigationController.swift
yuliaveres: yes, i think if needed, `init(withAutoLayout: Bool)` might be used on any `UIView` right?

Toshi/Views/BlurView.swift
marijnschilling: Should i copy the code from my custom init in here then?

Elland: No copying please. Abstract it into a separate method, call from both places. 👍

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
yuliaveres:
```
required init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
        
        self.effect = UIBlurEffect(style: .extraLight)
        self.addSubviewsAndConstraints()
    }
```

Toshi/Views/BlurView.swift
yuliaveres:
```
required init?(coder aDecoder: NSCoder) {
        super.init(coder: aDecoder)
        
        self.effect = UIBlurEffect(style: .extraLight)
        self.addSubviewsAndConstraints()
    }
```

Toshi/Controllers/Messaging/ChatsNavigationController.swift
marijnschilling: Yup, I gave it some investigation! Just enabling interaction on dimingview is enough ✌️

Toshi/Views/BlurView.swift
marijnschilling: Wht would be a good name for such an abstract separate method?

Elland: I usually go with something like `setup()`. 👍

yuliaveres: come on. there is nothing to extract @Elland 
those initializers are different, until you recommend to extract `addSubviewsAndConstraint()` call into `setup` method

Elland: @yuliaveres if they're different, no need indeed. 💪🏽

PR #357, opened 2017-07-06
Fix caret overlap

marijnschilling:
💚🥕💚
 screenshot 2017-07-06 14 34 26

I'm just wondering if we shouldn't remove the placeholder when the textField becomes active so to have either the caret or the placeholder.


faymontage:
@marijnschilling as far as I know it should only disappear on typing, not on focus

marijnschilling:
I'm overthinking it 👌 😅

PR #359, opened 2017-07-06
Browse: Replace collective reloading with targeted reloading.

roberthein:
Fix for Browse refresh bug. Instead of reloading the whole collectionView only reload the cell for this particular data.


faymontage:
Example of this bug:

Toshi/APIClients/AppsAPIClient.swift
```diff
@@ -79,26 +79,6 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
         }
     }
 
-    func downloadImage(for app: TokenUser, completion: @escaping (_ image: UIImage?) -> Void) {
```
marijnschilling: This function wasn't used?

roberthein: This function wasn't used.

PR #362, opened 2017-07-07
Refactor reputation score to average rating

tristan:
Design decided that we should show the average rating instead of the reputation score, but keep the sorting order by reputation score. For this I added an `average_rating` field to the backend and refactored the field names from the reputation service to match the ones used on the id service.

This PR makes these same changes for the iOS client.


yuliaveres:
Thank you @tristan

PR #369, opened 2017-07-13
Fix bot wake up

tristan:
Fixes a regression where the app no longer sends the wake up message to bots.

as always, i have no idea what i'm doing so feel free to just dump this PR and fix it properly in your main branch.


yuliaveres:
Sure , thx @tristan  😸

PR #371, opened 2017-07-13
Active network view

marijnschilling:
Show on new payment and new request:
request and pay

Show when receiving a payment request in `ChatController`:
payment request

Show on Payment Confirmation:
activenetworkscan


Toshi/Theme/Theme.swift
```diff
@@ -120,6 +120,10 @@ extension Theme {
         return UIColor(white: 0.95, alpha: 1)
     }
 
+    public static var greyBackgroundColor: UIColor {
```
yuliaveres: I really believe we have this color already. 
Also maybe it is worth rethink naming for colors in this file?
(I would suggest trying `darkTextColor,` it is called this way now ¯\_(ツ)_/¯ )

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,54 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import TinyConstraints
+import SweetUIKit
+
+class ActiveNetworkView: UIView {
+    static let height: CGFloat = 32.0
+    fileprivate static let margin: CGFloat = 6.0
```
yuliaveres: why it is static ?

marijnschilling: Because it's a constant, in swift you use static for that right?

Toshi/Theme/Theme.swift
marijnschilling: I looked for the hex number, it wasn't in the Theme file. Maybe it's one of these whiteColor function ones though, but not sure how to find that out 🤔

Toshi/Views/ActiveNetworkView.swift
yuliaveres: it is used only in this file

Toshi/Theme/Theme.swift
yuliaveres: no, just check `darkTextColor` if that is one

yuliaveres: okay, i checked, it is same as `mediumTextColor`,
so please, use this one

marijnschilling: ah cool thanks!

Toshi/Views/ActiveNetworkView.swift
marijnschilling: yes, good point 🙃  I will make that a normal let :)

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,94 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import TinyConstraints
+import SweetUIKit
+
+protocol ActiveNetworkViewDelegate: class {
+    func updateLayoutRequest(from: ActiveNetworkView)
```
yuliaveres: the title makes no sense, "update request" ?

Toshi/Controllers/PaymentConfirmationController.swift
```diff
@@ -25,10 +26,11 @@ class PaymentConfirmationController: AlertController {
 
     fileprivate var review: String = ""
 
-    lazy var titleLabel: TitleLabel = {
-        let view = TitleLabel("Rate \(String(describing: self.userInfo.username))")
+    fileprivate lazy var activeNetworkView: ActiveNetworkView = {
```
yuliaveres: it would be much better to keep active network view support in a protocol.
currently, if we are going to introduce ANV in any other controller, we will have to add this lazy property and constraints activation to its class, again. apart from 2 already existing places.

in case of protocol implementation, there would be one place where the same view is created. also each controller should just return  ACV constraints and those should be activated inside protocol, in one place

marijnschilling: But the chat view controller has different constraints, and needs to adjust the tableview insets. Where the scan controller does not.

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+
+        setupView()
+    }
+
+    private func setupView() {
+        backgroundColor = Theme.mediumTextColor.withAlphaComponent(0.98)
```
marijnschilling: Looks better.

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,94 @@
...
+
+class ActiveNetworkView: UIView {
+    weak var delegate: ActiveNetworkViewDelegate?
+
+    fileprivate let height: CGFloat = 32.0
+    fileprivate let margin: CGFloat = 6.0
+
+    var heightConstraint: NSLayoutConstraint?
+
+    fileprivate var isDefaultNetworkActive = NetworkSwitcher.shared.isDefaultNetworkActive
```
marijnschilling: Because i need it to check wether it's necessary to do a hide() or a show()

Toshi/Protocols/PaymentPresentable.swift
```diff
@@ -27,6 +27,8 @@ protocol PaymentPresentable: class {
 
     func paymentDeclined()
     func paymentApproved(with parameters: [String: Any], userInfo: UserInfo)
+
+    func setStatusBarHidden(_ bool: Bool)
```
marijnschilling: Yesterday you told me i should try to only remove it when this confirmation view is shown.

Toshi/Controllers/PaymentConfirmationController.swift
yuliaveres: so what is the problem, you can have an optional method for any additional ui updates;
yes, each controller should return those constraints, which are needed. and activate them inside setup method, when the view is added to the hierarchy

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -922,7 +936,10 @@ extension ChatController: PaymentSendControllerDelegate {
         guard let valueInWei = valueInWei else { return }
 
         showActivityIndicator()
+        showActiveNetworkView()
+
         viewModel.interactor.sendPayment(in: valueInWei)
+
```
yuliaveres: remove please 😉

Toshi/Controllers/Settings/ProfileVisibility.storyboard
```diff
@@ -31,7 +31,7 @@
                                         <rect key="frame" x="0.0" y="35" width="375" height="44"/>
                                         <autoresizingMask key="autoresizingMask"/>
                                         <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" tableViewCell="fkb-aM-4DO" id="d0B-8I-A7v">
-                                            <rect key="frame" x="0.0" y="0.0" width="375" height="43.5"/>
```
yuliaveres: we agreed we should not commit those changes?

Toshi/Controllers/Settings/Settings.storyboard
```diff
@@ -179,15 +179,15 @@
                             </tableViewSection>
                             <tableViewSection headerTitle="Settings" id="ZUh-Zh-8AK">
                                 <cells>
-                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" selectionStyle="none" hidesAccessoryWhenEditing="NO" indentationLevel="1" indentationWidth="0.0" id="dGU-NG-AvB">
+                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" selectionStyle="none" accessoryType="disclosureIndicator" hidesAccessoryWhenEditing="NO" indentationLevel="1" indentationWidth="0.0" id="dGU-NG-AvB">
```
yuliaveres: same here

Toshi/Controllers/Sign In/Views/PaymentRequestInfoView.xib
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="12120" systemVersion="16E195" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
```
yuliaveres: same here

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,113 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import TinyConstraints
+import SweetUIKit
+
+protocol ActiveNetworkDisplaying: class {
```
yuliaveres: we should keep protocols separately, in Protocols folder

Toshi/Views/ActiveNetworkView.swift
yuliaveres: i mean you can just access `NetworkSwitcher.shared.isDefaultNetworkActive` ..
the same way you do with `NetworkSwitcher.shared.activeNetworkLabel`

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -306,19 +312,26 @@ final class ChatController: OverlayController {
...
         view.addSubview(ethereumPromptView)
 
         ethereumPromptView.heightAnchor.constraint(equalToConstant: ChatsFloatingHeaderView.height).isActive = true
         ethereumPromptView.topAnchor.constraint(equalTo: topLayoutGuide.bottomAnchor).isActive = true
         ethereumPromptView.leftAnchor.constraint(equalTo: view.leftAnchor).isActive = true
         ethereumPromptView.rightAnchor.constraint(equalTo: view.rightAnchor).isActive = true
 
-        view.addSubview(tableView)
+        view.addSubview(activeNetworkView)
```
yuliaveres: you are adding it inside protocol method default implementation `setupActiveNetworkView `
which is super cool,
why adding here and not using `setupActiveNetworkView ` .

just call it here 😉

Toshi/Controllers/PaymentConfirmationController.swift
```diff
@@ -54,14 +53,23 @@ class PaymentConfirmationController: AlertController {
         super.viewDidLoad()
 
         self.setupCustomContentView()
+
+        view.addSubview(activeNetworkView)
```
yuliaveres: the same about adding view, it is done inside protocol, so just call this method here.

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,113 @@
...
+    func hideActiveNetworkView()
+
+    func updateLayoutRequest()
+}
+
+extension ActiveNetworkDisplaying where Self: UIViewController {
+    func setupActiveNetworkView() {
+        guard let activeNetworkView = self.activeNetworkView as ActiveNetworkView? else { return }
+
+        self.view.addSubview(activeNetworkView)
```
yuliaveres: nice method ! love it ❤️  😻 😍 ,
but the perfection would be if you add another `var activeNetworkViewConstraints: [NSLayoutConstraints] { get }` here and return those from each conforming controller.
then use this variable here in this method.

doing this way in each conforming controller you will just say : `self. setupActiveNetworkView()` 🌟 ⭐️ 🌟 

✨

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -306,19 +312,26 @@ final class ChatController: OverlayController {
...
 
-        tableView.topAnchor.constraint(equalTo: ethereumPromptView.bottomAnchor).isActive = true
-        tableView.left(to: view)
-        tableView.right(to: view)
-        tableView.bottomToTop(of: controlsView)
+        activeNetworkView.topAnchor.constraint(equalTo: ethereumPromptView.bottomAnchor, constant: -1).isActive = true
```
yuliaveres: return this inside `activeNetworkViewConstraints ` and let protocol care about adding view and managing its constraints, you just tell which one

Toshi/Controllers/PaymentConfirmationController.swift
```diff
@@ -54,14 +53,23 @@ class PaymentConfirmationController: AlertController {
...
+
+        activeNetworkView.top(to: view)
```
yuliaveres: SAME HERE: 

return this inside activeNetworkViewConstraints and let protocol care about adding view and managing its constraints, you just tell which one

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    func showActiveNetworkView() {
+        guard !NetworkSwitcher.shared.isDefaultNetworkActive else { return }
+
+        activeNetworkView.heightConstraint?.constant = self.activeNetworkView.height
+        updateLayoutRequest()
+    }
+
+    func hideActiveNetworkView() {
```
yuliaveres: also please rename method to `hideActiveNetworkViewIfNeeded` cause it doesn't hide it in 100% cases

Toshi/Views/ActiveNetworkView.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    func defaultActiveNetworkView() -> ActiveNetworkView {
+        let activityIndicator = ActiveNetworkView(withAutoLayout: true)
+
+        return activityIndicator
+    }
+
+    func showActiveNetworkView() {
+        guard !NetworkSwitcher.shared.isDefaultNetworkActive else { return }
```
yuliaveres: same here about naming

Toshi/Views/ActiveNetworkView.swift
faymontage: Even though it's such a tiny thing - it really should come from the design spec. Can you get Robert's input?

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -254,6 +254,8 @@ open class SettingsController: UITableViewController {
         switch balanceCellType {
         case .addMoney:
             self.pushAddMoneyController()
+        case .balance:
+          self.pushAddMoneyController()
```
yuliaveres: 🐛

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -126,6 +130,8 @@ final class ChatController: OverlayController {
         view.keyboardDismissMode = .interactive
         view.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: 10, right: 0)
 
+//        view.contentInset = UIEdgeInsets(top: ChatsFloatingHeaderView.height + 64.0, left: 0, bottom: 0, right: 0)
```
yuliaveres: @marijnschilling  should be back? otherwise remove it, please

yuliaveres: no it should be uncommented, table view is underneath now.

PR #372, opened 2017-07-13
Network persistency

yuliaveres:


Toshi/Network/NetworkSwitcher.swift
```diff
@@ -0,0 +1,129 @@
...
+
+public final class NetworkSwitcher {
+
+    static let shared = NetworkSwitcher()
+    fileprivate let keychain = KeychainSwift()
+
+    init() {
+        self.keychain.synchronizable = false
+
+        NotificationCenter.default.addObserver(self, selector: #selector(userDidSignOut(_:)), name: .UserDidSignOut, object: nil)
```
marijnschilling: I like to add notification subscriptions in a separate method. But since its only one might be unneseccary, but in case it would get more...

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    func setActiveNetwork(with index: Int) {
+        guard let validNetwork = Network(rawValue: index) as Network? else { return }
+
+        self.cachedActiveNetwork = validNetwork
+        self.keychain.set(validNetwork.ID, forKey: NetworkInfo.ActiveNetwork)
+    }
+
+    var isDefaultNetworkActive: Bool {
```
marijnschilling: These public methods can go above the private methods, so we can clearly see what the interface of this class is. ✌️

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -0,0 +1,129 @@
...
+
+import Foundation
+import KeychainSwift
+
+enum NetworkInfo {
+
+    static let ActiveNetwork = "ActiveNetwork"
+
+    struct ID {
+        static let Distribution = "3"
```
yuliaveres: i am not sure if proper network names should be localized

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -0,0 +1,145 @@
...
+    static let shared = NetworkSwitcher()
+    fileprivate let keychain = KeychainSwift()
+
+    init() {
+        self.keychain.synchronizable = false
+
+        NotificationCenter.default.addObserver(self, selector: #selector(userDidSignOut(_:)), name: .UserDidSignOut, object: nil)
+    }
+
+    fileprivate var activeNetwork: Network {
```
marijnschilling: I would need this method to be public so can get the active network enum val in my `NetworkSettingsController` and map it to the rows ✌️ 

```
let currentNetwork = NetworkSwitcher.shared.current
selectedIndexPath = IndexPath(row: currentNetwork.rawValue, section: 0)
```

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -0,0 +1,145 @@
...
+        guard let cachedNetwork = self.cachedActiveNetwork as Network? else {
+            return self.storedNetwork
+        }
+
+        return cachedNetwork
+    }
+
+    func setActiveNetwork(with index: Int) {
```
marijnschilling: I would like to use this method just with a `Network` type from my networkswitching controller
```
func setActiveNetwork(_ : Network)
```

Toshi/Network/NetworkSwitcher.swift
marijnschilling: I need to reach this var in my network settings controller, so i can map the selected network cell to it:
```
let currentNetwork = NetworkSwitcher.shared.current
selectedIndexPath = IndexPath(row: currentNetwork.rawValue, section: 0)
```

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    var activeNetworkBaseUrl: String {
+        return self.activeNetwork.baseURL
+    }
+
+    var activeNetworkID: String {
+        return self.activeNetwork.ID
+    }
+
+    fileprivate var defaultNetwork: Network {
```
marijnschilling: I also need to reach this var from my settings controller to default my network value to it:
```
fileprivate var network = NetworkSwitcher.shared.defaultNetwork
```

Toshi/Network/NetworkSwitcher.swift
yuliaveres: why ? noone else should create Network object apart from **NetworkSwitcher**;
you should tell him only which index have been selected (`indexPath.row` in your case)

Toshi/Network/NetworkSwitcher.swift
yuliaveres: noone else should know anything about `defaultNetwork`;
you just ask him for **activeNetwork** itself;

if no persistence yet, he will figure out default one himself and return it, don't worry ✋

Toshi/Network/NetworkSwitcher.swift
yuliaveres: also i will add in future option to select network with a given **label** as well, if needed.
for now there is no usage for it

Toshi/Network/NetworkSwitcher.swift
faymontage: The proper noun 'Ropsten' is the only canonical part of the name, the rest are just adjectives and we should keep the door open to localisation (though there's no plan to translate network names)

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -36,40 +31,37 @@ public class EthereumAPIClient: NSObject {
...
+    }
 
-                    completion(String(timestamp))
-                case .failure(_, _, let error):
-                    print(error)
-                }
-            }
-        }
+    fileprivate override init() {
+        self.mainTeapot = Teapot(baseURL: URL(string: EthereumAPIClient.teapotUrl)!)
```
faymontage: Bug: Switch to non-default network, relaunch app. 

self.mainTeapot is initialized using `NetworkSwitcher.shared.activeNetworkBaseUrl` which points to the switched network.

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -31,7 +31,7 @@ enum NetworkInfo {
 
     struct Path {
         static let RopstenTestNetwork = "https://ethereum.service.tokenbrowser.com"
-        static let ToshiTestNetwork = "https://token-eth-service.herokuapp.com"
+        static let ToshiTestNetwork = "https://token-eth-service-development.herokuapp.com"
```
faymontage: Correct 👌

PR #373, opened 2017-07-13
ChatController cleanup

marijnschilling:
Cleanup 🛁 

- [x] code order
- [x] swift format


yuliaveres:
if we should do reorder according to decided guidelines, we should do this once and with everything, not to tackle merge issues in future for multiple times.
Chat is the largest so far. But it is worth include other controllers as well here, if you decided to do this.

marijnschilling:
I could restructure more classes, but I think it's better to make the process of making improvements as flexible as possible and so to merge this already even when I haven't gotten to restructure the other classes. Now i only changed one class the merge conflicts will be only easier to maintain right? Or do i miss something here?

yuliaveres:
Ok, if that would be better for you guys, i am for it.

PR #375, opened 2017-07-17
[WIP] Fix loading of reputation score

roberthein:
Replace loading of reputation with model data


Toshi/Controllers/Apps/Controllers/BrowseController.swift
```diff
@@ -209,8 +209,8 @@ extension BrowseController: UISearchBarDelegate {
                     cell.avatarImageView.setImage(from: AsyncImageURL(url: url))
                 }
                 
-                RatingsClient.shared.scores(for: item.address) { ratingScore in
-                    cell.ratingView.set(rating: Float(ratingScore.averageRating))
+                if let reputationScore = item.reputationScore {
+                    cell.ratingView.set(rating: Float(reputationScore))
```
tristan: @roberthein Is there any reason to having the value as a `Double` when it's loaded in the `TokenUser.swift` file and casting it here, rather that just loading it as a `Float` by default? Seems unnecessary to me to have to cast it every time it's used when there's no other uses than this use.

roberthein: I have no good reason for doing this 😞

tristan:
@roberthein something else that needs to change with this:
There is also `averate_rating` in the user object. This should be used for the value for the star ratings. not `reputation_score` as they're different things. see https://github.com/toshiapp/toshi-ios-client/pull/362 for reference

faymontage:
Added @yuliaveres for code review and myself for QA

faymontage:
@roberthein [WIP]  is removed but is tristan's average_rating vs reputation_score comment dealt with?

roberthein:
@faymontage No, that still has to be done.

tristan:
looks good now 👍

PR #377, opened 2017-07-18
Update profile edit

yuliaveres:


Toshi/Controllers/Settings/ProfileEditController.swift
```diff
@@ -79,13 +70,16 @@ open class ProfileEditController: OverlayController, Editable {
     }()
 
     fileprivate lazy var tableView: UITableView = {
-        let view = UITableView(withAutoLayout: true)
+        let view = UITableView.init(frame: self.view.frame, style: .grouped)
+        view.translatesAutoresizingMaskIntoConstraints = false
```
roberthein: I see you use `.init` 3 times in this file where you don't need it.

Toshi/Controllers/Settings/ProfileEditController.swift
```diff
@@ -79,13 +70,16 @@ open class ProfileEditController: OverlayController, Editable {
...
 
+        view.contentInset = UIEdgeInsets(top: 0.0, left: 0.0, bottom: 64.0, right: 0.0)
         view.backgroundColor = UIColor.clear
```
roberthein: You don't have to change it but I'm just curious.

Toshi/Controllers/Settings/ProfileEditController.swift
```diff
@@ -396,6 +404,38 @@ open class ProfileEditController: OverlayController, Editable {
         }
     }
 
+    fileprivate let itemAnyLength = -1
+    fileprivate func validateUserName(_ username: String) -> Bool {
```
roberthein: This `let` seems out of place and the naming is not clear to me. Can you think of a better name?

Toshi/Controllers/Settings/ProfileEditController.swift
```diff
@@ -344,27 +338,40 @@ open class ProfileEditController: OverlayController, Editable {
...
+        var isPublic = false
 
-        for item in self.dataSource.items {
-            if item.fieldName == "username" {
-                if item.validate() {
-                    username = item.value as? String ?? (TokenUser.current?.username ?? "")
-                } else {
-                    let alert = UIAlertController.dismissableAlert(title: "Error", message: "Username is invalid! Use numbers, letters, and underscores only.")
-                    Navigator.presentModally(alert)
+        let items = self.editingSections.flatMap { section in
```
roberthein: It might not be super clear to everyone that you are flattening an array of arrays here. What do you think?

roberthein: Maybe a different name for `items` can make this more clear?

Toshi/Controllers/Settings/ProfileEditController.swift
```diff
@@ -448,15 +488,34 @@ extension ProfileEditController: UITableViewDelegate {
...
+        
+        return editingSection.items.count
     }
 
     public func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeue(ProfileItemCell.self, for: indexPath)
-        let formItem = self.dataSource.item(at: indexPath)
-        cell.selectionStyle = .none
-        cell.formItem = formItem
+        guard let cell = tableView.dequeue(InputCell.self, for: indexPath) as InputCell? else { return UITableViewCell() }
```
roberthein: However this will never happen, returning an instance of `UITableViewCell` here is wrong. It's better to decouple the dequeueing from the type checking. Like this for example:

```
let cell = tableView.dequeue(InputCell.self, for: indexPath)
        
if let cell = cell as InputCell? {
    let section = self.editingSections[indexPath.section]
    let item = section.items[indexPath.row]
            
    let configurator = ProfileEditConfigurator(item: item)
    configurator.configure(cell: cell)
}
        
return cell
```

roberthein: Actually the type checking is not necessary at all, because you are using `SweetUIKit`'s convenience method.
So you can just do this:
```
let cell = tableView.dequeue(InputCell.self, for: indexPath)

let section = self.editingSections[indexPath.section]
let item = section.items[indexPath.row]

let configurator = ProfileEditConfigurator(item: item)
configurator.configure(cell: cell)

return cell
```

Toshi/Controllers/Settings/Views/ProfileEditConfigurator.swift
```diff
@@ -0,0 +1,104 @@
...
+    var footerTitle: String?
+
+    init(items: [ProfileEditItem], headerTitle: String? = nil, footerTitle: String? = nil) {
+        self.items = items
+        self.headerTitle = headerTitle
+        self.footerTitle = footerTitle
+    }
+}
+
+public class ProfileEditItem {
```
roberthein: Can this class have it's own file? Or make it a `struct` maybe? 😍

yuliaveres: nope, it would nt be efficient since struct / enum is copied not referenced. and having one updated here, inside configurator, makes no difference on EditController datasource items :shrug:

Toshi/Controllers/Settings/ProfileEditController.swift
yuliaveres: that is true 👍

Toshi/Controllers/Settings/ProfileEditController.swift
yuliaveres: hmm i will do my best

Toshi/Controllers/Settings/ProfileEditController.swift
yuliaveres: how can I help him/her, to whom that is not clear ?

Toshi/Controllers/Settings/Views/ProfileEditConfigurator.swift
roberthein: Right. I am used to having one class per file, what's your take on this?

Toshi/Controllers/Settings/ProfileEditController.swift
yuliaveres: i am just used to this ? :/

Toshi/Controllers/Settings/ProfileEditController.swift
yuliaveres: oh yeah

Toshi/Controllers/Settings/ProfileEditController.swift
roberthein: Not sure, maybe best to add a comment like:
`// we use flatmap here to create a normal array from our 2d array`
🤷‍♀️

Toshi/Controllers/Settings/ProfileEditController.swift
roberthein: #objectivecforlife 😉

Toshi/Controllers/Settings/Views/ProfileEditConfigurator.swift
yuliaveres: yeye , sure 😄

Toshi/Controllers/Settings/ProfileEditController.swift
yuliaveres: objc forever ☝️

PR #382, opened 2017-07-20
Messages UI improvements.

roberthein:
- Round corners based on previous and next message contact.
- Reinstate life size emoji.

img_2309

img_2310


Toshi/MessagesUI/Cells/MessagesBasicCell.swift
```diff
@@ -2,6 +2,12 @@ import Foundation
 import UIKit
 import TinyConstraints
 
+enum MessageCorners {
```
yuliaveres: By message corners I would expect to see a collection of corners like: [.topLeft, .topRight]
this enum declares MessageCornersType i would say

roberthein: Ok 👌

PR #385, opened 2017-07-21
Workaround for browse page loading issues

tristan:
So, the main thing i notice when my app starts, is that it pre loads ~220 avatars on load. most of these requests fail.

What i think is happening is that each time a new `Teapot` is created, it has to go through all the SSL handshakes or whatever it needs to initialise, and this requires a lot of open file descriptors, and eventually the system just runs out and starts failing.

This fixes it by keeping each teapot for each host open and reusing it for each avatar download request, so now with my ~220 avatar loads, i only create 3-4 `Teapot`s.

Why it's loading ~220 avatars and why `Teapot` can't handle this with a new one per avatar is left as an exercise for the reader. 😭 😭 😭


yuliaveres:
@tristan "Why it's loading ~220 avatars"  it is preloading avatars and caching them

tristan:
i'm curious about it though, because there aren't 220 users in my recent tab, or my favourites tab. maybe there's something left over in the database from old installs when i was bot spamming?

also, it seems i've added an extra space at the start of the `downloadAvatar` method, do you want to remove that too while you're cleaning things up @yuliaveres ?

btw, while this will fix the problem right now, i don't think we should count it as a 100% fix. it's very possible that every avatar being preloaded comes from a different base URL, in which case this fix wouldn't help at all.

PR #391, opened 2017-07-24
Add Danger configuration.

roberthein:
For now I added Danger with some default configuration. Let's all dig a little deeper into the possibilities to see where Danger can help us even more.


roberthein:
😏

PR #393, opened 2017-07-24
Add SwiftLint

roberthein:
Current configuration:
```
disabled_rules:
  - file_length
  - line_length
  - function_body_length
  - type_body_length
  - type_name
  - identifier_name
  - force_cast
  - force_try
  - force_unwrapping
  - function_parameter_count
  - empty_count
  - private_outlet
  - vertical_whitespace
  - cyclomatic_complexity
  - todo
  - nesting
opt_in_rules:
  - explicit_init
  - first_where
  - operator_usage_whitespace
  - overridden_super_call
  - prohibited_super_call
  - redundant_nil_coalescing
  - switch_case_on_newline
trailing_whitespace:
  ignores_empty_lines: true
```


marijnschilling:
@roberthein If we use swift lint does that mean we can add it as a plugin to XCode to show warnings for the violations in the IDE?

roberthein:
@marijnschilling SwiftLint is added to the project though CocoaPods and indeed generates issues in the IDE. ✨

.swiftlint.yml
```diff
@@ -0,0 +1,27 @@
+disabled_rules:
```
marijnschilling: Where can I find the default rule set for swift lint?

roberthein: First you need to install SwiftLint:
`brew install swiftlint`
Then you can ask SwiftLint it's rules in the terminal:
`swiftlint rules`

marijnschilling:
I have a couple of questions already I think about the rules.

Are the `disabled_rules:` the disabled by default? Or is it a choice to disable them by you?

I think it might help a lot if you put shortly in the comments why you chose disabled or opted in to rules. Could be as simple as: `We should discuss first`or something.

marijnschilling:
And another question is. I saw in the movie example that you can put in your file to not have libraries and podfiles linted. Shouldn't we add that as well?

.swiftlint.yml
```diff
@@ -0,0 +1,27 @@
...
+  - force_try
+  - force_unwrapping
+  - function_parameter_count
+  - empty_count
+  - private_outlet
+  - vertical_whitespace
+  - cyclomatic_complexity
+  - todo
+  - nesting
+opt_in_rules:
```
marijnschilling: I think this one should be added to disabled too: `private_over_fileprivate` We use fileprivate a lot right to be able to use var in extensions

.swiftlint.yml
```diff
@@ -0,0 +1,27 @@
...
+  - type_body_length
+  - type_name
+  - identifier_name
+  - force_cast
+  - force_try
+  - force_unwrapping
+  - function_parameter_count
+  - empty_count
+  - private_outlet
+  - vertical_whitespace
```
marijnschilling: I think this one should be enabled. It makes sure you never use more than one white line right? That is in our guideline.

.swiftlint.yml
```diff
@@ -0,0 +1,27 @@
...
+  - line_length
+  - function_body_length
+  - type_body_length
+  - type_name
+  - identifier_name
+  - force_cast
+  - force_try
+  - force_unwrapping
+  - function_parameter_count
+  - empty_count
```
marijnschilling: This sone is disabled by default right?

.swiftlint.yml
```diff
@@ -0,0 +1,27 @@
...
+  - private_outlet
```
marijnschilling: This one is disabled by default right? I actually think it might be interesting to add it. But maybe after discussing.

.swiftlint.yml
```diff
@@ -0,0 +1,27 @@
...
+opt_in_rules:
+  - explicit_init
+  - first_where
+  - operator_usage_whitespace
+  - overridden_super_call
+  - prohibited_super_call
+  - redundant_nil_coalescing
+  - switch_case_on_newline
+trailing_whitespace:
+  ignores_empty_lines: true
```
marijnschilling: I don't understand this.

.swiftlint.yml
roberthein: We only do this when it's necessary right now and the compiler will tell us when it's necessary. Also in Swift 4 `private` will be sufficient to use var in extensions too.

.swiftlint.yml
roberthein: True! Removed it.

.swiftlint.yml
roberthein: True! Removed it.

.swiftlint.yml
marijnschilling: Oke, so till then whenever we use a var in an extension it will be not marked when we mark it as fileprivate?

.swiftlint.yml
roberthein: Ok! Enabled it.

.swiftlint.yml
roberthein: The `trailing_whitespace` is enabled by default but you can add two configurations if you want:
- ignores_empty_lines
- ignores_comments
I think it's good practice to not remove the whitespace when the line is empty

.swiftlint.yml
roberthein: What do you mean with:
> it will be not marked when we mark it as fileprivate

marijnschilling: I mean swift lint won't wine about it when we declare a var as fileprivate if we use it in an extesion? But it we don't use it in an extension it will? And change it to private?

roberthein: Yes. Do you agree with this behaviour?

.swiftlint.yml
marijnschilling: aha, why would you want a whitespace on an empty line though? Or why do you think it is good practice to leave it?

roberthein: I think it's good practice to leave, because it's added by Xcode and it would just clutter our PR's.

Carthage/Build/iOS/EtherealCereal.framework/Headers/EtherealCereal.swift
```diff
@@ -51,7 +51,6 @@ public class EtherealCereal: NSObject {
         return self.ether.generatePublicKey(from: privateKeyData)
     }
 
-
```
marijnschilling: ✨ 👌 ✨

.swiftlint.yml
marijnschilling: yeah that sounds perfect! Was just wondering, wasnt really clear from the explanation they have with the rule.

roberthein:
@marijnschilling In reply to your other questions:
- The `disabled_rules` should only have the rules that we want to disable and that are enabled by default.
- I added comments to the disabled rules.
- I excluded all Carthage and Cocoapods dependencies.

PR #397, opened 2017-07-25
Sofa messages show keyboard

marijnschilling:
Hiding the keyboard:
hidekeyboard

Showing the keyboard (it's kinda hard to show off this functionality because i first need to hide it manually very quick):
showkeyboard

https://trello.com/c/EjDdZm19/180-use-showkeyboard-property-in-sofa-message


Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -797,7 +797,15 @@ extension ChatController: ChatViewModelOutput {
 
 extension ChatController: ChatInteractorOutput {
 
-    func didHandleSofaMessage(with buttons: [SofaMessage.Button]) {
+    func didHandleSofaMessage(with buttons: [SofaMessage.Button], showKeyboard: Bool?) {
```
yuliaveres: there is low chance it would be executed on main thread, 
buttons adjustment is switched async there.
please, switch to main thread here if you touch UI, and remove it from `adjustToNewButtons`.

marijnschilling: You mean i should make two different delegate methods?n Or what is `adjustToNewButtons`?

yuliaveres: NO, 
switch to main thread here as you need it here.
remove it from `adjustToNewButtons`

Toshi/Models/SofaWrapper.swift
```diff
@@ -77,6 +78,10 @@ open class SofaWrapper: SofaWrapperProtocol {
         guard let jsonObject = try? JSONSerialization.jsonObject(with: data, options: []) else { return [String: Any]() }
         guard let json = jsonObject as? [String: Any] else { return [String: Any]() }
 
+        if let showKeyboardValue = json["showKeyboard"] as? NSNumber {
```
marijnschilling: No it should be nil as well, check the card: https://trello.com/c/EjDdZm19/180-397-use-showkeyboard-property-in-sofa-message

3 different cases: true false and nil

yuliaveres: ok

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -797,8 +794,20 @@ extension ChatController: ChatViewModelOutput {
 
 extension ChatController: ChatInteractorOutput {
 
-    func didHandleSofaMessage(with buttons: [SofaMessage.Button]) {
-        self.buttons = buttons
+    func didHandleSofaMessage(with buttons: [SofaMessage.Button], showKeyboard: Bool?) {
+        DispatchQueue.main.async {
+            if let showKeyboard = showKeyboard {
+                if showKeyboard == true {
+                     DispatchQueue.main.asyncAfter(seconds: 0.1) {
```
marijnschilling: Only after a tiny delay can the inputField become first responder. I'm not sure why. Should i put a comment with this?

yuliaveres: hm, yep, put it

PR #399, opened 2017-07-26
Show display name instead of username

marijnschilling:
Deep inside the `ContactsManager` the `firstname` was returned to be the display name, while the `lastname` property contains the display name.

‼️ to test this on device I needed to reinstall the app, because it's all database related stuff.

on the TSContactThread this property was always wrong actually. When you open a new chat thread the username was used for the title. It _only_ went right when opening a chat from the ChatsController, here the `cachedDisplayName` (which links to this display property) get's overwritten by the name of the `TokenUser`. So this is why it showed the right name most of the time! That also explains why Tristan mentioned that it did show the display name when opening it from the `Favorites`, that is only because the `ChatController` was in view before and overriding the username on the contact thread with the display name. If you open the app on favorites, and start a chat from here it also displayed the wrong name.

I removed this override as well, since now the `cachedDisplayName` will already contain the write name.

namebug

Links to this issue: https://trello.com/c/PHOl784m/160-fix-user-display-name-being-twisted-with-username-on-chat-view-controller


faymontage:
Fresh install of v.1.2.5, chat with CoffeeBot. Thread title is blank. Leave thread and return, thread title is now CoffeeBot. Upgrade to #399, return to chat, thread title is blank. Leave chat and return, thread title remains blank.

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -217,7 +217,7 @@ final class ChatController: OverlayController {
         super.init(nibName: nil, bundle: nil)
 
         hidesBottomBarWhenPushed = true
-        title = thread.cachedContactIdentifier
+        title = thread.name()
```
marijnschilling: Ah so we just don't use the cached name anymore?

faymontage:
Seems that my reproduction has vanished!

Toshi/Controllers/Messaging/ChatController.swift
yuliaveres: cached contact identifier should not be used for chat title

PR #400, opened 2017-07-26
Messages-UI Improvements

roberthein:
This PR fixes:
- layout issues in the chat controller
- content insets in the table view
- wrong positioning of buttons and menus
- wrong bottom corner in case of grouped messages
- showing avatars for every message even when messages are grouped

img_2432

img_2433


Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -288,58 +260,36 @@ final class ChatController: OverlayController {
...
-        subcontrolsView.leftAnchor.constraint(equalTo: view.leftAnchor, constant: 16).isActive = true
-        subcontrolsView.bottomAnchor.constraint(equalTo: controlsView.topAnchor, constant: ChatController.buttonMargin).isActive = true
-        subcontrolsViewDelegateDatasource.subcontrolsCollectionView = subcontrolsView
+        subcontrolsViewHeight = subcontrolsView.height(0)
 
-        navigationItem.rightBarButtonItem = UIBarButtonItem(customView: avatarImageView)
+        ethereumPromptView.top(to: view, offset: 64)
+        ethereumPromptView.left(to: view)
+        ethereumPromptView.right(to: view)
+        ethereumPromptView.height(ChatsFloatingHeaderView.height)
```
marijnschilling: Nice much cleaner with TinyConstraints.

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -358,11 +308,10 @@ final class ChatController: OverlayController {
     }
 
     fileprivate func updateConstraints() {
-        textInputViewBottom.constant = heightOfKeyboard < -textInputHeight ? heightOfKeyboard + textInputHeight + buttonsHeight : 0
-
-        textInputViewHeight.constant = textInputHeight
+        textInputViewBottom?.constant = heightOfKeyboard < -textInputHeight ? heightOfKeyboard + textInputHeight + buttonsHeight : 0
```
marijnschilling: I find this line a bit hard to follow. Can you put these together in a bool:
```
let isKeyboardShown = heightOfKeyboard < -textInputHeight
``` 
*Not necessarily `isKeyboardShown` i'm not sure what is checked for here.

and put these together in one var: 
```
let totalHeight = heightOfKeyboard + textInputHeight + buttonsHeight
```
 *(but with a better name than `totalHeight` pls)

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -179,21 +140,10 @@ final class ChatController: OverlayController {
...
-    }()
-
-    fileprivate lazy var controlsViewHeightConstraint: NSLayoutConstraint = {
-        self.controlsView.heightAnchor.constraint(equalToConstant: 0)
-    }()
-
-    fileprivate lazy var subcontrolsViewHeightConstraint: NSLayoutConstraint = {
-        self.subcontrolsView.heightAnchor.constraint(equalToConstant: 0)
-    }()
+    private var textInputViewBottom: NSLayoutConstraint?
```
marijnschilling: Could you call all these `textInputViewBottomConstraint` again? 😬  At least I always expect something called `Height` to be a `CGFloat` and get confused when it turns out to be a Constraint 🙏

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -690,20 +621,41 @@ extension ChatController: UITableViewDataSource {
         return cell
     }
 
-    private func cornerType(for indexPath: IndexPath) -> MessageCornerType {
-        guard let currentMessage = viewModel.messageModels.element(at: indexPath.row) else { return .top }
-        guard let previousMessage = viewModel.messageModels.element(at: indexPath.row - 1) else { return .top }
+    private func positionType(for indexPath: IndexPath) -> MessagePositionType {
```
marijnschilling: I think this method could be much clearer if you don't use guards. i think guards are more for simple returns like if json is nil return. like guarding the function of what is def's needed.

I think its better to use if statements here. And wrap up all the checks you do in separae Bool let's with a clear name to define what they are checking for. This way you can also remove the code comments.

So something like this:
```
let isFirstMessage = viewModel.messageModels.element(at: indexPath.row - 1)
if isFirstMessage { etc...
```

maybe you could chop it up in separate functions `positionTypeForFirstMessage` etc.
At least this function needs rewriting to make it readable. You can commit the comments when the code is readable. Let me know if this sounds doable to you, i can help if it's not clear.

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -932,7 +884,7 @@ extension ChatController: ChatsFloatingHeaderViewDelegate {
 
     func messagesFloatingView(_: ChatsFloatingHeaderView, didPressPayButton _: UIButton) {
         view.layoutIfNeeded()
-        controlsViewHeightConstraint.constant = 0.0
+        controlsViewHeight?.constant = 0.0
```
marijnschilling: Same here, i don't get that its a Constraint when it ends in `Height`

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -1035,7 +989,7 @@ extension ChatController: ControlViewActionDelegate {
         subcontrolsViewDelegateDatasource.items = []
         viewModel.currentButton = nil
 
-        subcontrolsViewHeightConstraint.constant = 0
+        subcontrolsViewHeight?.constant = 0
```
marijnschilling: And same here

Toshi/Controllers/Messaging/ChatController.swift
marijnschilling:
```
    private func positionType(for indexPath: IndexPath) -> MessagePositionType {

        guard let currentMessage = viewModel.messageModels.element(at: indexPath.row) else {
            return .single
        }

        let previousMessage = viewModel.messageModels.element(at: indexPath.row - 1)
        if isCurrentMessageFirstMessage = previousMessage == nil {
            return positionType(forFirstMessage: currentMessage, atIndexPath: indexPath)
        }

        let nextMessage = viewModel.messageModels.element(at: indexPath.row + 1)
        if isCurrentMessageLastMessage = nextMessage == nil {
            return positionType(forLastMessage: currentMessage, atIndexPath: indexPath)
        }

        return positionType(forMiddleMessage: currentMessage, atIndexPath: indexPath)
    }

    private func positionType(forFirstMessage firstMessage: Message, atIndexPath indexPath: IndexPath) -> MessagePositionType {
        let nextMessage = viewModel.messageModels.element(at: indexPath.row + 1)

        if isCurrentMessageLastMessage = nextMessage == nil {
            return .single
        } else {
            return firstMessage.isOutgoing == nextMessage.isOutgoing ? .top : .single
        }
    }

    private func positionType(forMiddleMessage middleMessage: Message, atIndexPath indexPath: IndexPath) -> MessagePositionType {
        if currentMessage.isOutgoing != previousMessage.isOutgoing, currentMessage.isOutgoing != nextMessage.isOutgoing {
            // the previous and next messages are not from the same user
            return .single
        } else if currentMessage.isOutgoing == previousMessage.isOutgoing, currentMessage.isOutgoing == nextMessage.isOutgoing {
            // the previous and next messages are from the same user
            return .middle
        } else if currentMessage.isOutgoing == previousMessage.isOutgoing {
            // the previous message is from the same user but the next message is not
            return .bottom
        } else {
            // the next message is from the same user but the previous message is not
            return .top
        }
    }
```

marijnschilling: I was wrong, this is clearly the best way to do this! Good job 👏  😬

Toshi/MessagesUI/Cells/MessagesBasicCell.swift
```diff
@@ -79,43 +67,18 @@ class MessagesBasicCell: UITableViewCell {
         }
     }
 
-    var cornerType: MessageCornerType = .top {
+    var positionType: MessagePositionType = .single {
         didSet {
-            messageBorderImageView.image = cornerImage
+            let first = positionType == .single || positionType == .top
```
roberthein: done

Toshi/MessagesUI/Cells/MessagesBasicCell.swift
```diff
@@ -79,43 +67,18 @@ class MessagesBasicCell: UITableViewCell {
...
+            contentLayoutGuideTopConstraint?.constant = first ? 8 : 4
+            
+            let hidden = positionType == .top || positionType == .middle || isOutGoing
```
roberthein: done

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -121,15 +94,10 @@ final class ChatController: OverlayController {
...
         view.estimatedRowHeight = 64.0
-        view.allowsSelection = true
         view.dataSource = self
         view.delegate = self
-        view.tableFooterView = UIView()
         view.separatorStyle = .none
         view.keyboardDismissMode = .interactive
-        view.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: 10, right: 0)
-
-        view.contentInset = UIEdgeInsets(top: ChatsFloatingHeaderView.height + 64.0, left: 0, bottom: 0, right: 0)
```
roberthein: Calling it in `viewWillAppear()` fixes it. Thanks!

Toshi/MessagesUI/Images/ImagesViewControllerTransition.swift
```diff
@@ -71,13 +71,17 @@ class ImagesViewControllerTransition: NSObject, UIViewControllerAnimatedTransiti
...
-
-        print(imagesViewController)
-
+        
         guard let toCell = imagesViewController.collectionView.visibleCells.flatMap({ cell in cell as? ImageCell }).first(where: { imageCell in imageCell.frame.width != 0 }) else { return nil }
 
         return toCell.imageView
     }
+    
+    func cornerType(for messagesViewController: ChatController, _ imagesViewController: ImagesViewController) -> MessagesCornerType? {
```
roberthein: What would you suggest?

marijnschilling: I don't really know what it does 🙈  Can you explain?

Toshi/Controllers/Messaging/ChatController.swift
roberthein: This is not part of this PR.

Toshi/Controllers/Messaging/ChatController.swift
roberthein: 🤘

Toshi/Controllers/Messaging/ChatController.swift
roberthein: yes!

Toshi/MessagesUI/Images/ImagesViewControllerTransition.swift
marijnschilling: oke so you need it to determine what kind of corners you need for the transition from and to the imagesViewController to ChatController right? And that is based on the cornerType of the cell that is at the indexPath of the current indexPath of the imagesViewController.

What about:
`func cornerType(for messagesViewController: ChatController, atIndexPath indexPath: IndexPath) -> MessagesCornerType? {`

roberthein: `cornerType(for: messagesViewController: ChatController, at indexPath: IndexPath)`
is het geworden 👍

PR #401, opened 2017-07-26
I made the tests run!

marijnschilling:
- I added a whole bunch of classes to the test target (for me it's normal practise to add any class that is not a View or ViewController to the test target, because these should be tested 😏 )
-  I added the test target to the podfile. probably not all podfiles are necessary for the test target. But now the PAIClients are closely interwoven with some of the Podfiles. This structure will probably change as soon as I write tests for the APIClients with mocked data. But for now it's most important that the Test run, than we can start to make it better!
-  I made the test target aware of the bridging header for the Obj-C classes. (that was a tricky one to find)
- I added the Development Project as Host Application for the Test target. Not sure why this was necessary and if there are unwanted side-effects. But otherwise I got this error: https://stackoverflow.com/questions/44062784/linker-command-failed-with-exit-code-1-in-unit-test/45331293#45331293
- Oh yeah also moved the extension on UINotification+Name to a separate extension file, sine now it was added to a ViewController and we don't want to add those to the test target.

For now I made a simple test for the `getRate` on the EthereumClient. My plan is to dig a bit into mocking API calls and apply that to the tests. But for now this is a fairly innocent test to _actually_ go out to the server to prove that it works 💪

✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨
screenshot 2017-07-26 17 20 39
✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨✨


Tests/Tests.swift
```diff
@@ -1,11 +1,17 @@
...
+import UIKit
 
 class Tests: XCTestCase {
 
     func testExample() {
-        let balance = EthereumConverter.balanceAttributedString(forWei: NSDecimalNumber.zero, exchangeRate: 0.0005)
-        print(balance)
-        XCTAssertTrue(true)
+        let expect = self.expectation(description: "get ethereum rate")
+        EthereumAPIClient.shared.getRate { decimal in
```
roberthein: Yeah? `self`?

roberthein: 😉

Tests/Tests.swift
```diff
@@ -1,11 +1,17 @@
...
+            XCTAssertNotNil(decimal)
+            expect.fulfill()
+
+         }
+
+        self.waitForExpectations(timeout: 100)
+
```
roberthein: 🙅‍♂️

marijnschilling: 😬

PR #405, opened 2017-07-27
Fix navigate to url.

roberthein:
img_2430
img_2431


Toshi/Controllers/Browse/Controllers/BrowseController.swift
```diff
@@ -198,36 +198,36 @@ class BrowseController: SearchableCollectionController {
...
+            showOpenURLButton()
         } else {
             AppsAPIClient.shared.search(searchText) { apps, error in
                 if let error = error {
                     let alertController = UIAlertController.errorAlert(error as NSError)
                     Navigator.presentModally(alertController)
                 }
                 
-                self.searchResultView.searchResults = apps
+                self.searchResultView.searchResults = searchText.isURL ? [] : apps
```
yuliaveres: what is the point of checking isUrl ? it is in else block already .

roberthein: The idea was to double check because the search call is async. Should I remove it?

Toshi/Controllers/Browse/Controllers/BrowseController.swift
```diff
@@ -180,21 +253,26 @@ extension BrowseController: UISearchBarDelegate {
 
         if searchText.isEmpty {
             searchResultView.searchResults = []
+            hideOpenURLButtonIfNeeded()
         } else {
             AppsAPIClient.shared.search(searchText) { items, error in
```
yuliaveres: can we remove this, we perform search in reload(searchText..) right?

Toshi/Controllers/Browse/Controllers/BrowseController.swift
```diff
@@ -180,21 +253,26 @@ extension BrowseController: UISearchBarDelegate {
 
         if searchText.isEmpty {
             searchResultView.searchResults = []
+            hideOpenURLButtonIfNeeded()
```
yuliaveres: what if not empty and not url anymore ??

roberthein: Then it will show the search results, is it not?

Toshi/Controllers/Browse/Controllers/BrowseController.swift
roberthein: Absolutely right! Good catch!

Toshi/Controllers/Browse/Controllers/BrowseController.swift
roberthein: Removed it!

Toshi/Controllers/Browse/Controllers/BrowseController.swift
```diff
@@ -170,6 +195,54 @@ class BrowseController: SearchableCollectionController {
...
+    @objc
+    fileprivate func reload(searchText: String) {
+        
+        if searchText.isURL {
+            let title = NSAttributedString(string: searchText, attributes: openButtonAttributes)
+            openURLButton.setAttributedTitle(title)
+            showOpenURLButton()
+        } else {
+            AppsAPIClient.shared.search(searchText) { apps, error in
+                if let error = error {
```
yuliaveres: results are weird here , not what was searched for - please check the searchBar text and the the response's queried text .

yuliaveres: the same counts for button title

Toshi/Controllers/Browse/Controllers/BrowseController.swift
roberthein: I fixed it, can you re-check?

roberthein:
@faymontage I fixed it, can you re-check?

faymontage:
It seems like a network-dependent race condition - perhaps search results coming in even though the query has become a URL in the meantime? It's very difficult to reproduce for me, have to type very fast.

Toshi/Controllers/Browse/Controllers/BrowseController.swift
```diff
@@ -170,31 +195,79 @@ class BrowseController: SearchableCollectionController {
...
+            hideOpenURLButtonIfNeeded()
+
+            AppsAPIClient.shared.search(searchText) { apps, error in
                 if let error = error {
                     let alertController = UIAlertController.errorAlert(error as NSError)
                     Navigator.presentModally(alertController)
                 }
 
-                self.searchResultView.searchResults = items
+                self.searchResultView.searchResults = apps
```
yuliaveres: please, before updating result - DO check if the text in search bar is the same to which you did this request. what if it was delayed a bit ?

Toshi/Controllers/Browse/Views/LaunchAppButton.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+import UIKit
+import TinyConstraints
+
+class LaunchAppButton: UIControl {
```
yuliaveres: what is LaunchAppButton?

roberthein: I reinstated old code, this was the name before, should I change it?

Toshi/Controllers/Browse/Controllers/BrowseController.swift
roberthein: how would you suggest I do this?

yuliaveres: searchText == self.searchbar.text ?

Toshi/Controllers/Browse/Views/LaunchAppButton.swift
yuliaveres: yeaaa, it makes no sense

Toshi/Controllers/Browse/Controllers/BrowseController.swift
yuliaveres: too bad the completion actually doesn't return it, u can add it

Toshi/Controllers/Browse/Views/LaunchAppButton.swift
roberthein: can suggest a name also?

yuliaveres: the variable for this button has perfect name !
what does it do ? OpensUrl, pushes webView

PR #408, opened 2017-08-02
DB files persistency

yuliaveres:
QA purpose logs code still present

### DIFF
--- a/Pods/SignalServiceKit/src/Storage/TSStorageManager.m
+++ b/Pods/SignalServiceKit/src/Storage/TSStorageManager.m
@@ -200,9 +200,21 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
     };
 }
 
-- (void)setupForAccountName:(NSString *)accountName
+- (void)setupForAccountName:(NSString *)accountName isFirstLaunch:(BOOL)isFirstLaunch
 {
     self.accountName = [accountName copy];
+
+    if (isFirstLaunch) {
+
+        NSError *keyFetchError;
+        NSString *previousVersionDBPassword = [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount error:&keyFetchError];
+        if (previousVersionDBPassword) {
+
+            NSError *error;
+            [SAMKeychain setPassword:previousVersionDBPassword forService:self.accountName account:keychainDBPassAccount error:&error];
+        }
+    }
+
     [self setupDatabase];
 }
 
--- a/Pods/SignalServiceKit/src/Storage/TSStorageManager.m
+++ b/Pods/SignalServiceKit/src/Storage/TSStorageManager.m
@@ -546,21 +546,12 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 
 - (void)deletePasswordFromKeychain
 {
-    // [SAMKeychain deletePasswordForService:keychainService account:keychainDBPassAccount];
+    [SAMKeychain deletePasswordForService:self.accountName account:keychainDBPassAccount];
 }
 
 - (NSString *)backupDatabasePath
 {
     return [[self backupDirectoryPath] stringByAppendingFormat:@"/Signal-%@.sqlite", self.accountName];
-}
-
-- (void)deleteDatabaseFile
-{
-    NSError *error;
-    BOOL success = [[NSFileManager defaultManager] removeItemAtPath:[self dbPath] error:&error];
-    if (error) {
-        DDLogError(@"Failed to delete database: %@", error.description);
-    }
 }
 
 - (void)backupDataBaseFile
--- a/Pods/SignalServiceKit/src/Storage/TSStorageManager.m
+++ b/Pods/SignalServiceKit/src/Storage/TSStorageManager.m
@@ -24,7 +24,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 NSString *const TSStorageManagerExceptionNameDatabasePasswordInaccessible = @"TSStorageManagerExceptionNameDatabasePasswordInaccessible";
 NSString *const TSStorageManagerExceptionNameDatabasePasswordInaccessibleWhileBackgrounded =
-    @"TSStorageManagerExceptionNameDatabasePasswordInaccessibleWhileBackgrounded";
+@"TSStorageManagerExceptionNameDatabasePasswordInaccessibleWhileBackgrounded";
 NSString *const TSStorageManagerExceptionNameDatabasePasswordUnwritable = @"TSStorageManagerExceptionNameDatabasePasswordUnwritable";
 NSString *const TSStorageManagerExceptionNameNoDatabase = @"TSStorageManagerExceptionNameNoDatabase";
 
@@ -35,6 +35,8 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
 @interface TSStorageManager ()
 
 @property (nullable, atomic) YapDatabase *database;
+
+@property (nonatomic, copy) NSString *accountName;
 
 @end
 
@@ -99,7 +101,7 @@ static NSString *keychainDBPassAccount    = @"TSDatabasePass";
     static TSStorageManager *sharedManager = nil;
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        sharedManager = [[self alloc] initDefault];
+        sharedManager = [[self alloc] init];
 #if TARGET_OS_IPHONE
         [sharedManager protectSignalFiles];
 #endif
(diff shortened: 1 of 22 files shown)


yuliaveres:
@faymontage ready

Pods/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -489,37 +536,53 @@ - (void)setDate:(NSDate *)value forKey:(NSString *)key inCollection:(NSString *)
...
+        [transaction removeAllObjectsInCollection:[TSInteraction collection]];
+        [transaction removeAllObjectsInCollection:[TSAttachment collection]];
     }];
     [TSAttachmentStream deleteAttachments];
 }
 
 - (void)deletePasswordFromKeychain
 {
-    [SAMKeychain deletePasswordForService:keychainService account:keychainDBPassAccount];
+    // [SAMKeychain deletePasswordForService:keychainService account:keychainDBPassAccount];
```
heumn: I think you forgot this? :)

yuliaveres: true, also ignore all logs for now :) those are just for QA and debug purpose

Pods/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -489,37 +536,53 @@ - (void)setDate:(NSDate *)value forKey:(NSString *)key inCollection:(NSString *)
...
 - (void)resetSignalStorage
 {
+    [self backupDataBaseFile];
+    
     self.database = nil;
     _dbConnection = nil;
-
-    [self deletePasswordFromKeychain];
-
-    [self deleteDatabaseFile];
```
heumn: Looks like this method is never called after it was removed here? The method still persists and leaves 
```removeItemAtPath:[self dbPath]``` never to be called. I need to dig deeper and understand the architecture better to claim if this is safe or not :)

Comment (yuliaveres) [inline Pods/SignalServiceKit/src/Storage/TSStorageManager.m]: i haven't clean not used code yet :/
it is safe as the idea is to store db file (along with password in keychain), not remove it.

File: Toshi/APIClients/ChatAPIClient.swift [hunk 3]
```diff
@@ -67,23 +67,21 @@ public class ChatAPIClient: NSObject {
...
-                        print(json ?? "")
-                        print(response)
-                        print(error)
+            self.teapot.put(path, parameters: requestParameter, headerFields: fields) { result in
+                switch result {
+                case .success(_, let response):
+                    guard response.statusCode == 204 else {
+                        print("Could not register user. Status code \(response.statusCode)")
+
+                        return
```
heumn: I know this is old code and not part of this PR but this looks ... like something that could cause problems at some random point in the future when this app runs hundreds of thousands of user? We are trying to registister a user and if we fail we... return? 🤔  I know the request is supposedly idempotent thus I guess we should just re-try if the attempt failed?

If this is by design I would probably leave a warning for the next guy reading this 😃  This was kind of a head scratcher for me

Toshi/APIClients/ChatAPIClient.swift
```diff
@@ -67,23 +67,21 @@ public class ChatAPIClient: NSObject {
...
                     }
+
+                    TSStorageManager.storeServerToken(parameters.password, signalingKey: parameters.signalingKey)
+                    print("Successfully registered chat user with address: \(cereal.address)")
+                case .failure(let json, let response, let error):
+                    print(json ?? "")
+                    print(response)
+                    print(error)
```
heumn: This seems like something that should be dealt with? 🤔  The entrypoint from this is typical from the first launch of the app with the app doing:

`createNewUser` with no completion block or error checking, followed by a `didCreateUser` that sets 

````
[[NSUserDefaults standardUserDefaults] setBool:NO forKey:RequiresSignIn];
```

that looks... a bit dangerous? The app now appears to be in a state where it should not require a login, but clearly no user was created? 

I have problably misunderstood the flow here but even if I am confused 😅  maybe one could be more clear about why this is done like this :blush:

Pods/SignalServiceKit/src/Storage/TSStorageManager.m
heumn: Will do :)

Toshi/APIClients/idAPIClient.swift
```diff
@@ -148,11 +144,9 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
                             guard response.statusCode == 200 else { return }
                             guard let json = json?.dictionary else { return }
 
-                            TokenUser.createOrUpdateCurrentUser(with: json)
+                            TokenUser.createCurrentUser(with: json)
```
heumn: 100% agree with your decision to split this up 👌

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -35,7 +35,7 @@ open class FavoritesController: SweetTableController {
 
     fileprivate lazy var uiDatabaseConnection: YapDatabaseConnection = {
         let database = Yap.sharedInstance.database
-        let dbConnection = database.newConnection()
+        let dbConnection = database!.newConnection()
```
heumn: Sorry if this is obvious. I often miss things. But force unwrapping this? This shouldnt be optional? 🤔

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -89,15 +89,27 @@ open class FavoritesController: SweetTableController {
     public init() {
         super.init(style: .plain)
 
+        if TokenUser.current != nil {
+            setupForCurrentUserNotifications()
+        }
+
+        NotificationCenter.default.addObserver(self, selector: #selector(userCreated(_:)), name: .userCreated, object: nil)
```
heumn: I know you probably agree with me and this is probably obvious but I have been burned so many times by notifications :) It is quick and easy but such a slippery slope. But this particular notification (user created and lots of things in the app change) is diffictult to get right without it. So probably good for now? I do have some ideas on how you could solve this without notifications :) Lets discuss soon if you are interested

Toshi/Controllers/Favorites/FavoritesController.swift
heumn: I make bugs _all_ the time. Like.. I probably produce more bugs than I produce features. As a result of that I have  discovered that future me who cleans up after messy current me, prefers to have a stacktrace if something like this ☝️ happens (database being nil) ☺️

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -200,14 +200,18 @@ open class SettingsController: UITableViewController {
             alert.addAction(UIAlertAction(title: "Cancel", style: .cancel))
 
             alert.addAction(UIAlertAction(title: "Sign out", style: .destructive) { _ in
-                NotificationCenter.default.post(name: .UserDidSignOut, object: nil)
+
+                guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
```
heumn: 🤔  in what state is the app in where an appDelegate is missing?  That would put the user in a state where he would be .. not allowed to log out? Typical Nikolai produced bug is one where I have simply put a `guard` with a proceeding `return` where I should have crashed. Hard. Because else my boss comes with his phone and points me at the weirdes f**ing bugs I have seen and I have 0 chance of debugging my mistakes 😅

Toshi/Models/Cereal.swift
```diff
@@ -119,4 +131,12 @@ public class Cereal: NSObject {
     public func sha3WithWallet(string: String) -> String {
         return walletCereal.sha3(string: string)
     }
+    
+    @objc fileprivate func userCreated(_ notification: Notification) {
+        Yap.sharedInstance.insert(object: mnemonic.words.joined(separator: " "), for: Cereal.privateKeyStorageKey)
+
+        if let words = Yap.sharedInstance.retrieveObject(for: Cereal.privateKeyStorageKey) as? String {
+            print(words.components(separatedBy: " "))
```
heumn: I understand this is probably just for debugging. Just triplecheck that you _do not_ ship this 😉

Toshi/Models/Yap.swift
```diff
@@ -20,45 +20,188 @@ protocol Singleton: class {
...
+
+            database = YapDatabase(path: UserDB.dbFilePath, options: options)
+
+            print("\n\n------ ------ ------ \n *** DATABASE CREATED *** \n------ ------ ------ \n\n")
+
+            let url = NSURL(fileURLWithPath: UserDB.dbFilePath)
+            try! url.setResourceValue(false, forKey: .isUbiquitousItemKey)
+            try! url.setResourceValue(true, forKey: .isExcludedFromBackupKey)
+
+            mainConnection = database?.newConnection()
```
heumn: I dug a bit deeper into what is going on with `isCurrentUserDataAccessible` and maybe we should re-consider this approach as it is using `FileManager.default.fileExists(atPath: ` as the state of truth? 🤔 

From apple doc:

```
Attempting to predicate behavior based on the current state of the file system or a particular file on the file system is not recommended. Doing so can cause odd behavior or race conditions. It’s far better to attempt an operation (such as loading a file or creating a directory), check for errors, and handle those errors gracefully than it is to try to figure out ahead of time whether the operation will succeed. For more information on file-system race conditions, see Race Conditions and Secure File Operations in Secure Coding Guide.
```

Toshi/Models/Yap.swift
```diff
@@ -20,45 +20,188 @@ protocol Singleton: class {
...
 
-        let url = NSURL(fileURLWithPath: path)
+        let url = NSURL(fileURLWithPath: UserDB.dbFilePath)
         try! url.setResourceValue(false, forKey: .isUbiquitousItemKey)
         try! url.setResourceValue(true, forKey: .isExcludedFromBackupKey)
 
-        mainConnection = database.newConnection()
+        mainConnection = database?.newConnection()
+
+        self.insert(object: TokenUser.current?.JSONData, for: TokenUser.storedUserKey)
```
heumn: Could there ever be a scenario (o race condition) where `TokenUser.current` is nil and you are effectively deleing the stored user key as a result? 🤔

Toshi/Models/Yap.swift
```diff
@@ -20,45 +20,188 @@ protocol Singleton: class {
...
+                print("\n|| --- Something went wrong while moving to BACKUP PATH\n||\n||----------------\n||\n")
+            }
+        }
+    }
+
+    fileprivate func deleteFileIfNeeded(at path: String) {
+        if FileManager.default.fileExists(atPath: path) {
+            do {
+                try FileManager.default.removeItem(atPath: path)
+            } catch {}
```
heumn: guess it is not a point in wrapping this in in a do try catch if there are not error handling? 🤔

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -55,32 +59,32 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    [self configureAndPresentWindow];
+
     return YES;
 }
 
+- (void)handleFirstLaunchIfNeeded
+{
+    if (![[NSUserDefaults standardUserDefaults] boolForKey:LaunchedBefore]) {
+        [[NSUserDefaults standardUserDefaults] setBool:YES forKey:LaunchedBefore];
+        [[NSUserDefaults standardUserDefaults] synchronize];
```
heumn: We do not have to write `synchronize` any more :) 

https://twitter.com/catfish_man/status/510125894791208960

and since userDefaults is so fast (unless you store your whole DB in it like drunk me would do), just directly storing YES here without any checks is probably more than enough. We are talking about less than 1ms :)

heumn:
This looks great @yuliaveres ! Just some minor things that could maybe be improved? Great job 💪

heumn:
sporzaredactie-xT0Gqn53IOqZDr07V6

Toshi/APIClients/ChatAPIClient.swift
yuliaveres: oh ye, i absolutely agree with you here. the defaults should be manipulated with only if actually registration happened. definitely. thx!

Toshi/Controllers/Favorites/FavoritesController.swift
yuliaveres: i prefer to reflect realistic states of an instance, "must be there" or "might not be there";
db file is present only if:
 - user is logged in;
 - db file load didn't fail.

due to this I prefer to reflect this and use it as optional, cause it is actually optional. 😉 

but good point in general, agree with what u think about keep it as needed 😄

Toshi/Controllers/Favorites/FavoritesController.swift
yuliaveres: great idea, thank you! lets discuss 👍

Toshi/Models/Yap.swift
yuliaveres: Great feedback on this! thx

Toshi/Models/Yap.swift
yuliaveres: nope

Toshi/Models/Yap.swift
yuliaveres: can't trying to open yap db file for check purpose, cause it creates new one if nothing exists yet :/
not sure about any alternatives 🤔

yuliaveres: also looking inside yap code - it heavily relies on `fileExistsAt...`
but your point is important, will try to figure out some way

Toshi/Controllers/Settings/SettingsController.swift
yuliaveres: yes, i will change it here, calling function on optional would not add any mess.
(the purpose of this is just for AppDelegate casting 🤷‍♂️, obviously can use `as` ) 😄

yuliaveres: also would be good to extract all of this out of AppDelegate actually 🤦‍♂️

PR #409, opened 2017-08-03
Initial greeting chat

yuliaveres:
greet


Toshi/Supporting Files/AppDelegate.m
```diff
@@ -430,13 +436,19 @@ - (void)registerForRemoteNotifications {
...
 
         [[EthereumAPIClient shared] registerForMainNetworkPushNotifications];
 
         [[EthereumAPIClient shared] registerForSwitchedNetworkPushNotificationsIfNeededWithCompletion:nil];
 
+        typeof(self)strongSelf = weakSelf;
+        if (strongSelf.registeredNewUser) {
+            [ChatsInteractor triggerBotGreeting];
+        }
+
```
faymontage: @yuliaveres can we do this regardless of PN success or failure? For instance, PN registration will always fail when testing a Distribution build that has been installed from anywhere other than TestFlight/App Store - but the greeting message should still go out.

yuliaveres: sure

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -168,13 +168,21 @@ - (void)signOutUser
...
 {
     __weak typeof(self)weakSelf = self;
-    [[IDAPIClient shared] registerUserIfNeeded:^{
-        typeof(self)strongSelf = weakSelf;
+    [[IDAPIClient shared] registerUserIfNeeded:^(UserRegisterStatus status){
+
+        if (status != UserRegisterStatusFailed) {
 
-        [[ChatAPIClient shared] registerUser];
+            typeof(self)strongSelf = weakSelf;
```
heumn: 👌

Toshi/Resources/Development.plist
```diff
@@ -62,6 +62,8 @@
 	<string>Required to upload pictures.</string>
 	<key>TokenChatServiceBaseURL</key>
 	<string>https://token-chat-service-development.herokuapp.com</string>
+	<key>InitialGreetingAddress</key>
+	<string>0xdf0096052836251b0fafd4f0f12771deea5c8748</string>
```
faymontage: 😍

Toshi/APIClients/ChatAPIClient.swift
```diff
@@ -78,10 +78,12 @@ public class ChatAPIClient: NSObject {
 
                     TSStorageManager.storeServerToken(parameters.password, signalingKey: parameters.signalingKey)
                     print("Successfully registered chat user with address: \(cereal.address)")
+                    completion?(true)
                 case .failure(let json, let response, let error):
                     print(json ?? "")
                     print(response)
                     print(error)
+                    completion?(false)
```
heumn: I am little late to the party here, but what about instead of using optionals, just default to an empty block? 🤔  less `??!??"!!??` the better is my general rule of thumb 😆  cc @yuliaveres

Elland: Good idea!

yuliaveres: @heumn  will make in separate pr 🤷‍♂️ 🤷‍♀️ , but i have always used completions as optionals.  🤷‍♂️ 🤷‍♀️

heumn: I have no objectively good reasons for why default values are better so this is nitty

Apple seem to push us into using more defaults thought:

```
Take advantage of defaulted parameters when it simplifies common uses. Any parameter with a single commonly-used value is a candidate for a default.
```

also... the syntax of completion?() looks.... 
image

yuliaveres: 😹

PR #411, opened 2017-08-09
Show keyboard after one second only when view is visible

marijnschilling:
Fixes this issue: https://trello.com/c/FDKhtd2y/213-improve-keyboard-handling-in-chatscontroller-chat-transition

Showing keyboard after one second make it look very smooth. I make sure to check wether the view is still visible so it wont try to show the keybord after the view is being dismissed:
keyboard


yuliaveres:
Fixed in different PR along with performance improvements

PR #413, opened 2017-08-09
Removed retained self from closures

heumn:
Some minor stuff I spotted

`self` in all of these cases owns the `DatabaseConnection`. By referring to `self` in the async read block from this connection you could end up with a memory leak if the block is not nullified.

Generally this is an unlikely case, but I would still argue that it is safter to program with the consensus that "the object self owns, should not decide how long self lives" because even if this does not leak, you are forcing the software to do a lot of unneeded work in cases it is just waiting for the block to exit close.

Normally I tend to do `[unowned let]` in these types of situations because:
- If there ever was a memory leak in the owned object and the block would be called after self was deallocated I would have a crash. I always prefer crashes to weird bugs I have no stack trace to go after.
- By doing `[unowned let]` I don't have to deal with optionals 🎉 

Why did I do `strongSelf` in some cases and sometimes not? Just a personal preference. Up for debate. I prefer to not having the extra guard in the online closures. It will instantly exit scope regardless. For the longer ones, I add it go exit scope faster.

[WIP] because I have not tested if I broke anything


Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -232,15 +233,16 @@ final class ChatViewModel {
...
 
-            for i in 0 ..< self.mappings.numberOfItems(inSection: 0) {
+            for i in 0 ..< strongSelf.mappings.numberOfItems(inSection: 0) {
                 let indexPath = IndexPath(row: Int(i), section: 0)
                 guard let dbExtension = transaction.ext(TSMessageDatabaseViewExtensionName) as? YapDatabaseViewTransaction else { return }
-                guard let interaction = dbExtension.object(at: indexPath, with: self.mappings) as? TSInteraction else { return }
+                guard let interaction = dbExtension.object(at: indexPath, with: strongSelf.mappings) as? TSInteraction else { return }
 
                 var shouldProcess = false
                 if let message = interaction as? TSMessage, SofaType(sofa: message.body ?? "") == .paymentRequest {
```
yuliaveres: line 252 - ` messages.append(strongSelf.interactor...`

PR #414, opened 2017-08-10
Experiment/mocking

marijnschilling:
I added a `MockedTeapot` in Teapot that you need to initialise with a bundle, if you have a json file in your test bundle with the same name of your endpoint, it will use that mocked json.


marijnschilling:
Too many merge conflicts in podfile :(

PR #415, opened 2017-08-10
Mocking

marijnschilling:
I added a MockedTeapot in Teapot that you need to initialise with a bundle, if you have a json file in your test bundle with the same name of your endpoint, it will use that mocked json.


Tests/AppsAPIClientTests.swift
```diff
@@ -0,0 +1,38 @@
...
+import UIKit
+import Teapot
+
+class AppsAPIClientTests: XCTestCase {
+    var subject: AppsAPIClient!
+
+    override func setUp() {
+        super.setUp()
+
+        let mockTeapot = MockTeapot(baseURL: URL(string: "https://token-id-service-development.herokuapp.com")!, bundle: Bundle(for: AppsAPIClientTests.self))
```
yuliaveres: we have constant for this `TokenIdServiceBaseURLPath`

Toshi.xcodeproj/project.pbxproj
```diff
@@ -1194,6 +1194,7 @@
 		C1128E2CDD482DB9BE1BF5A3 /* libPods-CocoaPods-Distribution.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 7E67E3F2BFB7B0171BF61B30 /* libPods-CocoaPods-Distribution.a */; };
 		D197B11DD70DB5696FFC4FD2 /* UINotificationName+Additions.swift in Sources */ = {isa = PBXBuildFile; fileRef = D197B669EF87D9D6642E12BA /* UINotificationName+Additions.swift */; };
 		D197B401E977724E946A55BE /* ActiveNetworkDisplaying.swift in Sources */ = {isa = PBXBuildFile; fileRef = D197BAEA583CAF9862D17B38 /* ActiveNetworkDisplaying.swift */; };
+		D197B567CD56C97880DCDF96 /* apps?top=false&recent=true&limit=10.json in Resources */ = {isa = PBXBuildFile; fileRef = D197BA8F82B98173AA0D80D5 /* apps?top=false&recent=true&limit=10.json */; };
```
yuliaveres: i dnt like the file name

Toshi/APIClients/AppsAPIClient.swift
```diff
@@ -29,6 +29,12 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
         teapot = Teapot(baseURL: URL(string: TokenDirectoryServiceBaseURLPath)!)
     }
 
+    convenience init(mockTeapot: MockTeapot) {
+        self.init()
+        self.teapot = mockTeapot
+    }
+
```
yuliaveres: please remove the line

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -50,6 +50,13 @@ @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+    BOOL isAppRunFromTests = (NSClassFromString(@"XCTestCase") != nil);
```
yuliaveres: why do we need this ?

danger/shared/Dangerfile
```diff
@@ -92,9 +92,4 @@ files_to_check = (git.modified_files + git.added_files).uniq
 			end
 		end 
 	end
-
-	## Check wether our file is larger than 200 lines and doesn't include any Marks
```
yuliaveres: is this related to the PR ?

Toshi.xcodeproj/project.pbxproj
yuliaveres: maybe better is to name it in a more descriptive way ?

marijnschilling: hmmm i use the mechanism now that when you use this `MockTeapot` you can do any call to any endpoint and then it will look for a json file with the name of this endPoint. This way I can easily intercept every api call possible by just overriding the execute method. This does mean though that the json file should be named exactly as the endpoint. 

If you know a better way to do this i would be open for it, but this is the most elegant way i could think of!

It would be a lot better when i put this file in a folder, called `AppsAPIClientMocks` I'll do that for sure.

Toshi/Supporting Files/AppDelegate.m
marijnschilling: Otherwise it will start up the app when running the tests, this is time consuming. This way tests are faster (which is very important for tests!)

danger/shared/Dangerfile
marijnschilling: No its not, i will add this in a separate PR and pull it into this one I guess!

Toshi/Supporting Files/AppDelegate.m
yuliaveres: it would be better to extract it from `didFinishLaunchingWithOptions ` 
would u mind to take a look at this?
http://www.mokacoding.com/blog/prevent-unit-tests-from-loading-app-delegate-in-swift/

Carthage/Build/iOS/EtherealCereal.framework/Headers/EtherealCereal.swift
```diff
@@ -22,7 +22,7 @@ public class EtherealCereal: NSObject {
 
     fileprivate var _privateKeyData: Data?
 
-    public var privateKeyData: Data {
```
marijnschilling: It's just the carthage file update 🤷‍♂️

Toshi/Supporting Files/AppDelegate.m
marijnschilling: I looked into it and it seems that in Obj-C it's not that easy. I would have to create an empty TestAppDelegate. But this approach might still be better than adding sometihng in the appDelegate itself, what do you think @yuliaveres?

https://stackoverflow.com/questions/15714697/unit-testing-in-xcode-does-it-run-the-app

yuliaveres: right, separate `AppDelegate` would be much better in my point of view 👍

Toshi/Supporting Files/main.m
```diff
@@ -17,8 +17,13 @@
...
 
-int main(int argc, char * argv[]) {
+int main(int argc, char* argv[]) {
     @autoreleasepool {
-        return UIApplicationMain(argc, argv, nil, NSStringFromClass([AppDelegate class]));
+        BOOL runningTests = NSClassFromString(@"XCTestCase") != nil;
+        if(!runningTests) {
+            return UIApplicationMain(argc, argv, nil, NSStringFromClass([AppDelegate class]));
+        } else {
+            return UIApplicationMain(argc, argv, nil, @"TestAppDelegate");
```
yuliaveres: why u can't use the same `NSStringFromClass` as with `AppDelegate` ?

yuliaveres: also maybe better to put this return outside else and thus - get rid of else block ? :)

marijnschilling:
😅 🎉

PR #417, opened 2017-08-11
Fix payment message rendering

yuliaveres:
@heumn , @roberthein 
that is the only workaround I have figured out with the wrong table view content size calculation before scrolling to the bottom.
the ugliest here is initial scroll. i need to use snapshot in order to hide tableView layout blinks, without layout content size is calculated not correctly.
also - i need to use some delays :( I am open to any suggestions, and also, feel free to try to fix this in another way. But this is blocker for day-before-yesterday-Build 😢  ! so we need to have it asap. 

in order to try it - just open different chats, and try to message bots, f.e. **spambot**


Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -386,8 +390,41 @@ final class ChatController: OverlayController {
...
+
+            self.view.addSubview(snapshot)
+            snapshot.fillSuperview()
+            self.tableView.alpha = 0.0
+
+            let tableViewHeight = self.tableView.bounds.height
+
+            self.tableView.layoutIfNeeded()
+
+            self.tableView.scrollRectToVisible(CGRect(x: 0, y: self.tableView.contentSize.height - tableViewHeight, width: self.tableView.bounds.width, height: self.tableView.bounds.height), animated: false)
```
yuliaveres: without scrolling first time - content size is not correct as well 👎  must use it here

PR #418, opened 2017-08-11
Alternative solution to "scroll top bottom"

heumn:
Threw this together this proposal as an alternative route to all the content inset and scroll position code from this PR

https://github.com/toshiapp/toshi-ios-client/pull/417

Video:
https://www.dropbox.com/s/8wsublayibtinjb/alternative.mov?dl=0

So let's discuss. Scrolling to the top will in my experience *always* work and this saves you a lot of hassle. Appending new items are also done at the logical "end" of an array as opposed to "top" when dealing with a normal `UITableView`

It should be pretty clear what I do, but regardless: I just flip the `UITableView` 180 degrees and do the same thing on the `UITableViewCell` to counter the rotation. Lastly I simply reveres the messages. (Consider this a hack and only a demo to show you my suggestion. I am not too famililar with the data layer and would need more than 15 minutes to actually code something shippable).

I quickly skimmed over the rest of the view controller and I think you can delete a bunch of code by going in this direction.


Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -387,7 +391,7 @@ final class ChatController: OverlayController {
     }
 
     fileprivate func scrollToBottom(animated: Bool = true) {
-        tableView.scrollRectToVisible(CGRect(x: 0, y: tableView.contentSize.height - tableView.bounds.height, width: tableView.bounds.width, height: tableView.bounds.height), animated: false)
+        self.tableView.scrollToRow(at: IndexPath(item: 0, section: 0), at: .top, animated: true)
```
heumn: this is of course not safe. just a hack

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -123,7 +123,7 @@ final class ChatViewModel {
 
     var visibleMessages: [Message] = [] {
         didSet {
-            messageModels = visibleMessages.flatMap { MessageModel(message: $0) }
+            messageModels = visibleMessages.flatMap { MessageModel(message: $0) }.reversed()
```
heumn: This I am sure should not be done here

Toshi/Controllers/Messaging/ChatController.swift
yuliaveres: why is this not safe ?

Elland: If tableview is empty, won't this crash?If tableview is empty, won't this crash?

yuliaveres: will add this fix

heumn: > If tableview is empty, won't this crash?
bingo :)

Elland:
@heumn kinda funny that this is how I did it before, much more convenient and efficient.

heumn:
hehe yeah, classic UIKit fix 😅

tristan:
QA check with this on the iphone 5. it's completely usable again with this. current app store build is totally unusable. I can't find any serious issues with this that i think should delay it going out asap.

heumn:
glad I could help!

life-gets-down-olAik8MhYOB9K

heumn:
happy-bill-hader-AB29dGYfE9xO8

glad I could help!

roberthein:
This is a really hacky "solution". I'm fine with merging it for now if we need to address these issues now, but we should keep looking for a real solution too.

3 regressions I see:
- the `MessagesCornerView` is not rendered correctly with grouped messages
- the scrollIndicatorInsets are off by 10pt at the bottom and the top
- tapping the statusBar scrolls to the bottom instead of the top

yuliaveres:
@roberthein  thx for checking this!
I fixed all reported issues, am going to merge this now

PR #420, opened 2017-08-14
Add step to swift lint guide on how to edit the swift lint file

marijnschilling:
Since the file is hidden you will need to look up how to edit it, so it might be good to have that step here.


marijnschilling:
Ha yes, was a bit too short I realise now!

PR #421, opened 2017-08-14
Improve URL validation.

roberthein:
Improve URL validation.


Toshi/Extensions/String+isURL.swift
```diff
@@ -16,15 +16,19 @@
...
+        if let url = URL(string: self), UIApplication.shared.canOpenURL(url) {
+            let urlPattern = "^(http|https|ftp)\\://([a-zA-Z0-9\\.\\-]+(\\:[a-zA-Z0-9\\.&amp;%\\$\\-]+)*@)*((25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9])\\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[0-9])|localhost|([a-zA-Z0-9\\-]+\\.)*[a-zA-Z0-9\\-]+\\.(com|edu|gov|int|mil|net|org|biz|arpa|info|name|pro|aero|coop|museum|[a-zA-Z]{2}))(\\:[0-9]+)*(/($|[a-zA-Z0-9\\.\\,\\?\\'\\\\\\+&amp;%\\$#\\=~_\\-]+))*$"
+            return matches(pattern: urlPattern)
         }
 
         return false
     }
+    
+    private func matches(pattern: String) -> Bool {
+        let regex = try! NSRegularExpression( pattern: pattern, options: [.caseInsensitive])
```
yuliaveres: the initializer might throw an error in case of a failure, if the function is called with a weird String right?
so better to handle error with `do{} catch{}` and do not force unwrap `NSRegularExpression` assuming it will be always created.

also, please remove space before `pattern` label 😉

yuliaveres:
also, can you make sure with @faymontage is we still interested in "prefixing an implied `https://` in common cases" ?

Toshi/Extensions/String+isURL.swift
roberthein: You are absolutely right!
Fixed it, please re-check ✌️

faymontage:
For implied protocol at input-time we either need:

1. Whitelist of common domains and their protocol
or 
2. Background HTTP request to a URL to validate that it responds appropriately before showing it as a result.

Better to leave it it out for now I reckon.

PR #422, opened 2017-08-14
Implement QR intents, helper models, and tests

faymontage:
How we handle QR scan results and the data they contain is mostly broken: 

⁃	Wei integers are wrongly interpreted as Ether
⁃	Most Ethereum address standards are unsupported
⁃	No model for validating/normalizing addresses
⁃	Query string metadata is dropped in some cases
⁃	String patterns and parsing logic are mixed into control structures (ie. https://github.com/toshiapp/toshi-ios-client/blob/master/Toshi/Controllers/TabBarController.swift#L196)

This PR allows for fixing all of the above, and allows things like:  

```swift
if let intent = QRCodeIntent(result: "iban:XE420ENF06QHAD2B0729XZJ1OU26UVM0TSN?amount=0.025&memo=thanks") {
    switch intent {
        case .webSignIn(let loginToken):
            //
        case .paymentRequest(let weiValue, let address, let username, let memo):
            //
        case .addressInput(let address):
            //
        case .addContact(let username):
            //
    }
}
```

It can also provide address validation for features like Send Payment:

```swift
if let address = EthereumAddress(raw: "ethereum:0x037be053f866be6ee6dda11f258bd871b701a8d7")?.normalized {
	//address == “0x037be053f866be6ee6dda11f258bd871b701a8d7”
} else {
	//not a valid ethereum address input!
}
```


Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+
+var base16Alphabet = "0123456789abcdef"
+var base36Alphabet = "0123456789abcdefghijklmnopqrstuvwxyz"
```
heumn: I think I would use `fileprivate let` or `static let` for these

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,79 @@
...
+var base36AlphabetMap: [Character: Int] {
+    var reverseLookup = [Character: Int]()
+    for z in 0..<base36Alphabet.characters.count {
```
heumn: I guess you could just use `for z in base36Alphabet.characters` ? 🤔

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,79 @@
...
+        let x = base36Alphabet[base36Alphabet.index(base36Alphabet.startIndex, offsetBy: z)]
+        reverseLookup[x] = z
+    }
+
+    return reverseLookup
```
heumn: Asking because it is not entirely clear to me why you do this. What is the idea with the `reverseLoopup`? What is it used for?

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func hexFrom(base36: String) -> String? {
```
heumn: I think also it might be more appropriate to either:
- extend String
- or make this a fileprivate extension on String

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,79 @@
...
+    var bytes = [Int]()
+    for i in 0..<base36.characters.count {
```
heumn: the convention in swift is always to fully type your variable names as it makes it a bit easier to read what is going on :)

Toshi/Models/EthereumAddress.swift
faymontage: It is used to ask, for instance, "at what index in base36Alphabet is the character `g` located?"

It could just as well be precomputed and included as a constant in the sourcecode, but I've left it for now as all the base36 code may be extracted into an arbitrary base conversion util.

Toshi/Models/PaymentRequestMetadata.swift
```diff
@@ -0,0 +1,36 @@
...
+                case "amount":
+                    let ether = NSDecimalNumber(string: value)
+                    tempWei = ether.multiplying(byPowerOf10: EthereumConverter.weisToEtherPowerOf10Constant).toHexString
+                // "value" is used by Toshi and others as an integer denominated in Wei
+                case "value":
+                    tempWei = NSDecimalNumber(string: value).toHexString
+                case "memo":
+                    tempMemo = value
+                default:
+                    continue
```
heumn: Are you sure `continue` is what you are after here? 🤔

Toshi/Models/QRCodeIntent.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+func firstMatch(_ input: String, pattern: String) -> NSTextCheckingResult? {
```
heumn: I think I would restrict knowledge / or put this someplace else...

Wouldnt this be appropriate as a function on the enum `QRCodeIntent` 🤔

Toshi/Models/QRCodeIntent.swift
```diff
@@ -0,0 +1,37 @@
...
+    case addressInput(address: String)
+
+    init?(result: String) {
+        if let address = EthereumAddress(raw: result) {
+            if let metadata = PaymentRequestMetadata(with: result) {
+                self = .paymentRequest(weiValue: metadata.weiValue, address: address.normalized, username: nil, memo: metadata.memo)
+            } else {
+                self = .addressInput(address: address.normalized)
+            }
+        } else if let match = firstMatch(result, pattern: "^web-signin:(\\w+)$") {
```
heumn: I think I would pull out these strings as `fileprivate let` on QRCodeIntent if they are only used within that type

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,81 @@
...
+        case .icap:
+            guard match.numberOfRanges == 3 else { return nil }
+            let account = (input as NSString).substring(with: match.rangeAt(2))
+            guard let address = account.base36to16() else { return nil }
+            normalized = "0x" + address.lowercased()
+        }
+    }
+}
+
+fileprivate let base16Alphabet = "0123456789abcdef"
```
heumn: This one is never used :)

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,81 @@
...
+fileprivate let base36Alphabet = "0123456789abcdefghijklmnopqrstuvwxyz"
+fileprivate var base36AlphabetMap: [Character: Int] {
+    var reverseLookup = [Character: Int]()
+    for characterIndex in 0..<base36Alphabet.characters.count {
+        let character = base36Alphabet[base36Alphabet.index(base36Alphabet.startIndex, offsetBy: characterIndex)]
+        reverseLookup[character] = characterIndex
+    }
+
+    return reverseLookup
+}
```
heumn: Sorry for all the back and forth, but I would put this inside the fileprivate extension on String :) Just looks out of place and onle ever used inside the scope of that extension. I guess no reason to share it with anyone else? 🤔

Toshi/Models/QRCodeIntent.swift
```diff
@@ -0,0 +1,37 @@
...
+            let loginToken = (result as NSString).substring(with: match.rangeAt(1))
+            self = .webSignIn(loginToken: loginToken)
+        } else if let match = firstMatch(result, pattern: "^https?://[^\\.]+.toshi.org/add/@([^/\\?]+)") {
+            let username = (result as NSString).substring(with: match.rangeAt(1))
+            self = .addContact(username: username)
+        } else if let match = firstMatch(result, pattern: "^https?://[^\\.]+.toshi.org/pay/@([^/\\?]+)") {
```
heumn: I think I would even create enums for the patterns here. That would make it easier to parse what you are matching against. This "^https?://[^\\.]+.toshi.org/pay/@([^/\\?]+)" string means nothing to me as an outsider, but I guess it would make more sense if it was something like an enum called `RegexPattern` with cases like `payUser`, `addUser` with corresponding strings you could use?

heumn:
This is great 👌

Toshi/Models/QRCodeIntent.swift
faymontage: Yeah definitely - they actually started as the raw values of the QRCodeIntents, but I had to ditch that design when I added associated values to the enum (you can't have both associated values and raw values - why?)

heumn: You can _sort_ of have both. What about something like this?

```
enum QRCodeIntent {
    case webSignIn(loginToken: String)
    case paymentRequest(weiValue: String, address: String?, username: String?, memo: String?)
    case addContact(username: String)
    case addressInput(address: String)

   var pattern: String {
    // switch over self
    // then for each case, you return the corresponding pattern?
    return pattern
    }
}
```

heumn: Enums are really powerful in swift 😍 

skjermbilde 2017-08-14 kl 20 26 35

faymontage: I'll go with your let binding suggestion - I realized patterns and intents only have a 1-1 correlation by coincidence, and may not in the future, so I don't want to bind them too closely.

Toshi/Models/PaymentRequestMetadata.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+
+public struct PaymentRequestMetadata {
+
+    let weiValue: String
```
yuliaveres: do you need to have those as `let` ?

Toshi/Models/QRCodeIntent.swift
```diff
@@ -0,0 +1,43 @@
...
+            return nil
+        }
+    }
+}
+
+fileprivate extension String {
+    func firstMatch(pattern: String) -> NSTextCheckingResult? {
+        guard let regex = try? NSRegularExpression(pattern: pattern, options: []) else { return nil }
+        let matches = regex.matches(in: self, options: [], range: NSRange(location:0, length: self.characters.count))
+        return matches.first
```
yuliaveres: `return regex.firstMath(in: self, options: [], range: NSRange(location:0, length: self.characters.count))`

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    let normalized: String
+
+    init?(raw: String) {
+        let input = raw.lowercased()
+
+        guard let format = AddressFormat(input) else { return nil }
+        guard let regex = try? NSRegularExpression(pattern: format.rawValue, options: []) else { return nil }
+
+        let matches = regex.matches(in: input, options: [], range: NSRange(location:0, length: input.characters.count))
```
yuliaveres: u can use your extension method even 😎

Toshi/Models/QRCodeIntent.swift
yuliaveres: actually, i would use more readable SweetSwift - `.length ` on a string 😉

yuliaveres:
Looks great in general ! nice

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return reverseLookup
+}
+
+fileprivate extension String {
+    func base36to16() -> String? {
+        var bytes = [Int]()
+        for characterIndex in 0..<self.characters.count {
+            let value = base36AlphabetMap[self[self.index(self.startIndex, offsetBy: characterIndex)]]
```
faymontage: It's already rewritten as a character iteration in latest changes since that's all it's doing - was leftover from a more general implementation.

Toshi/Models/PaymentRequestMetadata.swift
faymontage: As opposed to what? My intention is that they be public constant stored properties on instances of PaymentRequestMetadata.

yuliaveres: as u wish, i would make them readonly vars not to use those temp variables 💭

Toshi/Models/EthereumAddress.swift
faymontage: 😆 of course `firstMatch` is already a thing

Toshi/Models/PaymentRequestMetadata.swift
faymontage: That works fine for memo 👌 - but how would you handle weiValue without a temp var? There's no guarantee that a value will be present in the query params, and the init should fail if weiValue doesn't get set.

yuliaveres: u can check in the end if `self.tempWei == nil { return nil }` the same way as you do with temp var right ?

yuliaveres: but that is not critical of course. then u need to have it as optional - may be not what u needed. 
so what i would do in this case - have empty by default. and in the end check for length - if empty - fail. but up to you, one temp is much better than two 😏

Toshi/Models/EthereumAddress.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    let normalized: String
+
+    init?(raw: String) {
+        let input = raw.lowercased()
+
+        guard let format = AddressFormat(input) else { return nil }
+        guard let regex = try? NSRegularExpression(pattern: format.rawValue, options: []) else { return nil }
+        guard let match = regex.firstMatch(in: input, options: [], range: NSRange(location:0, length: input.length)) else { return nil }
```
yuliaveres: please, use the extension method u created, otherwise why u created it ? 😼

PR #425, opened 2017-08-16
Remove force down-casts

marijnschilling:
This PR enables the `force_cast` rule in SwiftLint. It will give a compile error when you use force downcast from now on.

This PR needs thorough review I think, because it touches a lot of different things in the code base. So would be great if you could check precisely if the logic is right ✌️ 

https://trello.com/c/BKQEd75y/224-425-add-no-force-casts-try-and-unwrapping-rule-in-swift-lint-so-rewrite-these-cases-in-project


Toshi/Controllers/Messaging/ChatInteractor.swift
```diff
@@ -227,7 +227,7 @@ final class ChatsInteractor: NSObject {
 
             return message
         } else {
-            return Message(sofaWrapper: nil, signalMessage: interaction as! TSMessage, date: interaction.dateForSorting(), isOutgoing: false)
+            return Message(sofaWrapper: nil, signalMessage: (interaction as? TSMessage) ?? TSMessage(), date: interaction.dateForSorting(), isOutgoing: false)
```
roberthein: Wouldn't it be better if `signalMessage` was an optional in `Message`? Satisfying the initialiser an empty `TSMessage` feels wrong.

Toshi/Controllers/Messaging/ChatsController.swift
```diff
@@ -134,7 +134,9 @@ open class ChatsController: SweetTableController {
         let notifications = uiDatabaseConnection.beginLongLivedReadTransaction()
 
         // If changes do not affect current view, update and return without updating collection view
-        guard let viewConnection = self.uiDatabaseConnection.ext(TSThreadDatabaseViewExtensionName) as? YapDatabaseViewConnection else { return }
+        // swiftlint:disable force_cast
+        let viewConnection = uiDatabaseConnection.ext(TSThreadDatabaseViewExtensionName) as! YapDatabaseViewConnection
```
roberthein: I see that this code occurs 3 times:
```
let notifications = uiDatabaseConnection.beginLongLivedReadTransaction()

// If changes do not affect current view, update and return without updating collection view
// swiftlint:disable force_cast
let viewConnection = uiDatabaseConnection.ext(TSThreadDatabaseViewExtensionName) as! YapDatabaseViewConnection
// swiftlint:enable force_cast

let hasChangesForCurrentView = viewConnection.hasChanges(for: notifications)
if !hasChangesForCurrentView {
    uiDatabaseConnection.read { transaction in
        self.mappings.update(with: transaction)
    }

    return
}

let yapDatabaseChanges = viewConnection.getChangesFor(notifications: notifications, with: mappings)
```
Can we also do this only once?

Toshi/Extensions/YapDatabaseViewConnection+Additions.swift
```diff
@@ -0,0 +1,30 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+extension YapDatabaseViewConnection {
+    func getChangesFor(notifications: [Notification], with mappings: YapDatabaseViewMappings) -> (rowChanges: [YapDatabaseViewRowChange]?, sectionChanges: [YapDatabaseViewSectionChange]?) {
```
roberthein: Great idea! 😍

Toshi/Models/Message.swift
```diff
@@ -129,13 +129,13 @@ open class Message: NSObject, NOCChatItem {
         guard let sofaWrapper = self.sofaWrapper else { return "" }
         switch sofaWrapper.type {
         case .message:
-            return (sofaWrapper as! SofaMessage).body
+            return (sofaWrapper as? SofaMessage)?.body ?? ""
```
roberthein: I think `text` should be optional, so you can do:
```
var text: String? {
    guard let sofaWrapper = self.sofaWrapper else { return nil }
    switch sofaWrapper.type {
    case .message:
        return (sofaWrapper as? SofaMessage)?.body
    case .paymentRequest:
        return (sofaWrapper as? SofaPaymentRequest)?.body
    case .payment:
        return nil
    case .command:
        return (sofaWrapper as? SofaCommand)?.body
    default:
        return sofaWrapper.content
    }
}
```

Toshi/Models/SofaWrapper.swift
```diff
@@ -142,8 +142,11 @@ open class SofaMessage: SofaWrapper {
         open var subcontrols: [Button] = []
 
         public init(json: [String: Any]) {
-            type = ControlType(rawValue: json["type"] as! String)!
-            _label = json["label"] as! String
+            if let jsonType = json["type"] as? String {
+                type = ControlType(rawValue: jsonType) ?? .button
+            }
+            
+            _label = json["label"] as? String ?? ""
```
roberthein: I think it's better to make `label` and `_label` optional:
```
private var _label: String?

open var label: String?
````
So you can remove ` ?? ""`

File: Toshi/Models/UserBootstrapParameters.swift [hunk 6]
```diff
@@ -93,19 +93,17 @@ public class UserBootstrapParameter {
...
 
         registrationId = TSAccountManager.getOrGenerateRegistrationId()
 
         signalingKey = CryptoTools.generateSecureRandomData(52).base64EncodedString()
 
-        let keyPair = Curve25519.generateKeyPair()!
-        let keyToSign = (keyPair.publicKey() as NSData).prependKeyType()! as Data
-        let signature = Ed25519.sign(keyToSign, with: identityManager.identityKeyPair())! as Data
+        let keyPair = Curve25519.generateKeyPair()
+        let keyToSign = (keyPair?.publicKey() as? NSData)?.prependKeyType() as? Data
```
roberthein: Shouldn't this be:
`let keyToSign = (keyPair?.publicKey() as NSData?)?.prependKeyType() as Data?`
I'm seeing warnings.

Toshi/Models/UserBootstrapParameters.swift
```diff
@@ -93,19 +93,17 @@ public class UserBootstrapParameter {
 
         identityKey = ((identityKeyPair.publicKey() as NSData).prependKeyType() as Data).base64EncodedString()
         lastResortPreKey = storageManager.getOrGenerateLastResortKey()
-        prekeys = storageManager.generatePreKeyRecords() as! [PreKeyRecord]
+        prekeys = (storageManager.generatePreKeyRecords() as? [PreKeyRecord]) ?? [PreKeyRecord]()
```
roberthein: Is it not possible to make `prekeys` optional? Else you can just default to `[]` instead of `[PreKeyRecord]() `.

Toshi/Models/UserBootstrapParameters.swift
```diff
@@ -93,19 +93,17 @@ public class UserBootstrapParameter {
...
-        let keyToSign = (keyPair.publicKey() as NSData).prependKeyType()! as Data
-        let signature = Ed25519.sign(keyToSign, with: identityManager.identityKeyPair())! as Data
+        let keyPair = Curve25519.generateKeyPair()
+        let keyToSign = (keyPair?.publicKey() as? NSData)?.prependKeyType() as? Data
+        let signature = Ed25519.sign(keyToSign, with: identityManager.identityKeyPair()) as Data
 
-        let signedPK = SignedPreKeyRecord(id: Int32(0), keyPair: keyPair, signature: signature, generatedAt: Date())!
-
-        signedPrekey = signedPK
+        signedPrekey = SignedPreKeyRecord(id: Int32(0), keyPair: keyPair, signature: signature, generatedAt: Date()) ?? SignedPreKeyRecord()
```
roberthein: I don't like the fallback to `SignedPreKeyRecord()` here, but I also see it's hard to prevent.

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -178,26 +178,23 @@ final class ChatViewModel {
         // If changes do not affect current view, update and return without updating collection view
         // TODO: Since this is used in more than one place, we should look into abstracting this away, into our own
         // table/collection view backing model.
-        let viewConnection = uiDatabaseConnection.ext(TSMessageDatabaseViewExtensionName) as? YapDatabaseViewConnection
-        if let hasChangesForCurrentView = viewConnection?.hasChanges(for: notifications) as Bool?, hasChangesForCurrentView == false {
+        // swiftlint:disable force_cast
+        let viewConnection = uiDatabaseConnection.ext(TSThreadDatabaseViewExtensionName) as! YapDatabaseViewConnection
```
marijnschilling: Yeah I had some problems with this method, in one instance it wasnt working when optional downcasting, only when forcing. But I should give it another look in the cases where it was already unforced, i just added it there because i thought i had to change it back.

marijnschilling:
🎉  Thanks for the review RobertHein, I will look into it at the end of the day ✌️

Toshi/Controllers/Messaging/ChatInteractor.swift
yuliaveres: it should not be optional.

please, would you mind to change the method to receive `TSMessage`, not `TSInteraction`?.

We deal with messages only here, so no point expect a call and doing not needed casting.
Thus the small change will be needed in `ChatViewModel.swift` 202 and 284 to cast to `TSMessage` (just same as 221). Then, the part of condition in 287 `if let message = interaction as? TSMessage` won't be needed.
thx

Toshi/Controllers/Messaging/ChatViewModel.swift
yuliaveres: it is safe to do this force cast here, as documented, the method **Creates or fetches the extension with the given name.** , returns `YapDatabaseExtensionConnection`, the one `YapDatabaseViewConnection` inherits from

Toshi/Controllers/Messaging/ChatsController.swift
yuliaveres: the changes are being processed in three different `uiDatabaseConnection`-s, lets leave it as it is

Toshi/Extensions/YapDatabaseViewConnection+Additions.swift
```diff
@@ -0,0 +1,30 @@
...
+        var messageRowChanges = NSArray()
+        var sectionChanges = NSArray()
+
+        self.getSectionChanges(&sectionChanges, rowChanges: &messageRowChanges, for: notifications, with: mappings)
+
+        let yapDatabaseViewRowChange = messageRowChanges as? [YapDatabaseViewRowChange]
+        let yapDatabaseViewSectionChange = sectionChanges as? [YapDatabaseViewSectionChange]
```
yuliaveres: u can definitely force here, it would make much more sense. u create empty arrays for row/section changes, pass their references to `getSectionChanges` and it fills those with needed content or leaves empty. those will always be `YapDatabaseViewRowChange ` and `YapDatabaseViewSectionChange `. So why returning optional if we 100% have the collections and causing not needed further `guard`-s or `if let` ?  :)

yuliaveres: but the method is a great addition 👍

Toshi/Models/UserBootstrapParameters.swift
yuliaveres: also here, not needed to make odd optional chaining in case when we are sure about result, just for the sake of **not force unwrapping**, imo 
`storage manager` generated prekey records 100 times:
`PreKeyRecord *record = [[PreKeyRecord alloc] initWithId:preKeyId keyPair:keyPair];
[preKeyRecords addObject:record];`

so why we are not sure the result will be `[PreKeyRecord]`?

Toshi/Models/UserBootstrapParameters.swift
yuliaveres: same here, `generateKeyPair()` , `publicKey()`, `prependKeyType()` results can never be nil, just force it. the pod 25519 misses nullability specifiers 🤷‍♂️

Toshi/Models/UserBootstrapParameters.swift
yuliaveres: sorry, but it is not correct. 
1 - as soon as we have signature - `SignedPreKeyRecord` is going to be initialized **for sure.** (what we have in 104).
2 - creating empty `SignedPreKeyRecord ` without a signature is not allowed, and even if that would happen (never will happen here though) - the assertion won't be satisfied:
`NSAssert(FALSE, @"Signed PreKeys need a signature");`

Toshi/Models/UserBootstrapParameters.swift
marijnschilling: Yup i just force cast here as an exception. It's a good exception!

Toshi/Extensions/YapDatabaseViewConnection+Additions.swift
marijnschilling: Yes very good point, will force downcast here. For this its extra nice that this happens in a separate method ✌️ And not 3 times in the code.

Toshi/Controllers/Messaging/ChatViewModel.swift
marijnschilling: Yup i think since it's safe to use force downcast here i can better use it in all three cases, to be consistent.

Toshi/Controllers/Messaging/ChatsController.swift
marijnschilling: We might want to look in abstracting this away more than just the `getChangesFor` method. But I'm not really deep enough into this material. So if we decide we want to do that it's maybe something for another PR ✌️

PR #426, opened 2017-08-16
Test ethereum converter

marijnschilling:
Tested using the [quick](https://github.com/Quick/Quick) and [nimble](https://github.com/Quick/Nimble) framework to be able to provide more context.

I just created the context where the exchange rate is 100.0. 

I will wait for the cases where the Converter isn't working as expected to add tests for that.

One thing i do not understand is that i can't give in a higher `wei` value int he method as it is right now, because it overflows. This comes down to 1 ETH max, which seems a bit low, or is that one of the things that is broken?


marijnschilling:
I fixed the tests to not break over Circle CI using a different locale than I do. It expects `100.00 USD` while I expect `100,00 USD`. 

I could improve by writing the test with a regular expression maybe, to define i want either a comma or a dot in that place. Does anyone know a way to find out how to write that regular expression? 😬

yuliaveres:
@marijnschilling  do i get u correctly, u need to get current locale decimal separator?

Tests/EthereumConverterTests.swift
```diff
@@ -0,0 +1,84 @@
...
+                    let dollarSting = String(format: "$100%@00 USD", TokenUser.current?.cachedCurrencyLocale?.decimalSeparator ?? ".")
+                    it("has the right values") {
+                        expect(ethereumValueString.string).to(beginWith(dollarSting))
+                        expect(ethereumValueString.string).to(endWith("1.0000 ETH"))
+                    }
+
+                    it("has the right attributes") {
+                        let fiatAttributes = ethereumValueString.attributes(at: 0, effectiveRange: nil)
+                        expect(fiatAttributes[NSForegroundColorAttributeName] as? UIColor).to(equal(Theme.darkTextColor))
+                        expect(fiatAttributes[NSFontAttributeName] as? UIFont).to(equal(UIFont.systemFont(ofSize: 19.0)))
```
yuliaveres: it is like UI tests 🍪

PR #428, opened 2017-08-17
Load messages by chunks

yuliaveres:
**WIP** cause not QAd and checked if everything works good


Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -753,6 +757,12 @@ extension ChatController: ChatViewModelOutput {
         self.adjustToLastMessage()
     }
 
+    func didRequireScrollToBottom() {
+        DispatchQueue.main.asyncAfter(seconds: 0.5) {
```
heumn: I always get an icky feeling when I see delays in code. That is often the symptom that I have not really solved the root problem of something. UIKit is of course _really_ weird sometimes and this is needed. But when we match that weirdness it would be nice for the next guy to know why this day was added? :ok_man:

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -145,17 +147,28 @@ final class ChatViewModel {
...
+        let options = YapDatabaseViewRangeOptions.flexibleRange(withLength: 20, offset: 0, from: .end)!
+        options.growOptions = .onBothSides
+
+        return options
+    }()
+
     fileprivate lazy var mappings: YapDatabaseViewMappings = {
-        let mappings = YapDatabaseViewMappings(groups: [self.thread.uniqueId], view: TSMessageDatabaseViewExtensionName)
-        mappings.setIsReversed(true, forGroup: TSInboxGroup)
+        return YapDatabaseViewMappings(groups: [self.thread.uniqueId], view: TSMessageDatabaseViewExtensionName)
```
heumn: This return is actually implied in swift, so not need! Isnt that neat? 😄

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -145,17 +147,28 @@ final class ChatViewModel {
...
+    }()
 
-        return mappings
+    fileprivate lazy var loadedMappings: YapDatabaseViewMappings = {
+        return YapDatabaseViewMappings(groups: [self.thread.uniqueId], view: TSMessageDatabaseViewExtensionName)
```
heumn: Same here, no need for return 💯

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -145,17 +147,28 @@ final class ChatViewModel {
...
     }()
 
     fileprivate lazy var editingDatabaseConnection: YapDatabaseConnection? = {
         self.storageManager?.newDatabaseConnection()
     }()
 
+    fileprivate var messagesCount: UInt = 0
+    fileprivate var loadedMessagesCount: UInt = 0
```
heumn: 👌  👌  (for use of the _correct_ int)

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -170,6 +183,19 @@ final class ChatViewModel {
         }
     }
 
+    func updateMessagesRange(from indexPath: IndexPath? = nil) {
```
heumn: 🤔  looks like this should be `fileprivate`?

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -170,6 +183,19 @@ final class ChatViewModel {
...
+        loadNextChunk()
+    }
+
+    fileprivate func loadNextChunk(completion: (() -> Void)? = nil) {
```
heumn: I would probably advise you to use defaults and not optionals here. I try to limit my usage of optionals to when it is really needed and rather use default values (because that is what Apple says we should do  if it makes sense) :)

Toshi/Controllers/Messaging/ChatViewModel.swift
```diff
@@ -271,17 +309,37 @@ final class ChatViewModel {
...
+
+    fileprivate func nextChunkSize() -> UInt {
+        let notLoadedCount = messagesCount - loadedMessagesCount
+        let numberToLoad = min(notLoadedCount, 50)
+
+        return numberToLoad
+    }
+
+    func loadMessages(_ completion: (() -> Void)? = nil) {
+        uiDatabaseConnection.asyncRead { [weak self] transaction in
```
heumn: misssing self here :)

Toshi/Controllers/Messaging/ChatViewModel.swift
yuliaveres: oh yes, you already mentioned this, changing

Toshi/Controllers/Messaging/ChatViewModel.swift
yuliaveres: nope, controller calls it when the last/first item is going to appear on screen

Toshi/Controllers/Messaging/ChatViewModel.swift
yuliaveres: it is 👍

Toshi/Controllers/Messaging/ChatViewModel.swift
heumn: Got it :)

Toshi/Controllers/Messaging/ChatViewModel.swift
heumn: no need any more 🎉

Toshi/Controllers/Messaging/ChatController.swift
yuliaveres: Removed it at all, as not needed 😉

Toshi/Controllers/Messaging/ChatViewModel.swift
yuliaveres: also removed completions as not needed 😄

PR #430, opened 2017-08-17
Name viewConnections according to their type

marijnschilling:
Because of the mistake i made earlier with using the wrong database view extension name, i thought i would add more context to the `viewConnections` by adding the type to their name. This way I hope these kind of mistakes will be easier to prevent ✌️


heumn:
Great job. Names are hard... 😄

PR #431, opened 2017-08-17
Add exchange rate service api and currency selection menu.

tristan:
TODOs:
- [x] update balance in settings page after selecting a new currency
- [x] sort currencies to put popular currencies at the top
- [x] filter currencies to remove BTC (maybe filter to remove any that don't match a valid `Locale.availableIdentifiers`)
- [x] add cache for `TokenUser.getLocalCurrencyLocale` so it doesn't have to run the map+filter each time
- [x] currency selection gets reset to default sometimes on launch
- [x] keep user's currency selection after sign out and sign in

a89ee2066889c6128ed16fca07cd242c26f60596 is a bit of a crazy commit
i was having issues with a user object being loaded in the browse/chat history pages that overwrote my local user details (verified state and local currency), so to fix this i decided to separate the profile data from these settings.

I decided to remove the storage of the whole profile object in the `storedUserKey` key, and instead just use the `storedContactKey` values instead (as there's no reason to duplicate this data), and created a `localUserSettings` key to store the verified and local currency settings.

To help with loading these, i also store the current local user's address (`currentLocalUserAddressKey`). I feel this should be unnecessary since it can probably just come from `Cereal`? but i didn't really figure out how to make that work.

all this meant a bunch of migration code is not in the `retrieveCurrentUserFromStore` method, it seems like this should really live somewhere else?


roberthein:
@yuliaveres please resolve conflicts.

yuliaveres:
sure !!! a lot of merges today :)

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
```diff
@@ -26,14 +26,23 @@ class ChatsFloatingHeaderView: UIView {
 
     static let height = CGFloat(48)
 
-    private(set) lazy var balanceLabel: UILabel = {
+    private(set) lazy var fiatValueLabel: UILabel = {
```
yuliaveres: @marijnschilling  https://en.wikipedia.org/wiki/Fiat_money
🤑

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -915,14 +915,15 @@ extension ChatController: PaymentControllerDelegate {
...
-                "value": valueInWei.toHexString,
-                "destinationAddress": Cereal.shared.paymentAddress
-            ]
-            
-            let paymentRequest = SofaPaymentRequest(content: request)
-            viewModel.interactor.sendMessage(sofaWrapper: paymentRequest)
+        let request: [String: Any] = [
+        "body": "Request for \(EthereumConverter.balanceAttributedString(forWei: valueInWei, exchangeRate: ExchangeRateAPIClient.shared.exchangeRate).string).",
+        "value": valueInWei.toHexString,
+        "destinationAddress": Cereal.shared.paymentAddress
```
yuliaveres: i dnt know 🤔  should it ?

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
heumn: Fiat money is a currency without intrinsic value (rekt). Like the USD 😂

heumn: and I am not kidding. That is the definition.

https://en.wikipedia.org/wiki/Fiat_money

roberthein: Also check this link:
https://en.wikipedia.org/wiki/Fiat_money

yuliaveres: hahaha
those different people

Toshi/APIClients/ExchangeRateAPIClient.swift
```diff
@@ -0,0 +1,122 @@
...
+                completion(nil)
+                return
+            }
+
+            self.teapot.get("/v1/rates/ETH/\(code)") { (result: NetworkResult) in
+                switch result {
+                case .success(let json, _):
+                    guard let json = json?.dictionary,
+                        let usd = json["rate"] as? String,
+                        let doubleValue = Double(usd) as Double? else {
```
yuliaveres: we have completion + return in those :/ so i think it is better to put in separate line each ?

Toshi/APIClients/ExchangeRateAPIClient.swift
```diff
@@ -0,0 +1,122 @@
...
+
+        Timer.scheduledTimer(withTimeInterval: 300.0, repeats: true) { [weak self] _ in
+            self?.updateRate()
+        }
+    }
+
+    public func updateRate(_ completion: @escaping ((_ rate: Decimal?) -> Void) = { _ in }) {
+        getRate { rate in
+            if rate != nil {
+                Yap.sharedInstance.insert(object: rate, for: ExchangeRateAPIClient.collectionKey)
```
yuliaveres: it is with purpose of not making huge amount of requests (we update data each 5 minutes)
 + in future for offline

Toshi/Controllers/Messaging/ChatInteractor.swift
```diff
@@ -204,7 +204,7 @@ final class ChatsInteractor: NSObject {
             } else if let payment = SofaWrapper.wrapper(content: interaction.body ?? "") as? SofaPayment {
                 message.messageType = "Actionable"
                 message.attributedTitle = NSAttributedString(string: "Payment sent", attributes: [NSForegroundColorAttributeName: Theme.outgoingMessageTextColor, NSFontAttributeName: Theme.medium(size: 17)])
-                message.attributedSubtitle = NSAttributedString(string: EthereumConverter.balanceAttributedString(forWei: payment.value, exchangeRate: EthereumAPIClient.shared.exchangeRate).string, attributes: [NSForegroundColorAttributeName: Theme.outgoingMessageTextColor, NSFontAttributeName: Theme.regular(size: 15)])
+                message.attributedSubtitle = NSAttributedString(string: EthereumConverter.balanceAttributedString(forWei: payment.value, exchangeRate: ExchangeRateAPIClient.shared.exchangeRate).string, attributes: [NSForegroundColorAttributeName: Theme.outgoingMessageTextColor, NSFontAttributeName: Theme.regular(size: 15)])
```
roberthein: We could add some syntactic sugar that turns stuff like `ExchangeRateAPIClient.shared.exchangeRate` to `ExchangeRate.current`. 😏

Toshi/Controllers/Settings/CurrencyPicker.swift
```diff
@@ -0,0 +1,127 @@
...
+                .sorted(by: strongSelf.sortCurrencies)
+
+            strongSelf.tableView.reloadData()
+            strongSelf.tableView.scrollToRow(at: strongSelf.currentLocalCurrencyIndexPath, at: .middle, animated: false)
+        }
+    }
+
+    func addSubviewsAndConstraints() {
+        view.addSubview(tableView)
+        tableView.topAnchor.constraint(equalTo: view.topAnchor).isActive = true
```
yuliaveres: 👀

Toshi/Controllers/Settings/CurrencyPicker.swift
```diff
@@ -0,0 +1,127 @@
...
+        tableView.leftAnchor.constraint(equalTo: view.leftAnchor).isActive = true
+        tableView.rightAnchor.constraint(equalTo: view.rightAnchor).isActive = true
+        tableView.bottomAnchor.constraint(equalTo: view.bottomAnchor).isActive = true
+    }
+
+    func sortCurrencies(a: Currency, b: Currency) -> Bool {
```
yuliaveres: sure

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
marijnschilling: haha hanks guys! 😂 I was actually more looking for what it means in this context than the abstract definition. Why are we using it here in the name for the (what used to be) balance label?

Toshi/APIClients/ExchangeRateAPIClient.swift
marijnschilling: No i'm talking about the multiple guard statements. Completion and return is fine.

i mean this:
```
guard let json = json?.dictionary, let usd = json["rate"] as? String, let doubleValue = Double(usd) as Double? else {
```

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
yuliaveres: @marijnschilling  cause previously the value string was single attributed with both fiat + ethereum values,
now I had to make them separate

Toshi/APIClients/ExchangeRateAPIClient.swift
yuliaveres: ok, will look for those, thx

Toshi/Controllers/Messaging/ChatsFloatingHeaderView.swift
marijnschilling: yesss but what does fiat mean in this context 😭 So it's money? Shouldn't it be called `userCurrencyValueLabel`? Or something that is more commonly used?

yuliaveres: yes, money. and it is widely used in the project

marijnschilling:
👍

marijnschilling:
✅

PR #432, opened 2017-08-18
Reload browse view on each appear

yuliaveres:


faymontage:
The purpose here is to support the signed in user's username being changed from outside of the client, which is not accomplished by this change. I'll get you some more details.

yuliaveres:
no, the purpose is not one you mentioned here. sorry for not providing the description.
the purpose was just to reload Browse page datasource 😄 
@faymontage

PR #435, opened 2017-08-18
Remove test target check from app delegate

marijnschilling:
It's now checked in main, and it will open the TestAppDelegate in case of tests. Just forgot to remove this piece of code.


yuliaveres:
:shipit:

PR #436, opened 2017-08-18
Use QRCodeIntent to handle TabBarController scanner results

faymontage:
All knowledge of QR code formats/contents is removed from TabBarController and is delegated to QRCodeIntent instead.

By doing so, it has gained support for a number of new QR formats, as well as proper handling of values in payment requests.

Also fixed are some cases where the scanning sound would not be triggered, and where the scanner would get stuck because scanning was not restarted even when a scan had failed to successfully generate an action/dismissal.

Ignored for now is the case of an Ethereum address without a payment request value, as we need to design the flow that allows the user to enter a value for a payment.


Toshi/Controllers/TabBarController.swift
```diff
@@ -194,93 +194,80 @@ extension TabBarController: ScannerViewControllerDelegate {
...
+    private func proceedToPayment(username: String, weiValue: String?) {
         idAPIClient.retrieveContact(username: username) { contact in
             if let contact = contact as TokenUser? {
-                userInfo = contact.userInfo
-                parameters["to"] = contact.paymentAddress
+                var parameters = ["from": Cereal.shared.paymentAddress, "to": contact.paymentAddress]
+                parameters["value"] = weiValue
 
                 guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
                 appDelegate.contactsManager.refreshContacts()
```
yuliaveres: do we add user here? if not - we should not refresh them, just no reason

yuliaveres: if only payment - please remove 😉

Toshi/Controllers/TabBarController.swift
```diff
@@ -194,93 +194,80 @@ extension TabBarController: ScannerViewControllerDelegate {
...
+            scannerController.displayPaymentConfirmation(userInfo: userInfo, parameters: parameters)
+            SoundPlayer.playSound(type: .scanned)
+        } else {
+            scannerController.startScanning()
+        }
+    }
 
-                        parameters["value"] = valueInWei.toHexString
+    private func viewContact(with contactName: String) {
+        idAPIClient.retrieveContact(username: contactName) { contact in
```
yuliaveres: and here actually i think we should refresh contacts

PR #438, opened 2017-08-21
Remove force unwrap

marijnschilling:
I decided not to introduce the rule to not use force unwrap. This rule is by default disabled in Swift-lint and will only give a warning when enabled. So it clearly has a a different gravity then the by-default-enabled error-throwing rule to not use force downcast.

It mostly needs to be used on URL inits and handling JSON or native Obj-C objects. 

I did however went over the code and removed force unwraps where possible!

https://trello.com/c/faE9IGWn/239-438-add-no-force-unwrapping-rule-in-swift-lint-and-rewrite-these-cases-in-project


Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -211,7 +211,7 @@ open class FavoritesController: SweetTableController {
     @discardableResult
     fileprivate func registerTokenContactsDatabaseView() -> Bool {
         // Check if it's already registered.
-        guard Yap.sharedInstance.database!.registeredExtension(TokenUser.viewExtensionName) == nil else { return true }
+        guard let database = Yap.sharedInstance.database, database.registeredExtension(TokenUser.viewExtensionName) == nil else { return true }
```
yuliaveres: i think it is definitely not desired case, if we dnt have db here, let throw fatal error with message :)

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -229,7 +229,7 @@ open class FavoritesController: SweetTableController {
 
         let databaseView = YapDatabaseView(grouping: viewGrouping, sorting: viewSorting, versionTag: "1", options: options)
 
-        return Yap.sharedInstance.database!.register(databaseView, withName: TokenUser.viewExtensionName)
+        return database.register(databaseView, withName: TokenUser.viewExtensionName)
```
yuliaveres: same here 😉

Toshi/Models/Cereal.swift
```diff
@@ -63,8 +63,8 @@ public class Cereal: NSObject {
 
     // restore from local user or create new
     public override init() {
-        if let words = Yap.sharedInstance.retrieveObject(for: Cereal.privateKeyStorageKey) as? String {
-            mnemonic = BTCMnemonic(words: words.components(separatedBy: " "), password: nil, wordListType: .english)!
+        if let words = Yap.sharedInstance.retrieveObject(for: Cereal.privateKeyStorageKey) as? String, let mnemonicValue = BTCMnemonic(words: words.components(separatedBy: " "), password: nil, wordListType: .english) {
```
yuliaveres: i dnt think those conditions are homogeneous / equal or how to say properly;
first is the check if account is secured and words list is stored in db, it just defines the flow; both options are accepted.
if true - we SHOULD be able to initialize mnemonic value , otherwise smth is wrong.
would u mind pease to separate second condition and if not satisfied - throw a fatal error with message `Entropy has incorrect size or wordlist is not supported`

Toshi/Models/Yap.swift
```diff
@@ -107,7 +107,9 @@ public final class Yap: NSObject, Singleton {
             let keychain = KeychainSwift()
             keychain.synchronizable = false
 
-            return keychain.getData(UserDB.password)!
+            guard let userDatabasePassword = keychain.getData(UserDB.password) else { fatalError("No database password found in keychain") }
```
yuliaveres: that was missing! 👍

Toshi/Models/Yap.swift
```diff
@@ -150,7 +152,8 @@ public final class Yap: NSObject, Singleton {
         deleteFileIfNeeded(at: UserDB.shmFilePath)
 
         let keychain = KeychainSwift()
-        let currentPassword = keychain.getData(UserDB.password)!
+        guard let currentPassword = keychain.getData(UserDB.password) else { fatalError("No database password found in keychain") }
```
yuliaveres: could u please specify the problem by adding `while db file exits` ?

Toshi/Models/Notifications/MessageFetcherJob.swift
```diff
@@ -107,7 +107,7 @@ class MessageFetcherJob: NSObject {
             }
         }()
 
-        let envelopes = messageDicts.map { buildEnvelope(messageDict: $0) }.filter { $0 != nil }.map { $0! }
+        let envelopes = messageDicts.flatMap { buildEnvelope(messageDict: $0) }.filter { $0 != nil }
```
marijnschilling: Ha yeah, probably! Found it a weird function to begin with, not sure what the intentions were :) Wonder though if i can test easily if this has the wanted result... 🤔

Toshi/Controllers/Favorites/FavoritesController.swift
marijnschilling: Shoudl this check `database.registeredExtension(TokenUser.viewExtensionName) == nil` also throw this fatal error, or should that be handled differently?

Toshi/Controllers/Favorites/FavoritesController.swift
marijnschilling: What do you mean with same here? It will already have thrown the fatalError that it cant egt the database on top of the function right? So here i thought i could jut use the unwrapped value then.

Toshi/Controllers/Favorites/FavoritesController.swift
yuliaveres: no no, just for database being present

Toshi/Controllers/Favorites/FavoritesController.swift
yuliaveres: oh right, i missed it is the same place, so good 👍

Toshi/Controllers/Favorites/FavoritesController.swift
marijnschilling: oke, i keep the other to return true :)

PR #439, opened 2017-08-21
Remove force try

marijnschilling:
I enabled the (default) force try rule for SwiftLint and removed it where it was used.

Things that I would like to be looked at in the PR are:
- [x] How to handle the guard when the payload string can not be made, should i return an error or not?
- [x] I made contact and image cache an optional value, I wonder if this doesn't have any unforeseen implications

https://trello.com/c/apxETbVF/243-439-add-no-force-try-rule-in-swift-lint-and-rewrite-these-cases-in-project


Toshi/APIClients/idAPIClient.swift
```diff
@@ -33,7 +33,7 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
 
     public var teapot: Teapot
 
-    private var contactCache = try! Cache<TokenUser>(name: "tokenContactCache")
+    private var contactCache = try? Cache<TokenUser>(name: "tokenContactCache")
```
roberthein: I don't like that this variable is optional, maybe we can do something like this instead:
```
private lazy var contactCache: Cache<TokenUser> = {
    do {
        let cache = try Cache<TokenUser>(name: "tokenContactCache")
        return cache
    } catch {
        fatalError()
    }
}()
```
What do you think @yuliaveres?

Toshi/Avatars/AvatarManager.swift
```diff
@@ -21,7 +21,7 @@ final class AvatarManager: NSObject, CacheExpiryDefault {
 
     static let shared = AvatarManager()
 
-    private var imageCache = try! Cache<UIImage>(name: "imageCache")
+    private var imageCache = try? Cache<UIImage>(name: "imageCache")
```
roberthein: Same goes for `imageCache` if it's optional like this, we will never know when something goes wrong.

Toshi/MessagesUI/Cells/MessagesTextCell.swift
```diff
@@ -44,7 +44,7 @@ class MessagesTextCell: MessagesBasicCell {
         }
     }
 
-    fileprivate let usernameDetector = try! NSRegularExpression(pattern: " ?(@[a-zA-Z][a-zA-Z0-9_]{2,59}) ?", options: [.caseInsensitive, .useUnicodeWordBoundaries])
+    fileprivate let usernameDetector = try? NSRegularExpression(pattern: " ?(@[a-zA-Z][a-zA-Z0-9_]{2,59}) ?", options: [.caseInsensitive, .useUnicodeWordBoundaries])
```
roberthein: I don't think `usernameDetector` should be optional.

Toshi/APIClients/ChatAPIClient.swift
```diff
@@ -60,7 +60,8 @@ public class ChatAPIClient: NSObject {
             let parameters = UserBootstrapParameter()
             let path = "/v1/accounts/bootstrap"
             let payload = parameters.payload
-            let payloadString = String(data: try! JSONSerialization.data(withJSONObject: payload, options: []), encoding: .utf8)!
+
+            guard let data = try? JSONSerialization.data(withJSONObject: payload, options: []), let payloadString = String(data: data, encoding: .utf8) else { return }
```
yuliaveres: i think this way is good to handle it;
the exception might be thrown only if json (`payload`) is invalid, what I guess is not the case of `UserBootstrapParameter`s ? but `nil` might be returned if some error occurs, no idea which in particular 🤔 
but anyway, i guess the registration should not be just interrupted right? 
we should write log at least + call completion handler

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -113,9 +113,8 @@ public class EthereumAPIClient: NSObject {
                 "signature": transactionSignature
             ]
 
-            let payloadString = String(data: try! JSONSerialization.data(withJSONObject: params, options: []), encoding: .utf8)!
+            guard let data = try? JSONSerialization.data(withJSONObject: params, options: []),  let payloadString = String(data: data, encoding: .utf8) else { return}
```
yuliaveres: i think my previous comment count for all cases, wdyth ? we should now that request even hasn't started for the reason, and of course execute completion

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -237,9 +236,13 @@ public class EthereumAPIClient: NSObject {
         let cereal = Cereal.shared
         let path = "/v1/apn/register"
         let address = cereal.address
-
         let params = ["registration_id": appDelegate.token, "address": cereal.paymentAddress]
-        let payloadString = String(data: try! JSONSerialization.data(withJSONObject: params, options: []), encoding: .utf8)!
+
+        guard let data = try? JSONSerialization.data(withJSONObject: params, options: []), let payloadString = String(data: data, encoding: .utf8) else {
+            completion?(false)
```
yuliaveres: i am wondering may be we should add string description or error with description to completion handler apart from Bool indicating success ? so while debugging - we know what went wrong ?

Toshi/APIClients/idAPIClient.swift
```diff
@@ -209,8 +210,10 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
         fetchTimestamp { timestamp in
             let cereal = Cereal.shared
             let path = "/v1/user"
-            let payload = try! JSONSerialization.data(withJSONObject: userDict, options: [])
-            let payloadString = String(data: payload, encoding: .utf8)!
+            guard let payload = try? JSONSerialization.data(withJSONObject: userDict, options: []), let payloadString = String(data: payload, encoding: .utf8) else {
+                completion(false, "Invalid payload")
```
yuliaveres: in my previous comment I have meant smth like this 😉

Toshi/APIClients/idAPIClient.swift
yuliaveres: we can't got without this cache;
please, do try and catch error along with interrupting execution. smth is not ok in this case 😉

Toshi/Avatars/AvatarManager.swift
yuliaveres: same as with contacts cache :)

Toshi/MessagesUI/Cells/MessagesTextCell.swift
yuliaveres: not that critical, but why might we accept if detector initializer failed? we should at least log error i think

Toshi/Models/Users/TokenUser.swift
```diff
@@ -110,8 +110,8 @@ public class TokenUser: NSObject, NSCoding {
         return address == Cereal.shared.address
     }
 
-    public var JSONData: Data {
-        return try! JSONSerialization.data(withJSONObject: asDict, options: [])
+    public var JSONData: Data? {
+        return try? JSONSerialization.data(withJSONObject: asDict, options: [])
```
yuliaveres: this can't be nil, we never might want to access non-existent user JSONData, 
please, whether leave it forced or add fatal error with a message

Toshi/APIClients/EthereumAPIClient.swift
marijnschilling: yes, I think creating our own Error for this could work. But not all completion handlers of APICalls return an error at this moment. Should I adjust that in those cases? Or just print what went wrong?

yuliaveres: i think we can see depending on the context what is wrong, like here the call wasn't even made cause we have no payload, right? so just pass a string saying the reason.
if api call failed - then error description if there is one, or our own description that request failed. but mostly i think requests return error if smth went wrong ?

marijnschilling: This is just for cases when the payload couldn't be created. But i can't always pass a string, sometimes the completion just return true or false, sometimes an error.

marijnschilling: How do you even create a simple swift Error? Nothing to be found about that :(

yuliaveres: hmmmm, lets see

yuliaveres: have u seen this ? https://stackoverflow.com/questions/40671991/generate-your-own-error-code-in-swift-3

yuliaveres: but anyway @marijnschilling  that is not super crucial to throw error in articular, an if time consuming - just add string
what we need finally is just description

Toshi/Models/Users/TokenUser.swift
marijnschilling: Im changing this but I'm not really sure what this JSONData actually is or does. I see it's made form an `asDict`. Also don't know what this means. Do you happen to know this? Then I can write meaningful message and change the var names.

yuliaveres: it is just Data from Dictionary, in this format user information is stored in Yap

marijnschilling: oke so it's `userInfo` and `userInfoJSON`?

yuliaveres: so `asDict` -  is everything as` [String: Any]`
and `JSONData` is that turned into `Data`

honestly, i think as dict makes sense, cause since long time ago userInfo is something widely used and may cause confusing ?  you want to change naming right ? or i misunderstood :)

Toshi/APIClients/EthereumAPIClient.swift
yuliaveres: lets add string, just like here:
screenshot at aug 22 16-03-35

Toshi/Models/Users/TokenUser.swift
marijnschilling: oooh it's the whole struct `as` a Dict? and as JSON data? No i didnt get that... Would assume it be class methods in that case. I thoughs `as` was just some abbreviation that i'm not in on :)

I would like to change the name yes, cause it's unclear what data this is about.

yuliaveres: not struct - just variables of a user - asDict: [String: Any] and JSONData: Data .

look if the following makes more sense. f.e. : 

`let currentUset = TokenUser.current`
`currentUser.json / currentUser.dict`
`currentUser.data `

Toshi/APIClients/EthereumAPIClient.swift
marijnschilling: Than i have to add a lot of nil values to all the completion handlers where this string is not being used though 🤔  or is that oke?

yuliaveres: it is fine 
`completion(true, nil)`
`completion(false, message)`

yuliaveres: we can't use default values in tuples 😢

marijnschilling: I know I was trying to do that XD

Toshi/Controllers/Favorites/ContactController.swift
```diff
@@ -434,7 +434,7 @@ extension ContactController: ActivityIndicating {
 extension ContactController: RateUserControllerDelegate {
     func didRate(_ user: TokenUser, rating: Int, review: String) {
         dismiss(animated: true) {
-            RatingsClient.shared.submit(userId: user.address, rating: rating, review: review) {
+            RatingsClient.shared.submit(userId: user.address, rating: rating, review: review) { _ in
```
yuliaveres: si of smth went wrong lets alert user ?

marijnschilling: I added the alert in the rating client, because the success and failure are also handled there as alerts.

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -90,18 +93,18 @@ public class EthereumAPIClient: NSObject {
...
                         print(json ?? "")
-                        completion(json, nil)
+                        completion(json, nil, "json: \(json) response: \(response)")
                     case .failure(let json, let response, let error):
                         print(response)
                         print(json ?? "")
                         print(error)
                         guard let jsonError = (json?.dictionary?["errors"] as? [[String: Any]])?.first else {
-                            completion(nil, error)
+                            completion(nil, error, "json: \(json) response: \(response)")
```
yuliaveres: you have warnings here

marijnschilling: Yeah if i do it like this `json ?? ""` I get the error `/Users/marijnschilling/Projects/toshi-ios-client/Toshi/APIClients/EthereumAPIClient.swift:102:69: Cannot convert value of type 'String' to expected argument type 'RequestParameter'` 

Shall i just remove the json?

yuliaveres: i think the warning is about string interpolation, so proper would be `String(describing: json)` .

apart from it, i see we dnt use this message in handlers, it is _ everywhere so maybe dnt add it to this method  ?

PR #440, opened 2017-08-21
Change types, objects and copy to "passphrase"

roberthein:
Changed all localization from `backup phrase` to `passphrase`.
Renamed all types and objects from `backup phrase` to `passphrase`.
Renamed all files and folders from `backup phrase` to `passphrase`.


marijnschilling:
Maybe we should add the original copy as a comment.

Or write something that for example `Localizable("Request payment") automatically looks for a localizable string with the key `request_payment`. (and maybe has a max amount of words to use for this for really long localizable strings)

roberthein:
With `Localized` it's already possible to add a `comment`, you can do `Localized("key", comment: "comment")`
But adding the copy there just clutters the code in my opinion and it's never a guarantee that it's actually in sync with the **Localizable.strings** file.

PR #445, opened 2017-08-24
Apps api client tests

marijnschilling:
Added tests for the apps Api client using Quick & Nimble. 

Curious what you think about the clarity of the test contexts and cases!

And wether it covers all the cases to check wether the appsAPIClient is behaving properly. 

screenshot 2017-08-24 13 47 16


Tests/Mocks/AppsAPIClient/search/?query=Test&limit=100.json
```diff
@@ -0,0 +1,54 @@
...
+      "payment_address": "0x011c6dd9565b8b83e6a9ee3f06e89ece3251ef2f",
+      "public": 0,
+      "reputation_score": "2.7",
+      "review_count": 2,
+      "token_id": "0x8f9bdb7f562ccdedf3c24cf25e9cece9df62138b",
+      "toshi_id": "0x8f9bdb7f562ccdedf3c24cf25e9cece9df62138b",
+      "username": "testUsername"
+    },
+    {
+      "about": "The third most searchest of all the apps",
```
yuliaveres: can we try to test with weird data? what happens if `avatar` is empty or absent (like happened some time before, AvatarManager was expecting it and crashing)? what if backend send us various not expected stuff?

marijnschilling: Hmm I could do that, but in theory I think this are things that the back-end itself should test. Making sure that they don't send us weird stuff :) 

But in case of the empty avatar, if that is something that we should take into account in our app I can add a test for that. Anything else weird you have in mind except the empty or nil avatar?

yuliaveres: `{
 +      "about": "The second most searchest of all the apps",
 +      "avatar": "noturl",
 +      "average_rating": 100,
 +      "is_app": 1,
 +      "location": "",
 +      "name": "TestName",
 +      "payment_address": "",
 +      "public": 0,
 +      "reputation_score": "2.7",
 +      "review_count": 2,
 +      "token_id": "0x8f9bdb7f562ccdedf3c24cf25e9cece9df62138b"
 +    }`

yuliaveres: also, what is no id ? broken data ? :)

yuliaveres: We are responsible to be able to handle different cases, if crash happens - noone cares who is quilty 😉

marijnschilling: I'll try it! It's not about placing blame, its more about not doing unnecessary work ✌️  But these cases seams like we should handle them correctly so i will make a 'weird data test' 😬

marijnschilling:
We decided to add missing data in another PR, since that requires adding error handling in the appsAPIClient itself. So merging this one and opening a new one from there ✌️

PR #446, opened 2017-08-25
Specify continue behavior in guard else

faymontage:
Having only an assert in a guard's else clause is not valid production code, as asserts are evaluated only in debug builds.

This problem was introduced here: https://github.com/toshiapp/toshi-ios-client/commit/2d0301590d3918e194fcba0e4a46869795143d3f#diff-81edbd2a85deec3043a3fc1e26528ff7

To make it valid code - while matching the behavior of the code from before that commit - a continue statement is added to the else clause.


Toshi/Models/Converters/BaseConverter.swift
```diff
@@ -74,7 +74,10 @@ public struct BaseConverter {
 
         if !digits.isEmpty {
             for i in (0 ... (digits.count - 1)).reversed() {
-                guard let digit = stringToInt(digits[i]) else { assert(false, "Invalid digit") }
+                guard let digit = stringToInt(digits[i]) else {
+                    assert(false, "Invalid digit")
+                    continue
```
yuliaveres: i think after assertion continue makes no sense

faymontage: Read the linked commit - what I've done here exactly matches the removed code's behaviour in both debug and production. Am I wrong?

faymontage: To be clear - I'm just fixing a regression so we can build - I'm not giving consideration to how this class may be refactored to be more correct/useful.

roberthein: Why does it not make sense to use `continue` @yuliaveres?

Toshi/Models/Converters/BaseConverter.swift
```diff
@@ -76,7 +76,7 @@ public struct BaseConverter {
             for i in (0 ... (digits.count - 1)).reversed() {
                 guard let digit = stringToInt(digits[i]) else {
                     assert(false, "Invalid digit")
-                    continue
+                    break
```
faymontage: This does not match the behavior from before this regression - are you sure you want to do that?

roberthein: I think it should be `continue`.

yuliaveres: sorry my fault i understood it incorrectly 👎

PR #450, opened 2017-08-29
Tests for the ethereum api client

marijnschilling:
I had to rewrite the `timestamp` function in `EthereumAPIClient` to return an error when this fails, cause I need to catch that error in the tests (otherwise it wouldnt get into the completionblock at all). But I think this is generally a good addition, what if there's some problem getting a connection to the API, we wouldn't have found out about it otherwise.

So what i would like to have checked issss
- [x] Wether the added completionblock in `timestamp` in `EthereumAPIClient` is handled well
- [ ] Wether there are some error states that should be checked. (Now i'm only checking for unauthorized error)
- [x] And of course wether the tests in general make any sense to you 🙃


Tests/EthereumAPIClientTests.swift
```diff
@@ -0,0 +1,130 @@
...
+                            done()
+                        }
+                    }
+                }
+
+                it("gets the balance") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: EthereumAPIClientTests.self), mockFileName: "getBalance")
+                    subject = EthereumAPIClient(mockTeapot: mockTeapot)
+
+                    waitUntil() { done in
```
roberthein: I get a warning here for the empty parentheses.

Tests/EthereumAPIClientTests.swift
```diff
@@ -0,0 +1,130 @@
...
+                        subject.getBalance(address: "0x1ad0bb2d14595fa6ad885e53eaaa6c82339f9b98") { number, error in
```
roberthein: I get a warning here for number not being used.

Tests/EthereumAPIClientTests.swift
```diff
@@ -0,0 +1,130 @@
...
+//swiftlint:disable force_cast
+class EthereumAPIClientTests: QuickSpec {
+
+    override func spec() {
+        describe("the Ethereum API Client") {
+            var subject: EthereumAPIClient!
+
+            context("Happy path 😎") {
+                it("creates an unsigned transaction") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: EthereumAPIClientTests.self), mockFileName: "createUnsignedTransaction")
```
roberthein: ✨ ✨ ✨ ✨ ✨ 
✨ 😍 😍 😍 ✨ 
✨ ✨ ✨ ✨ ✨

Tests/EthereumAPIClientTests.swift
```diff
@@ -0,0 +1,130 @@
...
+                            done()
+                        }
+                    }
+                }
+
+                it("gets the balance") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: EthereumAPIClientTests.self), mockFileName: "getBalance", statusCode: .unauthorized)
+                    subject = EthereumAPIClient(mockTeapot: mockTeapot)
+
+                    waitUntil() { done in
```
roberthein: Empty parentheses here too.

Toshi.xcodeproj/project.pbxproj
```diff
@@ -2902,17 +2907,22 @@
...
+		D197B654F730E653B7C6F444 /* sendSignedTransaction.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = sendSignedTransaction.json; sourceTree = "<group>"; };
 		D197B669EF87D9D6642E12BA /* UINotificationName+Additions.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "UINotificationName+Additions.swift"; sourceTree = "<group>"; };
 		D197B6E47F5F979A8C154E9A /* ActiveNetworkView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ActiveNetworkView.swift; sourceTree = "<group>"; };
-		D197B81BF342CD022941F8CB /* apps?top=false&recent=true&limit=10.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = "apps?top=false&recent=true&limit=10.json"; sourceTree = "<group>"; };
+		D197B86A198E11D1255A9721 /* getFeaturedApps.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = getFeaturedApps.json; sourceTree = "<group>"; };
```
roberthein: Much better 😅

roberthein:
@marijnschilling I'm not sure about adding more tests to test all different error-states.

marijnschilling:
Oops I keep forgetting to check the errors 🙈  I will fix them!

And in that case i think these error states are enough, i think for the others the result that is expected is the same.

PR #451, opened 2017-08-29
Imrpove by minor fixes

yuliaveres:


Toshi/APIClients/idAPIClient.swift
```diff
@@ -41,7 +41,13 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
         }
     }()
 
-    let contactUpdateQueue = DispatchQueue(label: "token.updateContactsQueue")
+    private lazy var updateOperationQueue: OperationQueue = {
+        let queue = OperationQueue()
+        queue.maxConcurrentOperationCount = 2
```
faymontage: Where does the value "2" come from? Can we give it a name?

yuliaveres: it is enough to have 1 actually

Toshi/Avatars/AvatarManager.swift
```diff
@@ -114,25 +114,26 @@ final class AvatarManager: NSObject, CacheExpiryDefault {
     }
 
     func downloadAvatar(path: String, completion: @escaping (_ image: UIImage?) -> Void) {
+        DispatchQueue.global(qos: .userInitiated).async {
+            guard let url = URL(string: path) as URL?,
+                let teapot = self.teapot(for: url) as Teapot? else { return }
```
heumn: Shouldnt the completion block be called here? 🤔

yuliaveres: it should

PR #452, opened 2017-08-30
Recovered id api client tests

marijnschilling:
Made this branch new because there were merge conflicts in carthage files.
- [ ] Lots of tests tests tests
- [ ] Also tests for unauthorized state

- [ ] I added an error completion to the `fetchTimestamp` function to catch the errors in my test
- [ ] Removed the optional User array completion, since it's defaulted to an empty array anyway


Tests/Mocks/IDAPIClient/adminLogin.json
```diff
@@ -0,0 +1 @@
+{}
```
roberthein: Maybe we can rename this to `emptyResponse.json` so we can reuse this in the future without making duplicate files. What do you think?

Toshi/Controllers/Browse/Controllers/BrowseAllController.swift
```diff
@@ -82,7 +82,7 @@ class BrowseAllController: UITableViewController {
                     Navigator.presentModally(alertController)
                 }
 
-                self.searchResults = users ?? []
+                self.searchResults = users
```
roberthein: ✨

Tests/Mocks/IDAPIClient/adminLogin.json
marijnschilling: yes! I can just leave the mockfile name empty and then it works as well. So doing that for now, and will adjust in Teapot to make that parameter optional.

PR #458, opened 2017-08-31
IOS-297 - Add passphrase warning when trying to deposit.

roberthein:
Show a warning when going to `Deposit` when passphrase isn't verified.
Present `Passphrase backup` from `Deposit` screen.
Cleanup `PassphraseEnableController`.
Tweak dismiss handling of `PassphraseVerifyController`.

img_3865

img_3866


Toshi/Controllers/Settings/AddMoney/AddMoneyController.swift
```diff
@@ -61,6 +65,23 @@ class AddMoneyController: UIViewController {
         stackView.edges(to: scrollView)
         stackView.width(to: scrollView)
     }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        if !isAccountSecured {
```
yuliaveres: isn't that better to return if account is secure - and then keep the whole method not in `if` scope ?

Toshi/Controllers/Settings/Passphrase/PassphraseEnableController.swift
```diff
@@ -74,22 +69,35 @@ class PassphraseEnableController: UIViewController {
...
-            self.actionButton.bottomAnchor.constraint(equalTo: self.view.bottomAnchor, constant: -30)
+            actionButton.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            actionButton.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: -30)
         ])
     }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        if presentationStyle == .modal {
```
yuliaveres: 🤔  i think we can check for self.presentingViewController - if any, means presented modally, not pushed. ?
also, it would show up too much i guess. what if u pressed Verify and Cancel - again this alert. 🤔 
would u mind implementing some protocol informing when `PassphraseEnableController` is going to finish / cancel thus here u will know if there is a need to show it on appear which is going to happen? Right? 
then on each appear check this private state , just like `isAccountSecured`, which will depend on `presentingViewController != nil` and if we are cancelled from `PassphraseEnableController` ?

roberthein: I can check for `if presentingViewController != nil` but that looks less nice I think.
I think it's good that when you dismiss the message pops-up again, so that the user has to actively dismiss the warning that we gave him.
Not sure what you mean with:
> implementing some protocol

Toshi/Controllers/Settings/AddMoney/AddMoneyController.swift
roberthein: I don't know what you mean by this. Can you explain more?

yuliaveres: if isAccount secure -> return
and remove the whole implementation from if body ?

Toshi/Controllers/Settings/Passphrase/PassphraseEnableController.swift
yuliaveres: it is also less nice to instantiate controller with our own presentation style enum.

delegate of `PassphraseEnableController` `willCancel` or smth. but could u double check with someone from design if that is wanted ? my personal point of view - on each screen enter - yes, on dismiss - no.

Toshi/Controllers/Settings/AddMoney/AddMoneyController.swift
roberthein: Makes no difference at this moment, your suggestion could cause issues in the future when you want to add something to `viewDidAppear`. Let's keep it the way it is.

Toshi/Controllers/Settings/Passphrase/PassphraseEnableController.swift
yuliaveres: what about wrapping it into some:
`private var isPresentedModally: Bool {
     return presentingController != nil
  }`
 `

roberthein: done!

PR #459, opened 2017-08-31
Test rating client

marijnschilling:
What I changed:
- [ ] tests for the rating client!
- [ ] added an error to the fetchTimestamp method
- [ ] Removed the alert view related code to the FavoritesController. Since it makes way more sense to handle that in a ViewController. Otherwise i would need to add UIKit classes to the test target, and look its way less redundant 👯 

What needs to be tested:
- [ ] Maybe try to make an erroneous rating to check wether the alert view behaves as expected. I tested on simulator, but not on device.


Tests/RatingsClientTests.swift
```diff
@@ -0,0 +1,96 @@
...
+
+            context("Happy path 😎") {
+
+                it("fetches the top rated apps") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: RatingsClientTests.self), mockFileName: "")
+                    mockTeapot.overrideEndPoint("timestamp", withFileName: "timestamp")
+                    subject = RatingsClient(teapot: mockTeapot)
+
+                    waitUntil { done in
+                        subject.submit(userId: "testUseID", rating: 4, review: "") { success, message in
```
roberthein: This is generating a Swiftlint warning: ⚠️`Unused parameter "message"`

Tests/RatingsClientTests.swift
```diff
@@ -0,0 +1,96 @@
...
+
+            context("⚠ Unauthorized 🔒") {
+
+                it("fetches the top rated apps") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: RatingsClientTests.self), mockFileName: "", statusCode: .unauthorized)
+                    mockTeapot.overrideEndPoint("timestamp", withFileName: "timestamp")
+                    subject = RatingsClient(teapot: mockTeapot)
+
+                    waitUntil { done in
+                        subject.submit(userId: "testUseID", rating: 4, review: "") { success, message in
```
roberthein: This is generating a Swiftlint warning: ⚠️`Unused parameter "message"`

Tests/RatingsClientTests.swift
```diff
@@ -0,0 +1,96 @@
...
+                            expect(score.stars.four).to(equal(0))
+                            expect(score.stars.five).to(equal(0))
+                            done()
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
roberthein: I don't like to complain about whitespaces or newlines but this is generating a Swiftlint warning: ⚠️`Files should have a trailing newline`.
🤷‍♀️

Toshi/APIClients/RatingsClient.swift
```diff
@@ -162,6 +165,7 @@ class RatingsClient: NSObject {
             self.teapot.get("/v1/user/\(userId)") { result in
                 switch result {
                 case .success(let json, _):
+                    print(json)
```
roberthein: I don't think we need to print this, or do we?

PR #460, opened 2017-08-31
IOS-296 - Show confirmation alert when accepting payment request.

roberthein:
Add `approvePaymentForIndexPath(_:)` and `declinePaymentForIndexPath(_:)`.
Re-route cell delegate trough alert to ☝️these new methods.
Add formatted localization.

img_3867


Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -571,6 +571,33 @@ final class ChatController: OverlayController {
 
         showMediaPickerBlock(nil)
     }
+    
+    func approvePaymentForIndexPath(_ indexPath: IndexPath) {
```
yuliaveres: those should be fileprivate ?

roberthein: Added fileprivate.

yuliaveres: also let me know what do you think about this, please:
https://swift.org/documentation/api-design-guidelines/#naming

`func approvePayment(for indexPath: IndexPath)` then `approvePayment(for: indexPath)` vs `approvePaymentForIndexPath(indexPath)` ?

roberthein: `approvePayment(for indexPath: IndexPath)` causes a error in this case.

yuliaveres: ok, got it

PR #471, opened 2017-09-06
Fix crash for UI modification on a background thread.

roberthein:
Fix crash for UI modification on a background thread.


Toshi/Controllers/Messaging/ChatInteractor.swift
```diff
@@ -142,17 +142,22 @@ final class ChatsInteractor: NSObject {
...
                     guard let txHash = json["tx_hash"] as? String else { fatalError("Error recovering transaction hash.") }
                     guard let value = parameters["value"] as? String else { return }
 
                     let payment = SofaPayment(txHash: txHash, valueHex: value)
-                    self?.sendMessage(sofaWrapper: payment)
-
-                    completion?(true)
+                    
+                    DispatchQueue.main.async {
+                        self?.sendMessage(sofaWrapper: payment)
```
yuliaveres: maybe better to send in background ?

roberthein: What do you mean exactly?

roberthein: Yes! Ok 👌

roberthein: done!

yuliaveres:
Nice catch !

PR #472, opened 2017-09-06
[IOS-308] Use device locale or default locales map to match selected one

yuliaveres:


yuliaveres:
Due to pretty old base branch failed to fix CI test for one framework signing in the project. Reopened as new pr

PR #474, opened 2017-09-07
Send messages in main thread

yuliaveres:
We need to send messages in foreground as Signal requires it 😢


yuliaveres:
@roberthein  nope 👁 , transaction send as well 👍

PR #476, opened 2017-09-07
[IOS-311]: Ping bots on conversation start

yuliaveres:
Did reload was not called if there were no messages on chatViewModel.
now we call it and initial message to bot is being sent.
also changed network alert text in this pr. !


Toshi/Supporting Files/AppDelegate.m
```diff
@@ -120,7 +120,7 @@ - (void)configureAndPresentWindow {
 - (void)showNetworkAlertIfNeeded {
     // To drive this point really home we could show this for every launch instead.
     if (![[NSUserDefaults standardUserDefaults] boolForKey:@"DidShowMoneyAlert"]) {
-        UIAlertController *alert = [UIAlertController alertControllerWithTitle:@"Be aware!" message:@"Toshi is running on Testnet. Do not send Ethereum from Mainnet." preferredStyle:UIAlertControllerStyleAlert];
+        UIAlertController *alert = [UIAlertController alertControllerWithTitle:@"Be aware!" message:@"Toshi is not an appropriate wallet for storage of large sums of ETH." preferredStyle:UIAlertControllerStyleAlert];
```
roberthein: Can you localize these strings?

Toshi/Controllers/Messaging/ChatController.swift
```diff
@@ -231,6 +231,8 @@ final class ChatController: OverlayController {
 
         isVisible = true
 
+        viewModel.load()
```
roberthein: The viewModel is not loading here, what is loading here? Maybe give `load()` more context?

roberthein: Maybe something like `LoadFirstMessages`? What do you think?

Toshi/Supporting Files/AppDelegate.m
yuliaveres: yup

Toshi/Controllers/Messaging/ChatController.swift
yuliaveres: yup

PR #478, opened 2017-09-08
[WIP]: Suggest chat alternative IGListKit

yuliaveres:
Still in progress,
cell selection and position is not correct


yuliaveres:
We have decided to drop this idea as we do not benefit from it that much.

PR #480, opened 2017-09-11
[IOS-322]: Improve networking. Add caching

yuliaveres:
Got rid of weird request timeout intervals


yuliaveres:
Not needed as soon as those changes were made inside Teapot. No need to point to the specific commit there.

PR #481, opened 2017-09-12
Re-add files to project and test target

marijnschilling:
oops, the IDAPIClient tests weren't added to the project, and some of the mock files not to the test target. 

Now they're up and running again!

I also made the IDAPIClient sent through the error messages that tell me about a missing mock file while I was at it ✌️


Toshi/APIClients/idAPIClient.swift
```diff
@@ -277,9 +277,9 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
 
                     TokenUser.current?.update(json: json)
                     succeeded = true
-                case .failure(let json, _, _):
+                case .failure(let json, let response, let error):
```
yuliaveres: why do u need response here ?

Toshi/APIClients/idAPIClient.swift
```diff
@@ -277,9 +277,9 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
...
                     let errors = json?.dictionary?["errors"] as? [[String: Any]]
-                    errorMessage = (errors?.first?["message"] as? String) ?? ""
+                    errorMessage = (errors?.first?["message"] as? String) ?? "\(error)"
```
yuliaveres: which error do u refer here to ? it also created a warning here, maybe use some placeholder error message ?

yuliaveres: i think what u might want here - is to use 3d case parameter `error` and use its `localizedDescription` instead of `\(error)`

Toshi/APIClients/idAPIClient.swift
```diff
@@ -526,7 +526,7 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
                     succeeded = true
                 case .failure(let json, _, _):
                     let errors = json?.dictionary?["errors"] as? [[String: Any]]
-                    errorMessage = (errors?.first?["message"] as? String) ?? ""
```
yuliaveres: here exactly the same about error.

Toshi/APIClients/idAPIClient.swift
```diff
@@ -487,7 +487,7 @@ public class IDAPIClient: NSObject, CacheExpiryDefault {
...
+                    errorMessage = (errors?.first?["message"] as? String) ?? "\(error)"
```
yuliaveres: and here exactly the same about error.

Toshi/APIClients/idAPIClient.swift
marijnschilling: localizedDescription is some meaningless Teapot error thing. We need this "\(error)" to make it the associated string value of the TeapotError enum. Or is there another way to get this vlaue out?

yuliaveres: "to make it the associated string value of the TeapotError enum" ?

marijnschilling: yes 😬 

I need the associated string value of the enum here, that contains the error message

yuliaveres: so in this case i think string of \(error) is even worse, what does it provide us with ?

Can u just pass any placeholder string then.?

PR #485, opened 2017-09-13
Update carthage

yuliaveres:
Now we cache network requests responses if remote resource was not changed.

Also I changed the avatar local cache to be never stale, because now, as soon as user updates avatar, it will come with new url and there will be nothing for this key in the cache.
When avatar is being requested - cached image is returned immediately if there is any. Otherwise it will be fetched for the given path and returned.


Toshi/Avatars/AvatarManager.swift
```diff
@@ -17,10 +17,14 @@ import Foundation
 import Teapot
 import AwesomeCache
 
-final class AvatarManager: NSObject, CacheExpiryDefault {
+final class AvatarManager: NSObject {
 
     static let shared = AvatarManager()
 
+    internal lazy var cacheExpiry: CacheExpiry = {
+        return CacheExpiry.never
```
roberthein: You can just return `.never` here..

yuliaveres: yup

PR #489, opened 2017-09-14
offline alert view

marijnschilling:
Added a nice barry-bar with an alert when interwebs is not reachable. 

- I made the offline alert view with the same construction as I made the `ActiveNetworkView`: a view with a `displayable` protocol that the ViewController that wants to display it should implement. Each of the `NavigationController` subclasses that get created in the `TabBarController` implement this protocol 
- The `TabBarController` implements the `ReachabilityDelegate` methods, when the reachability status changes the `currentNavigationController` will be asked to respectively hide or show the Offline alert view.

And it looks like this: 
ofline

Things that need review:
- [ ] I added [Reachability](https://github.com/ashleymills/Reachability.swift) via Carthage. We might want to consider to write something for ourselves

https://toshiapp.atlassian.net/browse/IOS-274


heumn:
> I added it to a subclass of UINavigationController so we only have to implement it on the different viewControllers in the tabbar.

> I made the offline alert view with the same construction as i made the ActiveNetworkView: a view with a displayable protocol that the ViewController that wants to display it should implement. But im not sure if thats the right approach in this case since actually only NavigationController will implement it. Any tips on the structure are super welcome!

Did you consider composition instead of inheritance? 🤔 Looks like you are even halfway there with your protocol use 👌  Idiots like me will come barking into this subclass and say "hey, what if I just add this other general thing I use ALL over the place to this already existing and used ALL over the place UINavigationController?". Subclassing is often a slippery slope 😂 

I would maybe then propose it is better with some repetetive code in the specific UIViewControllers that wants this feature, rather than hiding it in in a shared UINavigationController subclass?

marijnschilling:
Thanks! this is exactly the kind of feedback that i was looking for! How i choose to approach this problem felt a bit random, and felt like it could be improved!

I guess i could just add a Protocol that the specific NavigationControllers conform to. Since all the different tabs in the app have a subclassed navigationController anyways. (`FavoritesNavigationController` etc)

I will definitely give it a try, I suspect you are very right about the fact that it is oke in this case that every nav controller that wants this feature has some duplicate code. It will make it more clear and safer to use!

heumn:
Glad I could be useful! Just ping me on slack if you have any questions 👊

marijnschilling:
Damn now i have conflicts in the Carthage files 🙈

marijnschilling:
Made a new branch that doesnt update carthage

PR #494, opened 2017-09-21
Use proper network as default on distribution

yuliaveres:
this fix should have been done along with https://github.com/toshiapp/toshi-ios-client/pull/491/files, but was missed.
1 - the switching on Development didnt work actually cause single Ethereum service path was used which just depended on target. so indicated pr fixed this. (thus when on Distribution - mainnet was used as default, this is fine)
2 - this just adds mainnet to networks info, so it can be used for Distribution, as currently only Ropsten and Toshi test networks are used.


Toshi/Network/NetworkSwitcher.swift
```diff
@@ -26,11 +26,13 @@ enum NetworkInfo {
 
     struct Label {
 
+        static let MainNet = Localized("Mainnet")
```
roberthein: There is no localization for this, or the others below, in the string file can you please add it? Maybe it should also be "MainNet" or "Main Net"?

yuliaveres: oh there is no, i was to add it but forgot ! adding now 👍

roberthein: 👍

PR #495, opened 2017-09-22
Offline alert

marijnschilling:
Added a nice barry-bar with an alert when interwebs is not reachable. 

- I added the alert to the bottom of the view instead, over the TabBar. Since in iOS11 it is not possible to get the right position for `under the navigationBar` from outside the viewController self. 

And it looks like this: 
offline-alert

https://toshiapp.atlassian.net/browse/IOS-274


Toshi/Views/OfflineAlertView.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import TinyConstraints
+import SweetUIKit
+
+class OfflineAlertView: UIView {
+    static let height: CGFloat = 32.0
```
roberthein: This `height` is not being used.

Toshi/Views/OfflineAlertView.swift
```diff
@@ -0,0 +1,62 @@
...
+    fileprivate let margin: CGFloat = 6.0
```
roberthein: This `let` can be `private`.

Toshi/Views/OfflineAlertView.swift
```diff
@@ -0,0 +1,62 @@
...
+        setupView()
+    }
+
+    func setupView() {
+        backgroundColor = Theme.offlineAlertBackgroundColor
+
+        addSubview(textLabel)
+        textLabel.edges(to: self, insets: UIEdgeInsets(top: margin, left: margin, bottom: -margin, right: -margin), priority: .defaultHigh)
+
+        heightConstraint = heightAnchor.constraint(equalToConstant: 0)
```
roberthein: Is there a reason why you animate the height of the view instead of the y position? Because now the label is being clipped in the animation. Looks a bit broken to me.
screen shot 2017-10-02 at 13 27 00

Toshi/Views/OfflineAlertView.swift
```diff
@@ -0,0 +1,62 @@
...
+        heightConstraint?.isActive = true
```
roberthein: If you would use TinyConstraints for this constraint you wouldn't need this line.. 🤷‍♂️

Toshi/Views/OfflineAlertView.swift
```diff
@@ -0,0 +1,62 @@
...
+        super.init(coder: aDecoder)
+
+        setupView()
+    }
+
+    func setupView() {
+        backgroundColor = Theme.offlineAlertBackgroundColor
+
+        addSubview(textLabel)
+        textLabel.edges(to: self, insets: UIEdgeInsets(top: margin, left: margin, bottom: -margin, right: -margin), priority: .defaultHigh)
```
roberthein: I guess you set `priority: .defaultHigh` because else the constraints break? I think you could also omit the top and bottom margins instead.

Toshi/Views/OfflineAlertView.swift
```diff
@@ -0,0 +1,62 @@
...
+import SweetUIKit
+
+class OfflineAlertView: UIView {
+    static let height: CGFloat = 32.0
+    fileprivate let margin: CGFloat = 6.0
+
+    var heightConstraint: NSLayoutConstraint?
+
+    fileprivate lazy var textLabel: UILabel = {
+        let textLabel = UILabel(withAutoLayout: true)
```
roberthein: You can omit `withAutoLayout: true` here.

Toshi/Views/OfflineAlertView.swift
marijnschilling: I use it in `OfflineAlertDisplaying` line 48. Is XCode saying it's not being used?

marijnschilling: Ow wait, I used the `ActiveNetworkView.height` where I should use this. Copy mistake, I fixed that in `OflineAlertDiplaying` line 48. Now it's used as it's supposed to!

Toshi/Views/OfflineAlertView.swift
marijnschilling: Buttt if i change the yPosition, it wont hide right? 

I wouldnt really know how to make it appear and dissapear with chaning the y pos?

roberthein: Would it not go behind the navigation-bar?

marijnschilling: nope.

marijnschilling: Ik kan wel eerst de tekst onzichtbaar maken :)

roberthein: Of `view.addSubview(offlineAlertView)` vervangen met `view.insertSubview(offlineAlertView, belowSubview: navigationBar)` en dan de y-pos animeren..

PR #496, opened 2017-09-22
[iOS-251] Favorites - Design changes

marijnschilling:
- [ ] iOS-252 Smaller cell height (and other design changes to the cell)
According to this document from Robert: https://toshiapp.atlassian.net/secure/attachment/10085/Artboard.png

The result in the app: 
screenshot 2017-09-22 17 44 54

- [ ] iOS-255 Adjust table insets while keyboard is shown 
tableview insets


Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -93,9 +94,34 @@ open class FavoritesController: SweetTableController {
...
+    }
+
+    func keyboardWillShow(_ notification: Notification) {
+        guard let keyboardFrame: NSValue = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? NSValue  else { return }
+        
+        let keyboardRectangle = keyboardFrame.cgRectValue
+        let keyboardHeight = keyboardRectangle.height
+
+        let tabBarControllerHeight = self.tabBarController?.tabBar.frame.height ?? 0.0
+        bottomConstraint?.constant = -(keyboardHeight - tabBarControllerHeight)
```
marijnschilling: Was wondering if I could also use the insets of the table view to get this done, but I tried it and it didn't work as expected. But if you think that would be a better solution do tell!

yuliaveres: Yeah, you should use insets. we already do this in project :)
could you please double check `Editable` protocol to which `ProfileEditViewController` does conform ? (if u need to change title of protocol as it doesn't make sense with what u need to use it for - feel free to)

marijnschilling:
Updated to make use of the `KeyboardAdjustable` protocol.

I changed the protocol  to store the basic insets on the scrollView so when the keyboard dissapears it can be reset to those insets.

Also updated the layout to make use of the safe areas

Toshi/Extensions/UIScrollView+Keyboard.swift
```diff
@@ -35,8 +41,8 @@ extension UIScrollView {
         scrollIndicatorInsets = insets
     }
 
-    func removeBottomInsets(from _: NSNotification) {
-        let insets = UIEdgeInsets(top: contentInset.top, left: contentInset.left, bottom: 0.0, right: contentInset.right)
+    func removeBottomInsets(from _: NSNotification, withBasicBottomInset basicBottomInset: CGFloat) {
```
yuliaveres: please, could we rename this method to not repeat words: i would suggest leaving `basicBottomInset ` without label, so removing `withBasicBottomInset `

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -182,6 +204,14 @@ open class FavoritesController: SweetTableController {
 
         tableView.reloadData()
         showOrHideEmptyState()
+
+        registerForKeyboardNotifications()
```
yuliaveres: i think we should not call this on each appear, (also in this case we might want to deregister in when disappear ..) but what we need here is to call in `viewDidLoad`. i think how it is now does not work when changing tab and going back, doing search - and insets are not adjusted anymore. could you check this ?

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -413,7 +438,7 @@ extension FavoritesController: Emptiable {
     }
 
     func isScrollable() -> Bool {
-        return true
+        return false
```
yuliaveres: why we changed this ?

yuliaveres: also could u double check with Robert if we want it to be scrollable ? or is that weird 🤔

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -456,8 +481,8 @@ extension FavoritesController: UITableViewDataSource {
 
 extension FavoritesController: UITableViewDelegate {
 
-    public func tableView(_: UITableView, estimatedHeightForRowAt _: IndexPath) -> CGFloat {
-        return UITableViewAutomaticDimension
+    public func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return ContactCell.height
```
yuliaveres: we dnt resize automatically ? 😞

Toshi/Protocols/Editable.swift
```diff
@@ -16,8 +16,9 @@
 import Foundation
 import UIKit
 
-protocol Editable: class {
+protocol KeyboardAdjustable: class {
```
yuliaveres: rename file as well ? ..

Toshi/Controllers/Settings/Profile/ProfileEditController.swift
```diff
@@ -17,11 +17,13 @@ import UIKit
 import SweetUIKit
 import SweetFoundation
 
-open class ProfileEditController: UIViewController, Editable, UINavigationControllerDelegate {
+open class ProfileEditController: UIViewController, KeyboardAdjustable, UINavigationControllerDelegate {
```
yuliaveres: not sure why, but it does not scroll properly to last field f.e. ... try iPhone 5 start editing location field.. it scrolls to the middle only .. maybe some calculations are but wrong ?

Toshi/Extensions/UIScrollView+Keyboard.swift
marijnschilling: hmmm but it is a method that removes the bottomInsets from the notification with keeping in mind the basic bottom insets that were already there... 

I will change it to:
`func removeKeyboardInsets(from _: NSNotification, withBasicBottomInset basicBottomInset: CGFloat) {`

Toshi/Controllers/Favorites/FavoritesController.swift
marijnschilling: I made it sccrollable but set `bounces` to false. Otherwise you can scroll this view while it doesn't exceed the views bounds, that looks less neat in my opinion.

yuliaveres: i think it is great solution

Toshi/Extensions/UIScrollView+Keyboard.swift
yuliaveres: ah i meant getting rid of `withBasicBottomInset` component . cause we duplicate phrase. 
and it is against swift naming conventions 😉 
`func removeKeyboardInsets(from _: NSNotification, basicBottomInset: CGFloat) {`

Toshi/Controllers/Settings/Profile/ProfileEditController.swift
marijnschilling: What do you mean `it scrolls in the middle only?` 

With me it works ok (or am i missing something?):
scroll

Toshi/Controllers/Favorites/FavoritesController.swift
marijnschilling: I doesn't work, that's why I changed it the cells where being compressed. I cant make it work on iOS10. iOS11 works fine! Let's drop ios10! :P

Toshi/Controllers/Settings/Profile/ProfileEditController.swift
yuliaveres: looks like middle of location field. not bottom 🤔

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -456,7 +442,7 @@ extension FavoritesController: UITableViewDataSource {
 
 extension FavoritesController: UITableViewDelegate {
 
-    public func tableView(_: UITableView, estimatedHeightForRowAt _: IndexPath) -> CGFloat {
+    public func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
```
yuliaveres: that is by default, we can remove this 👍  😉

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -17,14 +17,33 @@ import UIKit
 import SweetUIKit
 import SweetFoundation
 import SweetSwift
+import TinyConstraints
 
-public extension Array {
-    public var any: Element? {
-        return self[Int(arc4random_uniform(UInt32(self.count)))] as Element
+open class FavoritesController: SweetTableController, KeyboardAdjustable, Emptiable {
```
yuliaveres: can we extract all conformance to extensions ?

Toshi/Protocols/Emptiable.swift
```diff
@@ -18,127 +18,69 @@ import UIKit
 import SweetUIKit
 
 protocol Emptiable: class {
-    func sourceView() -> UIView
-    func isScrollable() -> Bool
```
yuliaveres: no scrollable anymore ?

marijnschilling: Yeah, I looked at it with @roberthein and seems pointless for now at leat. These emptiable screens will never exceed the view bounds.

Toshi/Controllers/Favorites/FavoritesController.swift
marijnschilling: I don't think so, you can't have properties in an extension

yuliaveres: okay 👍

Toshi/Controllers/Settings/Profile/ProfileEditController.swift
marijnschilling: True! The textField was not the height of the cell i fixed that now :V:

PR #507, opened 2017-09-27
Adjust fonts and textcolors to new design

marijnschilling:
- [ ] Made the whole app work with preferred fonts, so the textsize is dynamic. Some elements have fixed fonts:
    - views like custom alert bars, custom buttons
    - rating view
    - passphrase word blocks (should change this in other PR)
    - some extremely large text sizes (if you cant read that you shouldnt use a smartphone 😂 )
    - balance bar in chat controller (amount already doesnt fit atm)
    
Please check by enlarging text sizes and see if the app looks proper! 

- [ ] Adjust the dark and medium textcolor
- [ ] in many place adjust the font for titles and subtitles

- [ ] IOS-331 Problem where the header in Settings is cut off
- [ ] Removed the extra spacing for the searchField in BrowseController for iOS 11 (since it's part of the navigationbar there)

- [ ] Gave the whole settingsController a bit of a design upgrade to be more like Sketch. 

before:
screenshot 2017-09-27 15 08 45

✨ after✨ :
screenshot 2017-09-27 14 33 10


Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -382,6 +372,49 @@ extension SettingsController: UITableViewDelegate {
         }
     }
 
+    open func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        let sectionItem = sections[section]
```
yuliaveres: i think we can omit object extraction from an array here. 
we may just compare Int which is more efficient? 
`if section == SettingsSection.security.rawValue`

Toshi/Theme/Theme.swift
```diff
@@ -141,8 +141,8 @@ extension Theme {
 
 extension Theme {
 
-    static var sectionTitleFont: UIFont {
-        return .preferredFont(forTextStyle: .footnote)
+    @objc public static var sectionTitleFont: UIFont {
+        return .systemFont(ofSize: 13)
```
yuliaveres: we should not hard code this value. the app should support global user preferences. if text size is large on iOS, we have to adjust to it. see the version text footer: only it reacted properly:
img_6858

Toshi/Views/ChatCell.swift
```diff
@@ -126,7 +126,7 @@ class ChatCell: UITableViewCell {
 
     lazy var usernameLabel: UILabel = {
         let view = UILabel()
-        view.font = Theme.semibold(size: 16)
+        view.font = Theme.semibold(size: 17)
```
yuliaveres: same here, could you please adjust the font as it was hardcoded ... :/

yuliaveres:
we have to support user text size preferences as much as possible:

img_6860
img_6861
img_6862
img_6863
thus we need to discuss and agree on having good named font is Theme which are created in proper way, not with hardcoded size (when possible of course)

Toshi/Controllers/Browse/Views/SearchResultCell.swift
```diff
@@ -16,23 +16,22 @@
...
     }()
 
-    lazy var subLabel: UILabel = {
+    lazy var usernameLabel: UILabel = {
         let view = UILabel(withAutoLayout: true)
         view.setContentHuggingPriority(.defaultLow, for: .horizontal)
         view.setContentCompressionResistancePriority(.required, for: .horizontal)
         view.textColor = Theme.greyTextColor
-        view.font = Theme.regular(size: 13)
+        view.font = Theme.preferredFootnote()
```
yuliaveres: love it ! ❤️

Toshi/Controllers/Favorites/ContactController.swift
```diff
@@ -44,31 +44,31 @@ public class ContactController: UIViewController {
     lazy var nameLabel: UILabel = {
         let view = UILabel(withAutoLayout: true)
         view.numberOfLines = 0
-        view.font = Theme.regular(size: 24)
+        view.font = Theme.preferredTitle2()
```
yuliaveres: hmm is that the best title we might figure out ? 🤔  ...

Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -90,6 +89,10 @@ class BalanceController: UIViewController {
 
 extension BalanceController: UITableViewDelegate {
 
+    public func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return UITableViewAutomaticDimension
```
yuliaveres: that is by default right ? we can remove this method..

Toshi/Controllers/Settings/DepositMoney/DepositMoneyController.swift
```diff
@@ -35,19 +35,25 @@ class AddMoneyController: UIViewController {
     convenience init(for username: String, name _: String) {
         self.init(nibName: nil, bundle: nil)
 
-        title = "Add Money"
+        title = Localized("deposit_money_title")
```
yuliaveres: great

Toshi/Controllers/Settings/Passphrase/PassphraseEnableController.swift
```diff
@@ -18,13 +18,19 @@ import SweetUIKit
 
 class PassphraseEnableController: UIViewController {
     
-    lazy var titleLabel = TitleLabel(Localized("passphrase_enable_title"))
+    lazy var titleLabel: TextLabel = {
```
yuliaveres: can we make it private, or at least private(set) ?

Toshi/Controllers/Favorites/ContactController.swift
marijnschilling: Yeah it's what it is called :) `.title2` so I think it would be good to keep that consistent. (but in that logic `regular` should also be called `Body`)

yuliaveres: ahh ok then maybe it is best one ...

Toshi/Theme/Theme.swift
```diff
@@ -161,15 +167,47 @@ extension Theme {
...
+
+    @objc static func preferredTitle1() -> UIFont {
+        return .preferredFont(forTextStyle: .title1)
+    }
+
+    @objc static func preferredTitle2() -> UIFont {
+        return .preferredFont(forTextStyle: .title2)
+    }
+
+    @objc static func preferredTitle3() -> UIFont {
```
yuliaveres: great job @marijnschilling  on this! now we will need to use it in 99% places of the app 🙏  thx!

marijnschilling: Yes! We already do now, everywhere where it is wanted at least :) So except some things like the rating controller, but that will get a huge redesign soon, so not worht it to spent time on that. And the balance header in chat, because it wont fit.

And some other small elements, but that is fine to leave them in static heights.

The only thing we need to tackle is the login screen, but RH will tackle that :)

PR #510, opened 2017-09-28
[IOS-341, IOS-328]: Support paste on Sign in

yuliaveres:
- Show the 'paste-option' by tapping the placeholder, or tapping the ellipsis or by long pressing anywhere in the view.
- Dismiss 'paste-option' by tapping anywhere outside of the popover or start typing on the keyboard.


Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -225,6 +261,24 @@ extension SignInViewController: UICollectionViewDelegate {
...
+        userDidPastePassphrase = true
+        
+        typed = pasted.components(separatedBy: " ")
+        itemCount = typed.count
+        
+        signInView?.collectionView.reloadData()
+        signInView?.collectionView.collectionViewLayout.invalidateLayout()
+        signInView?.layoutIfNeeded()
+        
+        userDidPastePassphrase = false
```
yuliaveres: what if user pastes one word ? 
maybe should check here if count of pasted is less then required and:
- append empty string to the `typed`
- collectionView.selectItemAtIndexPath .. select proper last indexPath;

 and enable by this further input ?

tristan:
Pasting a phrase with an invalid word results in this:
screen shot 2017-10-03 at 19 01 19
no text can be entered into the empty bubble, and selecting other bubbles leaves still shows the cursor in the empty one.

On android the invalid word is shown to be a bad word and can be edited.

yuliaveres:
All issues found are fixed. It works !

Toshi/Controllers/Sign In/SignInCell.swift
```diff
@@ -27,10 +27,16 @@ final class SignInCell: UICollectionViewCell {
     override var isSelected: Bool {
         didSet {
             guard isSelected != oldValue else { return }
-            caretView.alpha = isSelected ? 1 : 0
-            backgroundImageView.isHidden = isSelected
+            self.isActive = isSelected
+        }
+    }
+
+    var isActive: Bool = false {
```
marijnschilling: What is the reason there is this property `isActive` besides `isSelected`? Couldn't all the things that happen in `isActive` just be happening in `isSelected` since they always have the same value: `self.isActive = isSelected`?

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -79,19 +93,20 @@ final class SignInViewController: UIViewController {
...
-        guard let cereal = Cereal(words: passphrase) else {
+        guard Cereal.areWordsValid(passphrase) else {
             let alertController = UIAlertController.dismissableAlert(title: Localized("passphrase_signin_error_title"), message: Localized("passphrase_signin_error_verification"))
             present(alertController, animated: true)
 
             return
         }
 
-        Cereal.shared = cereal
+        let validCereal = Cereal(words: passphrase)!
```
marijnschilling: oh no force unwrapped, couldnt we put this together in the guard in line 96?
 I see you do a check there to make sure the words are valid before you create the Cereal with those words. But what is wrong with how it was before? Just guard the optional initializerform Cereal with those words? Wouldnt the result be the same? if the words aint valid it will come into the guard block.

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -196,18 +211,40 @@ final class SignInViewController: UIViewController {
...
 
     func collectionView(_ collectionView: UICollectionView, shouldSelectItemAt indexPath: IndexPath) -> Bool {
         guard let activeIndexPath = activeIndexPath, indexPath == activeIndexPath else { return true }
         
-        shouldDeselectWord = true
-        acceptItem(at: indexPath, completion: { [weak self] _ in
-            guard let indexPath = self?.activeIndexPath else { return }
-            self?.cleanUp(after: indexPath)
-        })
+        if let text = typed.element(at: indexPath.item), text.isEmpty {
```
marijnschilling: aaah so this checks wether the word cell you selected contains any text, and if it doesn't it will show the tiny `Paste` controller. 

I don't really get it though what happens when there is text in the cell?

And whenI select a cell containing text it does pop up the `Paste` controller.

Maybe this if else statement could be made more clear with descriptive bool states.

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -225,6 +262,34 @@ extension SignInViewController: UICollectionViewDelegate {
...
 }
 
+extension SignInViewController: TextEditingOptionsViewControllerDelegate {
+    
+    func pasteSelected(from textEditingOptionsViewController: TextEditingOptionsViewController) {
+        guard let pasted = UIPasteboard.general.string else { return }
+        
+        userDidPastePassphrase = true
+        
+        typed = pasted.components(separatedBy: " ")
```
marijnschilling: I think this will be clearer when it's named `typedWords`

Toshi/Controllers/Sign In/SignInCell.swift
yuliaveres: the flow differs, selection is changed only if previous value is different, while through `isActive` it is possible to force cell adjustment from controller when needed.

Toshi/Controllers/Sign In/SignInViewController.swift
marijnschilling: Or maybe even `pastedWords` actually

Toshi/Controllers/Sign In/SignInViewController.swift
yuliaveres: we should not set shared cereal not knowing if the sig nin flow succeeded;
regarding guard - will do it

roberthein: @yuliaveres Could you explain?

yuliaveres: we should not set shared cereal until sign in succeeded. we set it before and even if words were valid, the user was not existing, but cereal was set up for this passphrase.
Then, when new user created, this passphrase was picked, as Cereal was set up for it, what is not correct

yuliaveres: @roberthein

Toshi/Controllers/Sign In/SignInViewController.swift
yuliaveres: they are not pasted always right ?

yuliaveres: 🤔

roberthein: They're also not words always, maybe `enteredStrings`?

Toshi/Controllers/Sign In/SignInViewController.swift
roberthein: Ok, but we could still remove the force unwrap right @yuliaveres?

yuliaveres: i did it @roberthein ,

yuliaveres: i replied to it "regarding guard - will do it"

roberthein: ohw.. I missed that comment.. XD

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -96,15 +96,13 @@ final class SignInViewController: UIViewController {
 
     private func signInWithPasshphrase(_ passphrase: [String]) {
         
-        guard Cereal.areWordsValid(passphrase) else {
+        guard Cereal.areWordsValid(passphrase), let validCereal = Cereal(words: passphrase) else {
```
marijnschilling: 👯 💂

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -43,6 +44,9 @@ final class SignInViewController: UIViewController {
         loadPasswords { [weak self] in
             self?.passwords = $0
             self?.signInView?.collectionView.selectItem(at: IndexPath(item: 0, section: 0), animated: false, scrollPosition: .top)
+            if let maxCharactersCount = $0.max(by: { $1.count > $0.count })?.count {
```
marijnschilling: Could you split this up in two things? I find it hard to understand this way.

```
if let something = $0.max(by: { $1.count > $0.count })?, let maxCharactersCount = something.count {
``` 

*where something will explain what id does 😃

PR #520, opened 2017-10-02
[IOS-335]: Fix payment response being not saved to db

yuliaveres:


Pods/SignalServiceKit/src/Messages/Interactions/TSInteraction.m
```diff
@@ -121,6 +121,22 @@ - (NSString *)description {
     return @"Interaction description";
 }
 
+- (NSString *)paymentStateText
+{
+    switch (self.paymentState) {
+        case TSPaymentStateFailed:
+            return NSLocalizedString(@"Failed", nil);
```
roberthein: Please add to localization and use descriptive keys here.

PR #521, opened 2017-10-02
[IOS-329] Multisession

yuliaveres:
**Needs QA**
the QA flow is basically related to user session:
 
1.  create new user
2. check if registered for chat properly
3. check if registered for PNs
4. check if databases are created. (for user / for chat)
5. check if cereal id address/ payment address are created properly.
6. try sending messages
7. try adding favorites
9. store passphrase
8. logout
9. create new user - the same flow.
10. log in with previous user - same flow + check for database files backups
etc...

Check logs in xcode to help QA the flow 🤞 

signin_


petejkim:
Is this still an issue?

yuliaveres:
the issue is still present, yes.

designatednerd:
@petejkim This is one of the issues I mentioned when discussing some of the dependencies around SSK which were causing us problems.

PR #522, opened 2017-10-03
Update badge number only for valid current user session

yuliaveres:
We need to manipulate with unread messages badge only if there is any session. (when user exists and the chat db was opened successfully)


yuliaveres:
that sounds great, will do @marijnschilling

PR #530, opened 2017-10-05
Improve app loading behaviour and splash screen.

roberthein:
Improve app loading behaviour and splash screen.


Toshi/Resources/DistributionDev.plist
```diff
@@ -30,7 +30,7 @@
 		</dict>
 	</array>
 	<key>CFBundleVersion</key>
-	<string>87</string>
+	<string>88</string>
```
yuliaveres: is that related to the pr ? ..

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -97,6 +103,8 @@ - (void)configureAndPresentWindow {
     self.window.rootViewController = tabBarController;
 
     [self.window makeKeyAndVisible];
+    
+    self.window.backgroundColor = UIColor.redColor;
```
yuliaveres: why red ?

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -319,38 +327,15 @@ - (void)applicationDidBecomeActive:(UIApplication *)application
...
-    if (shouldProceedToDBSetup) {
-
-        if ([self __tryToOpenDB]) {
-            [self configureForCurrentSession];
-        } else {
-
-            // There might be a case when filesystem state is weird and it doesn't return true results, saying file is not present even if it is.
-            // to determine this we might check keychain for database password being there
-            // in this case we want to wait a bit and try to open file again
-            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(3.0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
yuliaveres: what is the reason this part is removed ? ...

yuliaveres: it was added recently to handle rare file system and keychain race condition. please, add it back ..

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -319,38 +327,15 @@ - (void)applicationDidBecomeActive:(UIApplication *)application
...
-    self.hasBeenActivated = YES;
-
     [[TSAccountManager sharedInstance] ifRegistered:YES runAsync:^{
         // We're double checking that the app is active, to be sure since we
         // can't verify in production env due to code
         // signing.
         [TSSocketManager requestSocketOpen];
+        
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [SignalNotificationManager updateUnreadMessagesNumber];
```
yuliaveres: this should come only after DB is setup and there is a user session. only in this case.
look carefully into previous flow please 😉

Toshi/Resources/DistributionDev.plist
roberthein: It is not.. 😞

Toshi/Supporting Files/AppDelegate.m
roberthein: You don't like it? Ok I remove it. ✌️

Toshi/Supporting Files/AppDelegate.m
roberthein: I replaced it with an NSException. Because I have never seen anything like this. What was the idea here, letting the user wait at a black screen for 3 seconds and then try again?? You want to add it back??

Toshi/Supporting Files/AppDelegate.m
roberthein: I did, before I ran into a case that it was calling `updateUnreadMessagesNumber` when there was no user session and it crashed. But if you think this is not the right place to do this please help me with finding the right place, because I'm new to most of this.. ✌️

Toshi/Supporting Files/AppDelegate.m
yuliaveres: that is rare but might happen, when file system state is weird and it gives wrong results regarding particular files.. thus retrying bit later gives correct one. yes, the black screen is better than not allowing user to launch his/her session. we need it back

Toshi/Supporting Files/AppDelegate.m
yuliaveres: DM slack 🚶

PR #532, opened 2017-10-09
Fix open-url button layout.

roberthein:
Button was being shown behind the navigation bar.


Toshi/Controllers/Browse/Controllers/BrowseController.swift
```diff
@@ -233,20 +231,21 @@ class BrowseController: SearchableCollectionController {
 
     fileprivate func showOpenURLButton() {
         openURLButton.isHidden = false
-        openURLButtonTopAnchor.constant = searchBar.frame.minY - 20 + 64
-        UIView.animate(withDuration: 0.25) {
-            self.collectionView.layoutIfNeeded()
-        }
+        openURLButtonTopAnchor.constant = (navigationController?.navigationBar.bounds.height ?? 0) + searchController.searchBar.frame.height + UIApplication.shared.statusBarFrame.height
```
yuliaveres: `navigationController?.navigationBar.bounds.height` probably is not what we want on iOS10

roberthein: Fixed ✌️

PR #535, opened 2017-10-10
[IOS-347]: Cache Browse page objects

yuliaveres:
The Browse objects are stored in cache now, and updated each time on fetch. Cached data is display prior to fetched.


Toshi/APIClients/AppsAPIClient.swift
```diff
@@ -32,55 +40,78 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
...
+    private lazy var cache: Cache<TokenUsersCacheData> = {
+        do {
+            return try Cache<TokenUsersCacheData>(name: "appsCache")
+        } catch {
+            fatalError("Couldn't instantiate the apps cache")
+        }
+    }()
+
+    func getTopRatedApps(limit: Int = 10, cachedCompletion: @escaping TokenUserResults, fetchedCompletion: @escaping TokenUserResults) {
+        if let data = self.cache.object(forKey: topRatedAppsCachedDataKey) as TokenUsersCacheData?, let ratedUsers = data.objects as [TokenUser]? {
```
roberthein: too much explicit self's in this file

Toshi/APIClients/idAPIClient.swift
```diff
@@ -385,60 +399,74 @@ import Teapot
         }
     }
 
-    public func getTopRatedPublicUsers(limit: Int = 10, completion: @escaping (_ apps: [TokenUser], _ error: Error?) -> Void) {
+    public func getTopRatedPublicUsers(limit: Int = 10, cachedCompletion: @escaping TokenUserResults, fetchedCompletion: @escaping TokenUserResults) {
 
-        self.teapot.get("/v1/search/user?public=true&top=true&recent=false&limit=\(limit)") { (result: NetworkResult) in
+        if let data = self.cache.object(forKey: topRatedUsersCachedDataKey) as TokenUsersCacheData?, let ratedUsers = data.objects as [TokenUser]? {
```
roberthein: too much explicit self's in this file

Toshi/APIClients/AppsAPIClient.swift
```diff
@@ -32,55 +40,78 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
...
-        self.teapot.get("/v1/search/apps?top=true&recent=false&limit=\(limit)") { (result: NetworkResult) in
+    private lazy var cache: Cache<TokenUsersCacheData> = {
+        do {
+            return try Cache<TokenUsersCacheData>(name: "appsCache")
+        } catch {
+            fatalError("Couldn't instantiate the apps cache")
+        }
+    }()
+
+    func getTopRatedApps(limit: Int = 10, cachedCompletion: @escaping TokenUserResults, fetchedCompletion: @escaping TokenUserResults) {
```
roberthein: I don't understand why there are 2 completion blocks that do exactly the same. 
I think you can replace
```
cachedCompletion: @escaping TokenUserResults, fetchedCompletion: @escaping TokenUserResults
```
with:
```
completion: @escaping TokenUserResults
```
everywhere.

yuliaveres: i need different completions for different results - cached first - then fetched. cached are always instant, after fetch is complete (or if it is complete at all) - `fetchedCompletion` is called and data is refreshed in a cell .

roberthein: But you can just call the same `completion` block twice.

roberthein: there is no difference between the blocks right?

yuliaveres: 🤔  , no there is no, but i am wondering if i can. 
ok i will change to  one `completion`

roberthein: 👌

PR #537, opened 2017-10-12
[IOS-350]: Replace actionsheet with tableview for chat menus.

roberthein:
img_4294


Toshi/Controllers/Chat/ChatMenuTableViewCell.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+import UIKit
+import TinyConstraints
+
+class ChatMenuTableViewCell: UITableViewCell {
```
yuliaveres: please `final`

Toshi/Controllers/Chat/ChatMenuTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        tableView.separatorStyle = .none
+        tableView.register(ChatMenuTableViewCell.self, forCellReuseIdentifier: ChatMenuTableViewCell.reuseIdentifier)
+        tableView.reloadData()
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        if let superview = view.superview?.superview {
+            superview.alpha = 0
```
yuliaveres: why we need it ?

Toshi/Controllers/Chat/ChatMenuTableViewCell.swift
roberthein: ✅

Toshi/Controllers/Chat/ChatMenuTableViewController.swift
roberthein: For a nice and smooth in-animation of the pop-over.

PR #541, opened 2017-10-16
[IOS-351]: Enhanced logging

yuliaveres:
Add custom events to Crashlytics


Toshi/APIClients/CrashlyticsClient.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+@objc final class CrashlyticsLogger: NSObject {
+
+    @objc static func log(_ string: String, attributes: [String: Any]? = nil) {
+        CLSLogv("%@", getVaList([string]))
+
+        var resultAttributes: [String: Any] = [:]
+
+        if let userID = Cereal.shared.address as String? {
```
roberthein: This casting `as String?` makes something that isn't an Optional into an Optional so you have to unwrap. But if you don't cast it you also don't have to unwrap it. 🤷‍♂️

Toshi/APIClients/CrashlyticsClient.swift
```diff
@@ -0,0 +1,50 @@
...
+            resultAttributes["user_id"] = userID
+        }
+
+        if let passedAttributes = attributes as [String: Any]? {
```
roberthein: This casting `as [String: Any]?` does not do anything and can be removed.

Toshi/APIClients/CrashlyticsClient.swift
```diff
@@ -0,0 +1,50 @@
...
+            for (key, value) in passedAttributes {
+                resultAttributes[key] = value
+            }
+        }
+
+        Answers.logCustomEvent(withName: string,
+                               customAttributes: resultAttributes)
```
roberthein: Misplaced newline?

Toshi/APIClients/CrashlyticsClient.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
```
roberthein: You can also simplify this function and write it like this:
```
@objc static func log(_ string: String, attributes: [String: Any]? = nil) {
    CLSLogv("%@", getVaList([string]))

    var resultAttributes: [String: Any] = ["user_id": Cereal.shared.address]
    attributes?.forEach { key, value in resultAttributes[key] = value }

    Answers.logCustomEvent(withName: string, customAttributes: resultAttributes)
}
```
What do you think?

Toshi/Controllers/Favorites/ContactController.swift
```diff
@@ -568,6 +570,7 @@ extension ContactController: PaymentControllerDelegate {
                     strongSelf.messageSender?.send(outgoingMessage, success: {
                         print("message sent")
                     }, failure: { error in
+                        CrashlyticsLogger.log("Can not send message", attributes: ["Error": error.localizedDescription])
```
roberthein: In the attributes I see `error` and `Error` and 7 times `occured`, can you maybe make an `enum` for this? 🙏

roberthein: Or you can do something like this:
```
typealias LogAttributeType = String

extension LogAttributeType {
    
    static var error: String {
        return "error"
    }
    
    static var occured: String {
        return "occured"
    }
}
```
Then in the `log` function replace all `[String: Any]` with `[LogAttributeType: Any]` and then you can call `log` like this:
```
CrashlyticsLogger.log("No current user during session", attributes: [.occured: "Profile edit item"])
```
What do you think?

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: yup

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: wdym `makes something that isn't an Optional into an Optional` ? address is optional, depends on current session ..

roberthein: `Cereal.shared.address` is not Optional.

roberthein: Or am I missing something?

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: here agree :)

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: i dnt want to crash app while trying to catch unexpected event :D
so if that happens f.e. during session start and Cereal is not setup yet, it does not have any address, i can't pass it into dictionary like it is suggested 🤷‍♂️ 

great suggestion regarding attributes though 👍

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: oh, i meant NOT OPTIONAL sorry. so i need this cast here

Toshi/APIClients/CrashlyticsClient.swift
roberthein: This `Cereal.shared.address as String?` does not save you from a crash.

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: sorry, on master it is not optional right ! i am living in the context of multisession now :(
but we are targeting this. let support it here from now on ? or should i be back here after multisession is merged and change it back to how it is now ?

yuliaveres: as soon as multisession is there, all cereal data is optional and depends on user session

Toshi/APIClients/CrashlyticsClient.swift
yuliaveres: yes, i mean cereal being setup for user only . so i mean address being optional (very soon)

Toshi/APIClients/CrashlyticsClient.swift
roberthein: This cast is wrong even if `Cereal.shared.address` was optional.

roberthein: When something is optional, you unwrap it. No need for casting.

yuliaveres: can you explain ?

yuliaveres: got it

yuliaveres: now it is impossible to unwrap though . ok, will be back here later

roberthein: Yeah so, when `Cereal.shared.address` will be optional in the future, you should do:
```
if let userID = Cereal.shared.address {
    resultAttributes["user_id"] = userID
}
```
No casting.
Now it's not optional so you can use `Cereal.shared.address ` without casting or unwrapping.

Toshi/APIClients/CrashlyticsClient.swift
roberthein: Even if it's optional, you still don't have to cast it.

Toshi/Controllers/Favorites/ContactController.swift
yuliaveres: i like the idea

yuliaveres: but i am wondering if `LogAttributeType ` is smth we want. lets assume we are gping to use keys strings in other places of the app. and also would like to name keys in proper way. creating another typealias for `String` is not a good idea i guess. maybe we should create one for global use.. like "Key", "KeyTitle" .. ? any ideas ?

roberthein: Yeah a `typealias` on `String`  may not be the best idea. So what do you propose for the `KeyTitle`? A `struct`?

yuliaveres: what i meant is just other name for typealias to have it more generic 🙈  😄

roberthein: Aah yes, in that case a `typealias` makes more sense yes. 👍

yuliaveres: using struct here is not really possible for what we need. it should be `hashable` (cause used as key in dictionary) + it should operate on instance with properties. that is not need for string case .. ?

PR #542, opened 2017-10-16
Design improvements!

roberthein:
Improvements to chat, fonts and other design related tweaks.


Toshi.xcodeproj/project.pbxproj
```diff
@@ -385,6 +385,11 @@
 		A9202E701ED850270036E9B7 /* ImagesViewControllerTransition.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9202E4E1ED84F6A0036E9B7 /* ImagesViewControllerTransition.swift */; };
 		A9202E721ED854010036E9B7 /* ChatInputTextPanel.swift in Sources */ = {isa = PBXBuildFile; fileRef = A9202E711ED854010036E9B7 /* ChatInputTextPanel.swift */; };
 		A9207AF71F6A8088003B79BF /* passwords-library.txt in Resources */ = {isa = PBXBuildFile; fileRef = A9207AF61F6A8088003B79BF /* passwords-library.txt */; };
+		A93432CA1F94E48500A52F11 /* SF-Pro-Display-Regular.otf in Resources */ = {isa = PBXBuildFile; fileRef = A93432C51F94E48400A52F11 /* SF-Pro-Display-Regular.otf */; };
```
yuliaveres: please, double check if those used now do replace already existing ! i hope yes, in that case remove not used once. fonts are very impactful for app binary size

Toshi/Controllers/Chat/ChatButtonsCollectionViewCell.swift
```diff
@@ -26,13 +37,7 @@ final class ChatButtonsViewCell: UICollectionViewCell {
...
     }
     
-    private lazy var titleLabel: UILabel = {
-        let view = UILabel()
-        view.font = Theme.regular(size: 17)
-        view.textColor = Theme.tintColor
-        
-        return view
-    }()
+    private lazy var titleLabel = UILabel()
```
yuliaveres: it might be lazy but it also might be let, what do we want in that case , wdyt ?

Toshi/Resources/Development.plist
```diff
@@ -57,6 +57,11 @@
 		<string>SF-UI-Text-Medium.otf</string>
 		<string>SF-UI-Text-Regular.otf</string>
 		<string>SF-UI-Text-Semibold.otf</string>
+		<string>SF-Pro-Display-Bold.otf</string>
```
yuliaveres: what about other targets ?

Toshi.xcodeproj/project.pbxproj
```diff
@@ -385,6 +385,11 @@
...
+		A93432CB1F94E48500A52F11 /* SF-Pro-Display-Semibold.otf in Resources */ = {isa = PBXBuildFile; fileRef = A93432C61F94E48500A52F11 /* SF-Pro-Display-Semibold.otf */; };
+		A93432CC1F94E48500A52F11 /* SF-Pro-Display-Light.otf in Resources */ = {isa = PBXBuildFile; fileRef = A93432C71F94E48500A52F11 /* SF-Pro-Display-Light.otf */; };
+		A93432CD1F94E48500A52F11 /* SF-Pro-Display-Bold.otf in Resources */ = {isa = PBXBuildFile; fileRef = A93432C81F94E48500A52F11 /* SF-Pro-Display-Bold.otf */; };
```
yuliaveres: fonts should be included in all targets

Toshi.xcodeproj/project.pbxproj
roberthein: I removed all the fonts we are not using

Toshi/Resources/Development.plist
roberthein: yes! added! 🙏

Toshi.xcodeproj/project.pbxproj
roberthein: 🙏

Toshi/Theme/Theme.swift
```diff
@@ -150,64 +156,87 @@ extension Theme {
...
-    @objc static func preferredTitle1() -> UIFont {
-        return .preferredFont(forTextStyle: .title1)
+    static func preferredTitle3(range: ClosedRange<CGFloat> = 20...30) -> UIFont {
+        return dynamicType(for: regular(size: 16), withStyle: .title3, inSizeRange: range)
     }
-
-    @objc static func preferredTitle2() -> UIFont {
-        return .preferredFont(forTextStyle: .title2)
+    
+    static func preferredRegular(range: ClosedRange<CGFloat> = 17...30) -> UIFont {
```
yuliaveres: curious if we ever use the bottom of the range ?

Toshi/MessagesUI/Objects/Extensions.swift
```diff
@@ -28,6 +28,13 @@ extension NSAttributedString {
     }
 }
 
+extension CGFloat {
+    
+    func clamp(to range: ClosedRange<CGFloat>) -> CGFloat {
+        return Swift.min(range.upperBound, Swift.max(range.lowerBound, self))
```
yuliaveres: ok got it

Toshi/Theme/Theme.swift
```diff
@@ -150,64 +156,87 @@ extension Theme {
...
+        if #available(iOS 11.0, *) {
+            let metrics = UIFontMetrics(forTextStyle: style)
+            font = metrics.scaledFont(for: preferredFont, maximumPointSize: range.upperBound)
+        } else {
+            font = .preferredFont(forTextStyle: style)
+        }
+        
+        let augmentedFontSize = font.pointSize.clamp(to: range)
+        
+        return font.withSize(augmentedFontSize)
```
yuliaveres: 👍

Toshi/Theme/Theme.swift
```diff
@@ -150,64 +156,87 @@ extension Theme {
...
-        return .preferredFont(forTextStyle: .body)
+    
+    static func preferredRegularSmall(range: ClosedRange<CGFloat> = 17...30) -> UIFont {
+        return dynamicType(for: light(size: 17), withStyle: .subheadline, inSizeRange: range)
     }
-
-    @objc static func preferredRegularMedium() -> UIFont {
-        return .preferredFont(forTextStyle: .callout)
+    
+    static func preferredSemibold(range: ClosedRange<CGFloat> = 17...30) -> UIFont {
```
yuliaveres: i would not store all ranges in `func` parameter .. could u use some private readonly (`lazy` maybe ) variables here for all types of ranges ? and leave parameters empty ?

Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -7280,22 +7288,30 @@
...
+				CLANG_WARN_COMMA = YES;
 				CLANG_WARN_CONSTANT_CONVERSION = YES;
 				CLANG_WARN_DIRECT_OBJC_ISA_USAGE = YES;
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_EMPTY_BODY = YES;
 				CLANG_WARN_ENUM_CONVERSION = YES;
 				CLANG_WARN_INFINITE_RECURSION = YES;
 				CLANG_WARN_INT_CONVERSION = YES;
+				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES;
+				CLANG_WARN_OBJC_LITERAL_CONVERSION = YES;
```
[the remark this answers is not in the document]
roberthein: These came with updating the project file I guess, should it not be there you think?

Toshi/Theme/Theme.swift
roberthein: Yes we do, it prevents the fonts from scaling down when we don't want them to.

Toshi/MessagesUI/Objects/Extensions.swift
roberthein: hehe

Toshi/Theme/Theme.swift
roberthein: I agree with you, but I plan to have a app review session with Robert when he's back where we double check all the fonts being used in the app now and tweak the implementation where necessary.. 
So is it ok with you if we leave this in the function for now?

Toshi/Controllers/Chat/ChatButtonsCollectionViewCell.swift
roberthein: I would still like to see this as a lazy var to keep everything the same.

PR #554, opened 2017-10-24
Store fiatValueString in content json of payment and request

marijnschilling:
- [ ] I broke up `SofaWrapper` in separate files
- [ ] To the init methods of `Payment` and `PaymentRequest` i added a check that adds `fiatValueString` to the `content` if it's not already added.
- [ ] Added all the methods that take care of this as static methods to the `SofaWrapper`
- [ ] I created a String extension that takes care of adding and removing the Sofa Signature. And added tests for this.
- [ ] I removed the distinction between handling hex starting with `0x` and not because the `public convenience init(hexadecimalString hexString: String) ` in `NSDecimalNumber+Additions` handles that. But do double check if i'm not making a thought mistake

(- [ ] Also wondering if we should do the same for the `ethereumValueString` since that is calculated as well. But it is calculated with a static value, so I guess not)

https://toshiapp.atlassian.net/browse/IOS-332


Toshi/Models/SofaWrapper.swift
```diff
@@ -297,19 +297,32 @@ open class SofaPaymentRequest: SofaWrapper {
...
+        return NSDecimalNumber(hexadecimalString: hexValue)
     }
 
     public var destinationAddress: String? {
         return (json["destinationAddress"] as? String)
     }
+
+    public convenience init(valueInWei: NSDecimalNumber) {
+
+        let request: [String: Any] = [
```
yuliaveres: what about the case we do not create the request locally and send it, but just receive it from backend?
we receive payload without any `fiatValueString` thus payment request stays with `""` value for it.

yuliaveres: then we should use this initializer everywhere for payment requests ...

yuliaveres: or override `init(content)..` , retrieve value, convert it and store `fiatValueString`, only after call `super..`

yuliaveres: see `SofaWrapper` `94` where PaymentRequest is still created with just `content`

Toshi/Models/SofaWrapper.swift
```diff
@@ -297,19 +297,32 @@ open class SofaPaymentRequest: SofaWrapper {
         return ""
     }
 
+    public var fiatValueString: String {
+        guard let string = self.json["fiatValueString"] as? String else { return "" }
```
yuliaveres: after you fix initialization, this should never be a case when we have empty value, only if broken data received .. :/

Toshi/Models/SofaWrapper.swift
marijnschilling: The problem is that it calls that everytime it initialises a `PaymentRequest` That would mean it would update the value to the current exchange rate. 

I think I'll need to look into it more to see how to handle Requests and payments that we receive.

Toshi/Extensions/String+SofaHelpers.swift
```diff
@@ -0,0 +1,29 @@
...
+
+extension String {
+
+    func removeSofaIdentifier(for sofaType: SofaType) -> String {
+        return replacingOccurrences(of: sofaType.rawValue, with: "")
+    }
+
+    func addSofaIdentifier(for sofaType: SofaType) -> String {
+        var sofaString = sofaType.rawValue as String
+        sofaString.append(self)
```
yuliaveres: hmm not sure i get why we need this .... could u explain this part please ?

marijnschilling: I think the test for this will give you some insight! 

We get the sofa string from the server with a sofa identifier in the beginning, we need to take that off before we can turn that into a json dict. And the other way around: we need to add that when we want to safe a sofa json dict in a sofa message content String.

Tests/String_SofaHelpersTests.swift
```diff
@@ -0,0 +1,146 @@
...
+                    let newString = sofaString.addSofaIdentifier(for: SofaType.none)
+
+                    expect(newString).to(equal("{content: content}"))
+                }
+            }
+
+            context("message") {
+
+                it("removes message sofa identifier") {
+                    let sofaString = "SOFA::Message:{content: content}"
```
marijnschilling: Here it will go form this String: `SOFA::Message:{content: content}`

Tests/String_SofaHelpersTests.swift
```diff
@@ -0,0 +1,146 @@
...
+                    let newString = sofaString.removeSofaIdentifier(for: SofaType.message)
+
+                    expect(newString).to(equal("{content: content}"))
```
marijnschilling: To this `{content: content}` 

And we can turn that into json dict

Toshi/Models/SofaWrapper.swift
```diff
@@ -1,361 +0,0 @@
...
-
-open class SofaWrapper: SofaWrapperProtocol {
-    public var type: SofaType {
-        return .none
-    }
-
-    open var content: String = ""
-
-    open var json: [String: Any] {
-        let sofaBody = content.replacingOccurrences(of: type.rawValue, with: "")
```
marijnschilling: We used to remove the Sofa string like this:

Toshi/Models/SofaWrapper.swift
```diff
@@ -1,361 +0,0 @@
...
-
-    public init(content: String) {
-        self.content = content
-    }
-
-    public init(content json: [String: Any]) {
-        guard let data = try? JSONSerialization.data(withJSONObject: json, options: []) else { fatalError() }
-        guard let jsonString = String(data: data, encoding: .utf8) else { fatalError() }
-
-        content = type.rawValue + jsonString
```
marijnschilling: And we used to add the SofaString like this:

Toshi/Extensions/String+SofaHelpers.swift
```diff
@@ -0,0 +1,29 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import Foundation
+
+extension String {
+
+    func removeSofaIdentifier(for sofaType: SofaType) -> String {
```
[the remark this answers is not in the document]
yuliaveres: so basically i mean i would do this inside `SofaWrapper` which operates on its own type (and  its `rawValue`)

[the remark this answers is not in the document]
yuliaveres: wdyt ?

Toshi/Models/SofaWrapper.swift
yuliaveres: so yeaah, i think we dnt need an extenstion for this .. :/

Toshi/Extensions/String+SofaHelpers.swift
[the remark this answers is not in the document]
marijnschilling: I also need it before initialising the `SofaWrapper` so i'd have to make it a `static` function that case.

Toshi/Models/SofaWrappers/SofaInitialRequest.swift
```diff
@@ -0,0 +1,36 @@
...
+/// App receives that with list of values it needs to continue.
+/// Answered with a SOFA::Init message (SofaMetadataResponse).
+open class SofaInitialRequest: SofaWrapper {
+    public override var type: SofaType {
+        return .initialRequest
+    }
+
+    open lazy var values: [String] = {
+        guard self.content.hasPrefix(self.type.rawValue) else {
+            fatalError("Creating SofaMessage with invalid type!")
```
yuliaveres: "Creating" ? how do we know in this accessor we are creating sofa message ? .. 🤔

Toshi/Models/SofaWrappers/SofaInitialRequest.swift
```diff
@@ -0,0 +1,36 @@
...
+
+/// App receives that with list of values it needs to continue.
+/// Answered with a SOFA::Init message (SofaMetadataResponse).
+open class SofaInitialRequest: SofaWrapper {
+    public override var type: SofaType {
+        return .initialRequest
+    }
+
+    open lazy var values: [String] = {
+        guard self.content.hasPrefix(self.type.rawValue) else {
```
yuliaveres: also curious why do we need this guard here ?

Toshi/Models/SofaWrappers/SofaInitialResponse.swift
```diff
@@ -0,0 +1,38 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+// Each key is a value that came from the init request as stated above.
+// Example: SofaInitialResponse(contant: ["paymentAddress": "0xa2a0134f1df987bc388dbcb635dfeed4ce497e2a", "language": "en"])
+open class SofaInitialResponse: SofaWrapper {
```
yuliaveres: lets make all those subclasses `final` ? 🙏

Toshi/Extensions/String+SofaHelpers.swift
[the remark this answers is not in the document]
yuliaveres: yeah, then u will have to pass `SofaType` as param right?

Toshi/Models/SofaWrappers/SofaMessage.swift
```diff
@@ -0,0 +1,112 @@
...
+        return .message
+    }
+
+    var showKeyboard: Bool? {
+        return json["showKeyboard"] as? Bool
+    }
+
+    open lazy var body: String = {
+        guard self.content.hasPrefix(self.type.rawValue) else {
+            fatalError("Creating SofaMessage with invalid type!")
```
yuliaveres: same here, looks like some weird quard ? ........ 🤔

Toshi/Models/SofaWrappers/SofaMessage.swift
```diff
@@ -0,0 +1,112 @@
...
+        if let messageText = self.json["body"] as? String {
+            return messageText
+        }
+
+        return ""
+    }()
+
+    open lazy var buttons: [SofaMessage.Button] = {
+        guard self.content.hasPrefix(self.type.rawValue) else {
+            fatalError("Creating SofaMessage with invalid type!")
```
yuliaveres: and here :/ .. hmm 🤔

Toshi/Models/SofaWrappers/SofaPayment.swift
```diff
@@ -0,0 +1,79 @@
...
+        case error
+    }
+
+    public var status: Status {
+        guard let status = self.json["status"] as? String else { return .unconfirmed }
+        return Status(rawValue: status) ?? .unconfirmed
+    }
+
+    public var recipientAddress: String? {
+        guard let address = self.json["toAddress"] as? String else { return nil }
```
yuliaveres: hmm could we just return `self.json["toAddress"] as? String` ?

Toshi/Models/SofaWrappers/SofaPayment.swift
```diff
@@ -0,0 +1,79 @@
...
+        return address
+    }
+
+    public var senderAddress: String? {
+        guard let address = self.json["fromAddress"] as? String else { return nil }
```
yuliaveres: same here ? .. ^^^

Toshi/Extensions/String+SofaHelpers.swift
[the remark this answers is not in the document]
marijnschilling: yups, shall i add it to the static methods then?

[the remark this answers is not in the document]
yuliaveres: yeah, why dnt we want previous way of adding / removing type string component ?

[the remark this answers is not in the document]
yuliaveres: the ones u indicated ..

Toshi/Models/SofaWrappers/SofaPaymentRequest.swift
```diff
@@ -0,0 +1,72 @@
...
+import Foundation
+
+open class SofaPaymentRequest: SofaWrapper {
+    public override var type: SofaType {
+        return .paymentRequest
+    }
+
+    public var body: String {
+        guard content.hasPrefix(type.rawValue) else {
+            fatalError("Creating SofaMessage with invalid type!")
```
yuliaveres: hmm 🤔

Toshi/Models/SofaWrappers/SofaInitialRequest.swift
marijnschilling: I don't know. What do you suggest? `accessing`?

yuliaveres: i mean why check it on each access in general. i would do a check on initialization and be sure all existing are valid .. 🤔  ..?

Toshi/Models/SofaWrappers/SofaInitialRequest.swift
marijnschilling: I guess this is to doubleCheck the type of the content fits with the function? I'm also not 💯 sure... I didnt touch this

yuliaveres: it checks for whether it has in case of initial request  "SOFA:InitialRequest.." component. i dnt see the reason of doing this. we need to remove it and do all checks while initializing an instance ? do u agree ?

Toshi/Extensions/String+SofaHelpers.swift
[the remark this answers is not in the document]
marijnschilling: I thought it was super unclear what was going on there. And also should be tested right? Seems quite a vulnerable part of the app.

Before we only used it once so that's fine I guess. But since now I need to use it more i thought it would be better to just create a separate clearly named method for it with tests.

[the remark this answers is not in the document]
yuliaveres: ye that makes sense ! then lets create a static one as u proposed ? 👍

Toshi/Models/SofaWrappers/SofaInitialRequest.swift
marijnschilling: I really dare not say this, I don't know how and when an why this is used

yuliaveres: i think it is actually already handled in the way i described, otherwise we would have tonnes of fatal errors 😄

yuliaveres: i think we can get rid of it safely 🤞

marijnschilling: Oke i will dispose it! 🗑

yuliaveres:
@marijnschilling  the only issue left is received payments. The problem lies here:
when we send payments, we are creators and adjust `fiatValueString` upon save during initialization, then in `ChatInteractor` when "handle signal message" each time, the `fiatValueString` is not recreated and added each time as it is present already.

BUT IN CASE OF RECEIVED message we do not own it and we already have payload string ("SOFA:Payment....") then while "handle signal message" we do add `fiatValueString` and it looks ok. But that happens every single time as when retrieving that received message from DB the payload string is as it was before (without `fiatValueString`) BECAUSE WE DO NOT SAVE message after adding  this string. So I guess in `ChatInteractor` while "handling signal message" we should check if payload string of the message does not have `fiatValueString` - and it was just added - we should update the content string and save message (better in background) (`ChatInteractor.swift 218`:
```  
interaction.body = sofaWrapper.content
interaction.save()
```
) . Thus next time we process it here, it will contain fiat value string and we won't add it again. And the most important - it will be displayed on correct currency. 😉  let me know if you have any issue with that. 🤔

marijnschilling:
I checked this before i decided on this approach, and this does not happen. The fiatValue String only get's added the first time we receive a payment or paymentRequest.

yuliaveres:
it is not. you can check this while changing user currency - only received message is displayed in new currency. every time

marijnschilling:
Hmmm.. That is a weird bug somewhere else. I will look into it!

yuliaveres:
it is not somewhere else. u need to save message with changed content string

PR #555, opened 2017-10-27
[IOS-362]: Search only among local favorite users

yuliaveres:
Use filtered view for search on favorites; we need to search only among those, do not retrieve remote users with api client. All users are searched in this was on Browse only.

*Also* : i decided to use filtered view as FTS is not purposed for local single view search, it is more for overall global search.


Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -46,7 +70,7 @@ open class FavoritesController: SweetTableController, KeyboardAdjustable, Emptia
     }
 
     fileprivate lazy var mappings: YapDatabaseViewMappings = {
-        let mappings = YapDatabaseViewMappings(groups: [TokenUser.favoritesCollectionKey], view: TokenUser.viewExtensionName)
+        let mappings = YapDatabaseViewMappings(groups: [TokenUser.favoritesCollectionKey], view: self.filteredDatabaseViewName)
```
marijnschilling: `self` is not needed here

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -226,17 +251,7 @@ open class FavoritesController: SweetTableController, KeyboardAdjustable, Emptia
...
-        let hasFavourites = mappings.numberOfItems(inSection: 0) > 0
-
-        if searchController.isActive {
-            shouldHideEmptyState = !searchContacts.isEmpty || hasFavourites
-        } else {
-            shouldHideEmptyState = hasFavourites
-        }
-
-        emptyView.isHidden = shouldHideEmptyState
+        emptyView.isHidden = mappings.numberOfItems(inSection: 0) > 0
```
marijnschilling: I think we should change the criterea of when we show or hide the empty view.

Now we also show it when we search in favorites and there are no search results. Which is weird.

When the search text is empty and `mappings.numberOfItems(inSection: 0) > 0` we should show the empty view, but when the searchText is not empty we should show nothing (or maybe show a message stating there where no results. but that should be discussed with Robert I guess)

PR #557, opened 2017-10-30
Distinguish between url schema.

roberthein:
Make a distinction in the different plists to not get url schema between different versions of the app mixed up.


roberthein:
👍

PR #558, opened 2017-11-01
[IOS-365]: Cache on currencies list

yuliaveres:
Currently we do fetch currencies on each appear which feels delayed. Now i have added caching. so we use persisted data first and then fetched.


Toshi/Models/Currency.swift
```diff
@@ -17,10 +17,48 @@
...
+    init?(coder aDecoder: NSCoder) {
+        if let apps = aDecoder.decodeObject(forKey: "objects") as? [Currency] {
+            self.objects = apps
+        }
+
+        super.init()
+    }
+}
+
+public final class Currency: NSObject, NSCoding {
```
roberthein: What's the advantage in using `NSCoding` vs the new `Codable` protocol?

Toshi/Models/Currency.swift
```diff
@@ -17,10 +17,48 @@
 
 import Foundation
 
-public struct Currency {
+final class CurrenciesCacheData: NSObject, NSCoding {
```
roberthein: If we could use `Codable` instead this could stay a struct.

Toshi/APIClients/ExchangeRateAPIClient.swift
```diff
@@ -83,12 +95,17 @@ public final class ExchangeRateAPIClient {
     }
 
     public func getCurrencies(_ completion: @escaping (([Currency]) -> Void)) {
-        self.teapot.get("/v1/currencies") { (result: NetworkResult) in
+
+        if let data = cache.object(forKey: currenciesCacheKey), let currencies = data.objects {
+            completion(currencies)
+        }
+
+        self.teapot.get("/v1/currencies") { [weak self] (result: NetworkResult) in
```
roberthein: ⚠️ `self`-warning ⚠️

Toshi/Models/Currency.swift
```diff
@@ -17,10 +17,48 @@
...
+        super.init()
+    }
+
+    func encode(with aCoder: NSCoder) {
+        aCoder.encode(objects, forKey: "objects")
+    }
+
+    init?(coder aDecoder: NSCoder) {
+        if let apps = aDecoder.decodeObject(forKey: "objects") as? [Currency] {
+            self.objects = apps
```
roberthein: ⚠️ `self`-warning ⚠️

Toshi/Models/Currency.swift
yuliaveres: i need to use  `Codable`  for the FIRST time 😏

yuliaveres: looks like AwesomeCache uses NSCoding cause i am having a crash `[_SwiftValue encodeWithCoder:]: unrecognized selector sent ... `

PR #561, opened 2017-11-02
Toshi error

marijnschilling:
To be able to make use of the new Teapot errors I made a `ToshiError` than can be initialised with a `TeapotError`, optionally you can add a custom `errorDescription` when you initialise, otherwise it will take the one from `TeapotError`

I created tests for the initialisation, there was just a super weird thing with the `Nimble` tests where the string comparison for the `errorDescription` did not work. (Thats why this took so long 🙄 )
I fixed it by adding a good ol' `XCTest`.

I think if we merge this we can start implementing this error in way more places in Toshi.

For now at least it fixed this issue: https://toshiapp.atlassian.net/browse/IOS-337

I needed to make some changes to Teapot, make some properties of TeapotError public to be able to do this. So the cartfile now points to that branch of teapot. I should look at that with Igor to merge that and make the Cartfile point to the master again.


Toshi/APIClients/ToshiError.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    let responseStatus: Int?
+    let underlyingError: Error?
+    let type: ErrorType
+
+    public var errorDescription: String
+}
+
+extension ToshiError {
+    init?(withTeapotError teapotError: TeapotError, errorDescription: String? = nil) {
```
yuliaveres: hm. wondering about 2 points:
 - why we extracted initializer to an extention ?
- i would write `init?(_ teapotError: TeapotError, errorDescription: String? = nil) ` ... wdyt ?

Toshi/Controllers/Payments/Payment.swift
```diff
@@ -23,7 +23,7 @@ class PaymentManager {
                 
                 if let error = error {
                     DispatchQueue.main.async {
-                        self?.showPaymentFailedMessage(for: error.localizedDescription)
+                        self?.showPaymentFailedMessage(for: error.errorDescription)
```
yuliaveres: `errorDescription` doesn't sound like the best name for the variable here ... 🤔  can't we name it `description` ?

Toshi/APIClients/ToshiError.swift
```diff
@@ -0,0 +1,34 @@
...
+import Foundation
+import Teapot
+
+public struct ToshiError: LocalizedError {
+    enum ErrorType: Int {
+        case dataTaskError
+        case invalidPayload
+        case invalidRequestPath
+        case invalidResponseStatus
+        case missingImage
```
yuliaveres: why do we need to have our own `ErrorType` ? 🤔  all of those cases are contained by `TeapotError.ErrorType` ...

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -61,18 +61,18 @@ public class EthereumAPIClient: NSObject {
...
             var resultString: String?
-            var resultError: Error?
+            var resultError: ToshiError?
 
             switch result {
             case .success(let json, _):
                 resultString = json?.dictionary!["tx"] as? String
             case .failure(_, _, let error):
-                resultError = error
+                resultError = ToshiError(withTeapotError: error, errorDescription: "Error creating transaction: do you have the correct recipient's waller address?")
```
yuliaveres: - "... wallet address" vs "...waller address"
- are we sure the error here is always related to recipient's address ?

Toshi/APIClients/ToshiError.swift
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    let responseStatus: Int?
+    let underlyingError: Error?
+    let type: ErrorType
```
yuliaveres: can't we use `TeapotError.ErrorType` here ?

Toshi/APIClients/ToshiError.swift
marijnschilling: Because i don't want to have the Error that strongly linked to a framework we use. I will defenately add it's own initializer when further implementing it. 

If we change using teapot to use something else we can just remove this extension and there's no harm done. Do thats also why i write `teapotError` in the initialiser, because it's special!

yuliaveres: but what about the label `withTeapotError ` ? can we use `_` ? i meant words duplication `withTeapotError teapotError: ....`

marijnschilling: I owuld only use a blank parameter title if it was clear from the context what the parameter would be.. So if it was a normal function I would call it `updateWithTeapotError(_ teapotError: TeapotError)` But since it's an init i would give the parameter a clear name...

Toshi/APIClients/EthereumAPIClient.swift
marijnschilling: Nope 🤐  I will make it more generic 😅

PR #562, opened 2017-11-02
[IOS-366]: Improve avatar loading

yuliaveres:
Added some improvement for avatars loading:
increased amount of concurrent operations on download queue;
launched avatars download for all apps/ users fetched on browse, thus when opening *More* on some category avatars are already present and not blinking;
removed not needed cache expiry on AvatarsManager as not each URL corresponds to the unique resource;
download each avatar in separate background queue;
do not update browse cell's avatar imageView after request response arrived and cell's index path is not visible anymore.

(Previously when scrolled to the right on some category and then back to the left, and connection was poor so avatars were still loading, the first cell f.e. was updated with the last cell response image and if printing out indexPath in completion - the last one was shown, that is why if check for visibility of given indexPath, cell won't be updated with wrong avatar).


Toshi/Avatars/AvatarManager.swift
```diff
@@ -119,29 +113,31 @@ final class AvatarManager: NSObject {
...
-                resultImage = image
-            case .failure(let response, let error):
-                print(response)
-                print(error)
-            }
-
-            DispatchQueue.main.async {
-                completion(resultImage)
+            teapot.get(url.relativePath) { [weak self] (result: NetworkImageResult) in
+                guard let strongSelf = self else {
```
roberthein: What's the advantage for using `strongSelf` instead of using `self?.doSomething()`?

Toshi/Avatars/AvatarManager.swift
```diff
@@ -119,29 +113,31 @@ final class AvatarManager: NSObject {
...
+                    completion(nil)
+                    return
+                }
+
+                var resultImage: UIImage?
+                switch result {
+                case .success(let image, let response):
```
roberthein: `response` is not being used.

Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -345,4 +353,22 @@ extension BrowseViewController: BrowseCollectionViewCellSelectionDelegate {
             Navigator.push(ContactController(contact: item))
         }
     }
+
+    func willDisplayCell(_ cell: UICollectionViewCell, at indexPath: IndexPath, _ collectionView: UICollectionView) {
+
+        if let sectionedCollectionView = collectionView as? SectionedCollectionView, let cell = cell as? BrowseEntityCollectionViewCell {
+
+            self.avatar(for: indexPath, in: sectionedCollectionView.section, completion: { [ weak self ] image in
```
roberthein: You don't need `self` here and there's some weird spacing with the weakifying.

Toshi/Avatars/AvatarManager.swift
yuliaveres: there is no advantage, i use both `strongSelf` and `self?` depending on what i need. in case of `.doSmth()` i prefer `self?.` but if i use *self* for something which might be optional due to self being optional, i prefer guarding to `strongSelf` and avoid further unwraps (in this case it is `visibleCategories` which would be optional if i used `self?.`)

PR #563, opened 2017-11-02
Add no camera access alert to scanner view controller.

roberthein:
Add no camera access alert to scanner view controller.

### DIFF
--- a/Toshi/Controllers/ScannerController.swift
+++ b/Toshi/Controllers/ScannerController.swift
@@ -15,6 +15,7 @@
 
 import UIKit
 import CameraScanner
+import AVKit
 
 class ScannerController: ScannerViewController {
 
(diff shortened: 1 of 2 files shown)


Toshi/Controllers/ScannerController.swift
```diff
@@ -37,6 +38,23 @@ class ScannerController: ScannerViewController {
         
         navigationController?.setNavigationBarHidden(false, animated: true)
     }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        guard AVCaptureDevice.authorizationStatus(for: .video) == .denied else { return }
```
marijnschilling: I would recommend changing this from a guard to a good old if statement. Since not having video access is an exception right? I find it a bit weird to use a guard for the case that we do not wish to be in.

Toshi/Resources/Localizable.strings
```diff
@@ -159,6 +159,10 @@
 "sign-out-failure-message" = "Error attempting to unregister from chat service. Our engineers are looking into it.";
 "cancel_action_title" = "Cancel";
 
+"no_camera_access_title" = "Camera access";
+"no_camera_access_description" = "Toshi does not have access to the camera. Would you like to change this in settings?";
```
marijnschilling: Huh i get the error with this copy:

screenshot 2017-11-03 14 24 53

yuliaveres:
@roberthein  can we make fixes and close this ?

roberthein:
Not applicable anymore.

yuliaveres:
why it is not applicable anymore ?

roberthein:
It's already implemented by the camera controller.

PR #569, opened 2017-11-07
Shared contact view, and preferredContentSizeChanged true all the things

marijnschilling:
- [ ] made the `ProfileController` and the `ContactController` share a `ContactView` that can be set up in the two different states.
- [ ] Don't use preferred fonts on navigation titles and labels
- [ ] `preferredContentSizeChanged = true` everywhere where we use preferred fonts. If you know a way to do this in a cleaner way hit me up! The only thing i could think of was always using a Subclass of `UILabel`, which sucks.

it fixes some of the issues of this task: https://toshiapp.atlassian.net/browse/IOS-369 but not yet all of them, otherwise the PR would become waaay too big 😬


Toshi/Controllers/Favorites/ContactController.swift
```diff
@@ -35,133 +36,11 @@ public class ContactController: UIViewController {
         return appDelegate?.messageSender
     }
 
-    lazy var avatarImageView: AvatarImageView = {
-        let view = AvatarImageView(withAutoLayout: true)
+    fileprivate lazy var contactView: ContactView = {
```
roberthein: I think it would be nicer to let the view controller load the `ContactView` as it's view like this:
```
override func loadView() {
    view = ContactView(.contact)
}
```
Then we don't have to add any subviews in the controller and we don't have to import TinyConstraints.

Toshi/Controllers/Settings/Profile/ProfileController.swift
```diff
@@ -23,126 +23,10 @@ open class ProfileController: UIViewController {
...
-        view.delaysContentTouches = false
-        view.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: 49, right: 0)
-
-        return view
-    }()
-
-    private lazy var container: UIView = {
-        let view = UIView()
-
+    fileprivate lazy var contactView: ContactView = {
```
roberthein: I think it would be nicer to let the view controller load the `ContactView` as it's view like this:
```
override func loadView() {
    view = ContactView(. profile)
}
```
Then we don't have to add any subviews in the controller and we don't have to import TinyConstraints.

Toshi/Views/ContactView.swift
```diff
@@ -0,0 +1,387 @@
...
+        } else {
+            nameLabel.text = user.displayUsername
+            usernameLabel.text = nil
+        }
+
+        aboutContentLabel.text = user.about
+        locationContentLabel.text = user.location
+    }
+
+    lazy var avatarImageView: AvatarImageView = {
```
roberthein: You can replace:
```
lazy var avatarImageView: AvatarImageView = {
    let view = AvatarImageView(withAutoLayout: true)

    return view
}()
```
with
```
lazy var avatarImageView = AvatarImageView()
```

Toshi/Views/ContactView.swift
```diff
@@ -0,0 +1,387 @@
...
+        let view = AvatarImageView(withAutoLayout: true)
+
+        return view
+    }()
+
+    lazy var nameLabel: UILabel = {
+        let view = UILabel(withAutoLayout: true)
```
roberthein: You can omit `withAutoLayout: true` for every view because your using TinyConstraints.

Toshi/Views/ContactView.swift
```diff
@@ -0,0 +1,387 @@
...
+        let view = UIView(withAutoLayout: true)
+        view.backgroundColor = Theme.lightGrayBackgroundColor
+        view.layer.borderColor = Theme.borderColor.cgColor
+        view.layer.borderWidth = 1.0 / UIScreen.main.scale
+        view.set(height: 1.0 / UIScreen.main.scale)
+
+        return view
+    }()
+
+    lazy var reputationView: ReputationView = {
```
roberthein: You can replace:
```
lazy var reputationView: ReputationView = {
    let view = ReputationView(withAutoLayout: true)

    return view
}()
```
with
```
lazy var reputationView = ReputationView()
```

Toshi/Views/ContactView.swift
```diff
@@ -0,0 +1,387 @@
...
+
+    private lazy var scrollView: UIScrollView = {
+        let view = UIScrollView()
+        view.showsVerticalScrollIndicator = true
+        view.delaysContentTouches = false
+
+        return view
+    }()
+
+    private lazy var container: UIView = {
```
roberthein: You can replace:
```
private lazy var container: UIView = {
    let view = UIView()

    return view
}()
```
with
```
private lazy var container = UIView()
```

Toshi/Views/ContactActionView.swift
```diff
@@ -125,6 +125,8 @@ class ContactActionView: UIStackView {
 
         self.distribution = .equalSpacing
         self.isLayoutMarginsRelativeArrangement = true
+        self.layoutMargins.top = 16
```
roberthein: I see `self.` 16 times too many, can we remove it?

Toshi/Controllers/Sign In/SplashNavigationController.swift
```diff
@@ -29,7 +29,7 @@ public class SplashNavigationController: UINavigationController {
         navigationBar.shadowImage = UIImage()
         
         let titleTextAttributes: [NSAttributedStringKey: Any] = [
-            .font: Theme.preferredRegular(),
+            .font: Theme.regular(size: 17),
```
roberthein: 👍

Toshi/Theme/Theme.swift
```diff
@@ -29,12 +29,12 @@ extension Theme {
     
     @objc public static func setupBasicAppearance() {
         let navBarAppearance = UINavigationBar.appearance()
-        navBarAppearance.titleTextAttributes = [.font: Theme.preferredSemibold(), .foregroundColor: Theme.navigationTitleTextColor]
+        navBarAppearance.titleTextAttributes = [.font: Theme.semibold(size: 17), .foregroundColor: Theme.navigationTitleTextColor]
```
roberthein: 👍

Toshi/Theme/Theme.swift
```diff
@@ -29,12 +29,12 @@ extension Theme {
...
         navBarAppearance.tintColor = Theme.tintColor
         navBarAppearance.barTintColor = Theme.navigationBarColor
         
         let barButtonAppearance = UIBarButtonItem.appearance(whenContainedInInstancesOf: [UINavigationBar.self])
-        barButtonAppearance.setTitleTextAttributes([.font: Theme.preferredRegular(), .foregroundColor: Theme.tintColor], for: .normal)
+        barButtonAppearance.setTitleTextAttributes([.font: Theme.regular(size: 17), .foregroundColor: Theme.tintColor], for: .normal)
```
roberthein: 👍

Toshi/Controllers/Favorites/ContactController.swift
marijnschilling: hmmm but i I want to call something on the contactView i'll have to cast the view to a contactView first.
```
(view as? ContactView).reputationView.setScore(.zero)
```

Toshi/Views/ContactActionView.swift
marijnschilling: haha old code :)

roberthein: But we can improve it when we come across it, right? Now you added new code in the old style..

marijnschilling: Oops, forgot this issue 🙈

marijnschilling:
uh yeah! Organizing! 🗃  On it

PR #570, opened 2017-11-08
[IOS-378]: Exclude DB file from backup

yuliaveres:
We should not sync db file with iCloud to prevent from weird cases, f.e. after user changes device. And  for security reasons


Toshi/Models/Yap.swift
```diff
@@ -132,6 +132,21 @@ public final class Yap: NSObject, Singleton {
...
+            addSkipBackupAttributeToDatabasePath()
+        }
+    }
+
+    private func addSkipBackupAttributeToDatabasePath() {
+        var databaseUrl = NSURL.fileURL(withPath: UserDB.dbFilePath)
+
+        do {
+            var resourceValues = URLResourceValues()
+            resourceValues.isExcludedFromBackup = true
```
Elland: You could also check the `is ubiquitous` thing, but I think SSK or Yap already do it. 🤔

yuliaveres: not sure why i might want to check it here

Elland: You can double-check if you want, since you updated YapDatabase and SSK. I'm quite sure the last version I added removed the db from iCloud. 👍 Can't see why they'd change it.

yuliaveres: why might i want to check if a file at the given path was synced to icloud ?

yuliaveres: `if this item is synced to the cloud` .. i need to change this anyway

Elland: All already being handled https://github.com/toshiapp/toshi-ios-client/pull/570/files#diff-e792199191a879a8db3071bd60c4cf6eR128. But it's possible this is failing silently. Try using a `do/catch` block instead, so we can log if it fails? 👍

yuliaveres: 👍

Toshi/Models/Yap.swift
```diff
@@ -123,18 +123,28 @@ public final class Yap: NSObject, Singleton {
...
         mainConnection = database?.newConnection()
 
         if database == nil {
             CrashlyticsLogger.log("Failed to create user database")
+        } else {
+            addSkipBackupAttributeToDatabasePath()
         }
     }
 
+    private func addSkipBackupAttributeToDatabasePath() {
```
Elland: 👏🏽

Toshi/Models/Yap.swift
```diff
@@ -123,18 +123,28 @@ public final class Yap: NSObject, Singleton {
...
+        let url = NSURL(fileURLWithPath: UserDB.dbFilePath)
+        do {
+            try url.setResourceValue(false, forKey: .isUbiquitousItemKey)
+            try url.setResourceValue(true, forKey: .isExcludedFromBackupKey)
+        } catch let error {
```
roberthein: `error` is not being used..

Toshi/Models/Yap.swift
```diff
@@ -123,18 +123,28 @@ public final class Yap: NSObject, Singleton {
...
+            CrashlyticsLogger.log("Failed to exclude DB from backup and sync")
+        }
+
+        var databaseUrl = NSURL.fileURL(withPath: UserDB.dbFilePath)
```
roberthein: `databaseUrl` is not being used.

PR #571, opened 2017-11-08
Dismiss search when going from Browse on iOS 10

yuliaveres:
The issue described here:
https://toshiapp.atlassian.net/browse/IOS-376


Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -114,79 +112,78 @@ class BrowseViewController: SearchableCollectionController {
...
-
+        
         if let indexPathForSelectedRow = searchResultView.indexPathForSelectedRow {
             searchResultView.deselectRow(at: indexPathForSelectedRow, animated: true)
         }
+        
+        searchBar.text = nil
     }
-
+
```
yuliaveres: can we get rid of all those white spaces ?

yuliaveres:
Just need to hide cancel button after resigning first responder on search bar, and 👌 
and ✅

Toshi/Controllers/Browse/BrowseViewController.swift
roberthein: This is a problem throughout the app caused by *Swift Format* which removed the whitespace that Xcode automatically adds.

roberthein:
Good catch bout that cancel button!

PR #574, opened 2017-11-09
[IOS-382]: Disable messaging with current user

marijnschilling:
- [ ] Don't show the actionView when the profile is shown of current user
- [ ] also don't show the `...` menu in the top right corner, so you cant block or report yourself.

screenshot 2017-11-13 11 16 59

fixes: https://toshiapp.atlassian.net/browse/IOS-382


Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -50,6 +50,10 @@ class ProfileView: UIView {
 
         aboutContentLabel.text = user.about
         locationContentLabel.text = user.location
+
+        if user.address == (TokenUser.current?.address ?? nil) {
```
yuliaveres: we can use `isCurrentUser` on the `user` 😉

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -33,6 +33,7 @@ class ProfileView: UIView {
     enum ViewType {
         case personalProfile
         case profile
```
yuliaveres: aren't those different in a way we need ? (`profile` - someone's else profile, `personalProfile ` - current user profile only ..) afaik - yes

yuliaveres: IMHO, we do not need `PersonlaProfileViewController`. it is identical to `ProfileViewController`, which we could easily use for all cases. Initialize it with `TokenUser` object only,
inside - `viewType` will be picked depending on its `user`'s `isCurrentUserProperty`. Then we do not need `. isCurrentUserProfile `. Just `personalProfile ` and `profile `.

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -57,7 +57,11 @@ public class ProfileViewController: UIViewController {
     }
 
     open override func loadView() {
-        view = ProfileView(viewType: .profile)
+        if contact.isCurrentUser {
+            view = ProfileView(viewType: .isCurrentUserProfile)
```
yuliaveres: that is exactly why we can use just `ProfileViewController`, cause we handle current user and other users there already

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -31,9 +31,9 @@ protocol ProfileViewDelegate: class {
 class ProfileView: UIView {
 
     enum ViewType {
-        case personalProfile
         case profile
-        case isCurrentUserProfile
+        case personalProfile
+        case personalProfileReadOnly
```
yuliaveres: 👍

PR #575, opened 2017-11-13
Enable the cache when running tests

marijnschilling:
screenshot 2017-11-13 13 28 02

All tests succeed again!

Added option to disable cache in init with teapot, and put the cache return in if else block.


Toshi/APIClients/ExchangeRateAPIClient.swift
```diff
@@ -98,34 +102,37 @@ public final class ExchangeRateAPIClient {
 
         if let data = cache.object(forKey: currenciesCacheKey), let currencies = data.objects {
             completion(currencies)
-        }
-
-        teapot.get("/v1/currencies") { [weak self] (result: NetworkResult) in
-            var results: [Currency] = []
+        } else {
```
yuliaveres: what is the reason we dnt make api call if we have cached objects ?

marijnschilling: I thought it would be double, but i guess we first show the cashed information and thén go out to the server to update it? I will adjust!

PR #579, opened 2017-11-14
Added a cache queue and changed the avatar image assignment flow.

Elland:
Improved performance, by removing the bottleneck in the main queue. AwesomeCache uses a sync queue internally to fetch items. By calling it directly from the main queue, we would block it until the cache returned something.

By asking the collection view for the cell instead of using a pointer to the previously dequeued one, we avoid issues where the cache might hang and then set the avatar to the wrong cell, by it being dequeued again for a different index path, in the cases where scrolling happens very fast (now possible, since no more hiccups). 

Worth checking if this could also apply to other view controllers in the app.


Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -298,9 +298,15 @@ extension BrowseViewController: UISearchBarDelegate {
...
                 }
-                
-                cell.avatarImageView.image = AvatarManager.shared.cachedAvatar(for: item.avatarPath)
-                
+
+                cacheQueue.async {
+                    let image = AvatarManager.shared.cachedAvatar(for: item.avatarPath)
+                    DispatchQueue.main.async {
+                        let cell = sectionedCollectionView.cellForItem(at: indexPath) as? BrowseEntityCollectionViewCell
+                        cell?.avatarImageView.image = image
```
Elland: I'm not a fan of cascading calls to subviews of a view directly, and much prefer the `UIButton` style of abstracting them away. `cell?.avatar = image` perhaps?

Elland:
@marijnschilling if you want to take a look too.

Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -298,9 +298,15 @@ extension BrowseViewController: UISearchBarDelegate {
...
                     cell.nameLabel.text = item.isApp ? item.category : item.username
                 }
-                
-                cell.avatarImageView.image = AvatarManager.shared.cachedAvatar(for: item.avatarPath)
-                
+
+                cacheQueue.async {
+                    let image = AvatarManager.shared.cachedAvatar(for: item.avatarPath)
+                    DispatchQueue.main.async {
+                        let cell = sectionedCollectionView.cellForItem(at: indexPath) as? BrowseEntityCollectionViewCell
```
yuliaveres: why would u retrieve cell here ?

yuliaveres:
since we are using `cachedAvatar..`, please remove not needed `willDisplayCellFor..`. it is leftover.

Toshi/Controllers/Browse/BrowseViewController.swift
heumn: I presume it is because if you fetch a cell while scrolling and the avatar is being fetched you might end up putting the avatar on the wrong cell (because the cell you have a pointer to now has been re-used during the scroll). This could lead to some odd image flickering 🤔

yuliaveres: how fast should scroll be that cell is reused here ( in the cache retrieval block .. which is not network request completion that might take long, in that case - of course .. ) 🤔

Toshi.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
marijnschilling: What are these changes?

Toshi/Controllers/Browse/BrowseViewController.swift
Elland: It's a disc operation, if there's a process lock on the file, it could take a while.

Elland:
Should I merge or wait for @heumn ?

heumn:
😬  never wait for me! 🚪  ✅ 

👏

PR #580, opened 2017-11-15
Toshi error

marijnschilling:
Implement custom error types to handle of errors in toshi


Toshi/APIClients/AppsAPIClient.swift
```diff
@@ -58,12 +58,12 @@ public class AppsAPIClient: NSObject, CacheExpiryDefault {
...
 
         teapot.get("/v1/search/apps?top=true&recent=false&limit=\(limit)") { [weak self] (result: NetworkResult) in
-            var resultsError: Error?
+            var resultsError: ToshiError?
             var results: [TokenUser] = []
             switch result {
             case .success(let json, _):
                 guard let strongSelf = self, let json = json?.dictionary, let appsJSON = json["results"] as? [[String: Any]] else {
-                    completion(nil, nil)
+                    completion(nil, ToshiError.invalidResponseJSON)
```
yuliaveres: can we say `.invalidResponseJSON` ? or no ?

Toshi/APIClients/ToshiError.swift
```diff
@@ -6,29 +6,64 @@
 import Foundation
 import Teapot
 
-public struct ToshiError: LocalizedError {
+public struct ToshiError: Error, CustomStringConvertible {
+    static func dataTaskError(withUnderLyingError error: Error) -> TeapotError {
+        let errorDescription = String(format: NSLocalizedString("toshi_error_data_task_error", bundle: Teapot.localizationBundle, comment: ""), error.localizedDescription)
```
yuliaveres: seems we missed this one `toshi_error_data_task_error` in localized strings file ....

Toshi/APIClients/ToshiError.swift
```diff
@@ -6,29 +6,64 @@
...
         case dataTaskError
         case invalidPayload
         case invalidRequestPath
         case invalidResponseStatus
-        case missingImage
+        case invalidResponseJSON
     }
 
+    let type: ErrorType
+    public var description: String
```
yuliaveres: we can separate `public var` from all `let`s .. 😉

Toshi/APIClients/ToshiError.swift
```diff
@@ -6,29 +6,64 @@
...
+    init(withType errorType: ErrorType, description: String, responseStatus: Int? = nil, underlyingError: Error? = nil) {
+        self.type = errorType
+        self.description = description
+        self.responseStatus = responseStatus
+        self.underlyingError = underlyingError
+    }
 }
 
 extension ToshiError {
+    static func teapotErrorTypeToToshiErrorType(teapotErrorType: TeapotError.ErrorType) -> ErrorType? {
```
yuliaveres: maybe better to have it without label ? `static func teapotErrorTypeToToshiErrorType(_ teapotErrorType: TeapotError.ErrorType) -> ErrorType?`

**then call:** `ToshiError.teapotErrorTypeToToshiErrorType(teapotError.type)` **instead of** `ToshiError.teapotErrorTypeToToshiErrorType(teapotErrorType: teapotError.type)`

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -159,13 +159,13 @@ final class ChatInteractor: NSObject {
...
 
-            self?.etherAPIClient.sendSignedTransaction(originalTransaction: transaction, transactionSignature: signedTransaction) { [weak self] success, json, message in
+            self?.etherAPIClient.sendSignedTransaction(originalTransaction: transaction, transactionSignature: signedTransaction) { [weak self] success, json, error in
 
                 self?.output?.didFinishRequest()
 
                 guard success, let json = json?.dictionary else {
                     DispatchQueue.main.async {
-                        self?.output?.didCatchError(message ?? "Something went wrong")
+                        self?.output?.didCatchError(error?.description ?? Localized("request_generic_error"))
```
yuliaveres: 🤔  from what i see now we target to some default generic errors if some info is missing from a received error...
then why we do not return a generic type error from f.e. `teapotErrorTypeToToshiErrorType()` ? ..
in that function we do expect to have expected type, otherwise we do not have error type. While it might be better to have a generic error type if predefined can't be composed. ? wdyt ? 
the errors logic should be consistent prolly ..

Toshi/Resources/Localizable.strings
```diff
@@ -232,3 +232,8 @@
 /* Keychain and file system error */
 "yap_missing_password_error_description" = "No keychain password while database present";
 "yap_missing_db_file_error_description" = "No database file while keychain password present";
+
+
+/* Toshi error messages*/
+"toshi_error_invalid_payload" = "Invalid payload, request could not be executed";
```
yuliaveres: missing one key mentioned previously ^^ ..

Tests/AppsAPIClientTests.swift
```diff
@@ -75,7 +75,7 @@ class AppsAPIClientTests: QuickSpec {
                     waitUntil { done in
                         subject.getFeaturedApps { users, error in
                             expect(users?.count).to(equal(0))
-                            expect(error).toNot(beNil())
+                            expect(error?.description).to(equal("An error occurred: request response status reported an issue. Status code: 401."))
```
yuliaveres: not sure why do we expect this description here and on all other tests ? (with status code mentioned ..) ? 🙃

marijnschilling: Well, this way ti'm just checking wether we get the expected error in the fault case. maybe better to check for errorType though ✌️

Toshi/APIClients/ToshiError.swift
marijnschilling: I actually think we should make them all public let's :)

Toshi/Controllers/Chat/ChatInteractor.swift
marijnschilling: It's just because error is an optional in this case. The guard checks for a dictionary being not nil. So we can't be sure that there will be even an error :/ Maybe we should change the guard statement...

marijnschilling: Ooh i get it I think! Just have a 

ToshiError.genericError ✌️ Sounds good!

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -218,9 +218,9 @@ extension ProfileViewController: ActivityIndicating {
 extension ProfileViewController: RateUserControllerDelegate {
     func didRate(_ user: TokenUser, rating: Int, review: String) {
         dismiss(animated: true) {
-            RatingsClient.shared.submit(userId: user.address, rating: rating, review: review) { [weak self] success, message in
+            RatingsClient.shared.submit(userId: user.address, rating: rating, review: review) { [weak self] success, error in
                 guard success == true else {
-                    let alert = UIAlertController(title: "Error", message: message, preferredStyle: .alert)
+                    let alert = UIAlertController(title: "Error", message: error?.description, preferredStyle: .alert)
```
yuliaveres: lets localize those ?

Toshi/APIClients/ToshiError.swift
```diff
@@ -1,34 +1,80 @@
...
+        case .invalidResponseStatus:
+            return .invalidResponseStatus
+        case .dataTaskError:
+            return .dataTaskError
+        case .invalidPayload:
+            return .invalidPayload
+        case .invalidRequestPath:
+            return .invalidRequestPath
+        default:
+            return nil
```
yuliaveres: maybe better return `.genericError` if unexpected toshi error was received (some recently added which we do not support yet ?) ..

Toshi/APIClients/ToshiError.swift
```diff
@@ -1,34 +1,80 @@
...
+        }
+    }
+
     init?(withTeapotError teapotError: TeapotError, errorDescription: String? = nil) {
-        guard let errorType = ErrorType(rawValue: teapotError.type.rawValue) else { return nil }
+        guard let errorType = ToshiError.teapotErrorTypeToToshiErrorType(teapotError.type) else { return nil }
```
yuliaveres: maybe if some existing `teapotError ` occured, was passed here but for some reason , `ToshiError` initialization failed - we do not have an errro in the end as nil will be returned here ? maybe better to not let the initialization failure and in some unexpected case initialize just `.generic` error instead of `nil` ?

PR #582, opened 2017-11-15
[IOS-395]: Avatar API extension

yuliaveres:
Support avatar retrieval / download for a user id, apart from already avatarPath.


Toshi/Avatars/AvatarManager.swift
```diff
@@ -75,19 +60,19 @@ final class AvatarManager: NSObject {
...
-                self?.downloadAvatar(for: path)
+                self?.downloadAvatarForPath(path)
             }
         }
 
         downloadOperationQueue.addOperation(operation)
     }
 
-    func downloadAvatar(for path: String) {
+    func downloadAvatarForPath(_ path: String) {
```
roberthein: 😢

PR #584, opened 2017-11-16
Add ability to copy QR code image/content.

Elland:
Sharing your QR code is unnecessarily cumbersome. Being able to copy the image/URL helps. I could also look into adding a share sheet.

<image src="https://user-images.githubusercontent.com/54423/32895976-104c23b2-cae2-11e7-82c0-0f0dad92d4ef.png" width=300>

### DIFF
--- a/Toshi/Controllers/Settings/QRCodeController.swift
+++ b/Toshi/Controllers/Settings/QRCodeController.swift
@@ -18,11 +18,11 @@ import SweetUIKit
 import CoreImage
 import MobileCoreServices
 
-class SelectableImageView: UIImageView {
-    private var username: String
+final class SelectableImageView: UIImageView {
+    private var stringRepresentation: String
 
-    init(username: String) {
-        self.username = username
+    init(stringRepresentation: String) {
+        self.stringRepresentation = stringRepresentation
 
         super.init(frame: .zero)
 
@@ -50,13 +50,13 @@ class SelectableImageView: UIImageView {
     override func copy(_ sender: Any?) {
         UIPasteboard.general.items = [
             [kUTTypeJPEG as String: image!],
-            [kUTTypeUTF8PlainText as String: username]
+            [kUTTypeUTF8PlainText as String: stringRepresentation]
         ]
     }
 }
 
 class QRCodeController: UIViewController {
-    private var username: String!
+    private var username: String?
 
     static let addUsernameBasePath = "https://app.toshi.org/add/"
     static let addUserPath = "/add/"
@@ -68,7 +68,7 @@ class QRCodeController: UIViewController {
     }
 
     private lazy var qrCodeImageView: SelectableImageView = {
-        return SelectableImageView(username: "\(QRCodeController.addUsernameBasePath)\(username!)")
+        return SelectableImageView(stringRepresentation: "\(QRCodeController.addUsernameBasePath)\(username!)")
     }()
 
     private let subtitleLabel = TextLabel(Localized("profile_qr_code_subtitle"))
--- a/Toshi/Controllers/Settings/QRCodeController.swift
+++ b/Toshi/Controllers/Settings/QRCodeController.swift
@@ -16,8 +16,47 @@
 import UIKit
 import SweetUIKit
 import CoreImage
+import MobileCoreServices
+
+class SelectableImageView: UIImageView {
+    private var username: String
+
+    init(username: String) {
+        self.username = username
+
+        super.init(frame: .zero)
+
+        isUserInteractionEnabled = true
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override var canBecomeFirstResponder: Bool {
+        return true
+    }
+
+    override func becomeFirstResponder() -> Bool {
+        UIImpactFeedbackGenerator(style: .medium).impactOccurred()
+
+        return super.becomeFirstResponder()
+    }
+
+    override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        return action == #selector(copy(_:))
+    }
+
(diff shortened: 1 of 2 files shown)


Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
 import UIKit
 import SweetUIKit
 import CoreImage
+import MobileCoreServices
+
+class SelectableImageView: UIImageView {
+    private var username: String
```
yuliaveres: why a `username` property is something a `SelectableImageView` instance might own ? 🤔 
maybe it is better to have an `QRCodeImageView`? It might own `QRCodeIntent` which might be any type including username one ... ? wdyt ?

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
 import UIKit
 import SweetUIKit
 import CoreImage
+import MobileCoreServices
+
+class SelectableImageView: UIImageView {
```
yuliaveres: it is better to keep a class `final`

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
...
+    override func copy(_ sender: Any?) {
+        UIPasteboard.general.items = [
+            [kUTTypeJPEG as String: image!],
+            [kUTTypeUTF8PlainText as String: username]
+        ]
+    }
+}
 
 class QRCodeController: UIViewController {
+    private var username: String!
```
yuliaveres: we do not need `!`, do we ?

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -93,6 +139,16 @@ class QRCodeController: UIViewController {
 
         preferLargeTitleIfPossible(true)
     }
+
+    @objc private func handleLongPress(_ sender: UILongPressGestureRecognizer) {
+        guard sender.state == .began  else { return }
```
yuliaveres: `.began` or `.ended` ? 🤔

Elland: `.began` gives us a better user experience, similar to how tap to show menu works elsewhere in the OS.

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
 import UIKit
 import SweetUIKit
 import CoreImage
+import MobileCoreServices
+
+final class SelectableImageView: UIImageView {
+    private var stringRepresentation: String
```
roberthein: Why does a `SelectableImageView` has something called `stringRepresentation`?

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
 import UIKit
 import SweetUIKit
 import CoreImage
+import MobileCoreServices
+
+final class SelectableImageView: UIImageView {
```
roberthein: Every class should have it's own file.

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
...
+    private var stringRepresentation: String
+
+    init(stringRepresentation: String) {
+        self.stringRepresentation = stringRepresentation
```
roberthein: why is this being done before `init`?

Toshi/Controllers/Settings/QRCodeController.swift
roberthein: Why is this not a `UIControl`?

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
...
+        super.init(frame: .zero)
+
+        isUserInteractionEnabled = true
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override var canBecomeFirstResponder: Bool {
```
roberthein: This `FirstResponder` approach feels super hacky to me. This is not the way to this on iOS.

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
...
+
+        return super.becomeFirstResponder()
+    }
+
+    override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        return action == #selector(copy(_:))
+    }
+
+    override func copy(_ sender: Any?) {
+        UIPasteboard.general.items = [
```
roberthein: Why is it handling it's own action? Makes the whole thing unusable for anything else.

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -16,8 +16,47 @@
...
+            [kUTTypeJPEG as String: image!],
```
roberthein: We shouldn't explicitly unwrap anything.

Toshi/Controllers/Settings/QRCodeController.swift
```diff
@@ -28,13 +67,16 @@ class QRCodeController: UIViewController {
         return .default
     }
 
-    private lazy var qrCodeImageView: UIImageView = UIImageView()
+    private lazy var qrCodeImageView: SelectableImageView = {
+        return SelectableImageView(stringRepresentation: "\(QRCodeController.addUsernameBasePath)\(username!)")
```
roberthein: We shouldn't explicitly unwrap anything.

designatednerd:
Copy address functionality is being added as part of #723 - going to close this PR out since it will no longer be necessary

PR #585, opened 2017-11-21
Silence warnings about class delegate protocol 🤐

marijnschilling:
So i silenced the warnings about having to add `@objc` in front of class delegate protocols, because we can't ave them conform to objective C because of swift types being used in them.


marijnschilling:
Ellen already silenced these warnings in #594

PR #586, opened 2017-11-21
Fix tests failing locally, enable code coverage

designatednerd:
This PR updates the schemes for all versions of the app to enable Code Coverage so we can figure out what's covered and what isn't. 

I also fixed one crash at runtime in the tests (looks like from a bad merge) and one test which wasn't updated when the related code was updated.


designatednerd:
Looks like tests are puking on a code-signing issue on the CI box. Will try to figure out what's happening tomorrow.

PR #591, opened 2017-11-27
Pod Update

designatednerd:
This PR updates the Crashlytics, Fabric, and SwiftLint pods - C+F to deal with some real weird stack traces we're seeing, and SwiftLint to be on the latest version. 

Only changes to our code: 
- Updated Podfile to pin major + minor versions for Fabric and Crashlytics
- Removed unnecessary disabling of force-unwrap SwiftLint check.


Tests/IDAPIClientTests.swift
```diff
@@ -374,5 +373,3 @@ class IDAPIClientTests: QuickSpec {
         }
     }
 }
-
-//swiftlint:enable force_cast
```
yuliaveres: not needed 👍

designatednerd:
Yep, stack trace looks good from here.

designatednerd:
Per @yuliaveres 👍 I am mergin'

PR #592, opened 2017-11-27
Squish warnings

designatednerd:
This is a mishmash of stuff that barfed up a whole bunch of warnings - mostly Swift 4 string interpolation-related and adding a few things to help the compiler. Also threw in a test on something I wasn't 100% sure worked the way I thought it did in swift 4 (it now does!)

There's still a few warnings left but for now I'd like to get these changes in so we can actually notice new warnings.


Toshi/Supporting Files/AppDelegate.m
```diff
@@ -472,5 +472,12 @@ - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotif
     NSLog(@"Failed to register for remote notifications. %@", error);
 }
 
+#pragma mark - PKPushRegistryDelegate
+
+- (void)pushRegistry:(PKPushRegistry *)registry didUpdatePushCredentials:(PKPushCredentials *)pushCredentials forType:(PKPushType)type {
+    // TODO: Figure out what to do when these credentials get updated for VOIP
```
designatednerd: @roberthein @yuliaveres any idea what needs to go in here? this change shuts up a warning around VOIP credentials.

yuliaveres: we do not use VOIP now. we should not conform to PushRegistry

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -47,7 +47,7 @@ open class FavoritesController: SweetTableController, KeyboardAdjustable, Emptia
 
         let searchText = searchController.searchBar.text?.lowercased() ?? ""
 
-        let filteringBlock: YapDatabaseViewFilteringWithObjectBlock = { [weak self] transaction, group, colelction, key, object in
```
yuliaveres: why this is removed ? that is called on a property. we will have memory leak without weak here.

Toshi/Supporting Files/AppDelegate.m
designatednerd: Updating to remove that setup code - we should probably remove a bunch of the other code that references VOIP in another PR if we're not using it.

PR #593, opened 2017-11-27
Create db file in case of initialisation failure

yuliaveres:
It is better to create new db file if something went wrong and initialization failed. 
As the following is crash and re-login / re-install app. That means fresh database anyway.


Pods/SignalServiceKit/SignalServiceKit/src/Account/TSAccountManager.m
```diff
@@ -198,7 +199,9 @@ - (void)storeLocalNumber:(NSString *)localNumber
 
 + (uint32_t)getOrGenerateRegistrationId
 {
-    return [[self sharedInstance] getOrGenerateRegistrationId];
+    @synchronized (self) {
```
designatednerd: why does this need to be `@synchronized`?

Pods/SignalServiceKit/SignalServiceKit/src/Security/OWSHTTPSecurityPolicy.h
```diff
@@ -8,5 +8,7 @@
 @interface OWSHTTPSecurityPolicy : AFSecurityPolicy
 
 + (instancetype)sharedPolicy;
++ (void)setCertificateServiceName:(NSString *)serviceName;
 
 @end
+
```
designatednerd: excess newline

Pods/SignalServiceKit/SignalServiceKit/src/Account/TSAccountManager.m
yuliaveres: it used to be synchronized, was removed by mistake. that is not thread safe part 🤷‍♂️

Pods/SignalServiceKit/SignalServiceKit/src/Security/OWSHTTPSecurityPolicy.h
yuliaveres: 😬

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -293,7 +291,7 @@ - (BOOL)tryToLoadDatabase
     // from a corrupt keychain).
 
     YapDatabaseOptions *options = [[YapDatabaseOptions alloc] init];
-    options.corruptAction       = YapDatabaseCorruptAction_Fail;
+    options.corruptAction       = YapDatabaseCorruptAction_Rename;
```
designatednerd: The naming on this option is unclear - I had to look up the YapDB source code to figure out that it actually creates a new database.  Can you throw in a comment that indicates that this creates a new database rather than just renaming the file?

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -313,11 +311,11 @@ - (BOOL)tryToLoadDatabase
     _dbReadWriteConnection = self.newDatabaseConnection;
 
     YapDatabaseOptions *keysDBOptions = [[YapDatabaseOptions alloc] init];
-    keysDBOptions.corruptAction       = YapDatabaseCorruptAction_Fail;
+    keysDBOptions.corruptAction       = YapDatabaseCorruptAction_Rename;
 
     keysDBOptions.cipherKeyBlock = ^{
         typeof(self)strongSelf = weakSelf;
-        return [strongSelf keysDBPassword];
+        return [strongSelf databasePassword];
```
designatednerd: why does this need to change?

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
yuliaveres: @designatednerd  https://github.com/yapstudios/YapDatabase/blob/master/YapDatabase/YapDatabaseOptions.h#L44

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
yuliaveres: it is leftover code, we do not use separate password for those DBs, we can use same for chat and keys. (keysDBPassword which means different keychain key value was never used)

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
designatednerd: Yes I saw that, I'm asking you to put a note in our code so someone reading through it isn't like "wait, wtf?" 😄

yuliaveres: will add comment anyway 👍

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -290,6 +290,9 @@ - (BOOL)tryToLoadDatabase
     // this can be deleting any existing database file (if we're recovering
     // from a corrupt keychain).
 
+    // We use YapDatabaseCorruptAction_Rename option for database files which will create new one in case of failure
+    // documentation: https://github.com/yapstudios/YapDatabase/blob/master/YapDatabase/YapDatabaseOptions.h#L44
```
designatednerd: 👍 👍

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
designatednerd: [addressed]

Pods/SignalServiceKit/SignalServiceKit/src/Security/OWSHTTPSecurityPolicy.h
designatednerd: [addressed]

PR #594, opened 2017-11-27
Clean Up SwiftLint warnings

designatednerd:
New version of SwiftLint = New warnings to clean up. Whee!


Toshi/Models/Yap.swift
```diff
@@ -37,7 +37,7 @@ protocol Singleton: class {
     }
 }
 
-fileprivate struct UserDB {
+private struct UserDB {
```
yuliaveres: oh if speaking about `fileprivate` - we have lots !!! 😄

yuliaveres: can just do find->replace 😄

designatednerd: @yuliaveres done :)

PR #596, opened 2017-11-28
[IOS-391]: New group controller

yuliaveres:
screenshot at nov 28 09-24-36


Toshi/Controllers/GroupChat/GroupInfo/GroupInfoConfigurator.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        cell.subtitleLabel.text = cellData.subtitle
+        cell.detailsLabel.text = cellData.details
+        cell.leftImageView.image = cellData.leftImage
+        cell.switchControl.isOn = cellData.switchState == true
+
+        cell.leftImageView.layer.cornerRadius = 20
+        cell.leftImageView.layer.masksToBounds = true
+
+        guard let itemType = cellData.tag else { return }
```
designatednerd: Might be worth pulling this out as `GroupItemType` instead of the raw integer value so that a) You can switch directly on `.participant` etc below meaning b) You won't need a default case and c) you'll get warnings if another type is added which you need to handle.

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoConfigurator.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    override func configureCell(_ cell: UITableViewCell, with cellData: TableCellData) {
+        guard let cell = cell as? BasicTableViewCell else { return }
+
+        cell.titleTextField.text = cellData.title
+
+        cell.subtitleLabel.text = cellData.subtitle
+        cell.detailsLabel.text = cellData.details
+        cell.leftImageView.image = cellData.leftImage
+        cell.switchControl.isOn = cellData.switchState == true
```
designatednerd: for both readability and the poor dumb compiler, can you throw parens around `(cellData.switchState == true)` please?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+    private var groupModel: TSGroupModel
+    private var groupInfo: GroupInfo {
+        didSet {
+            setup()
+        }
+    }
+
+    init(_ groupModel: TSGroupModel) {
+        self.groupModel = groupModel
+        groupInfo = GroupInfo()
```
designatednerd: How is `GroupInfo` different from the `GroupModel`?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+
+    func adjustCompletionDelegate(_ delegate: GroupViewModelCompleteActionDelegate?) {
+        self.completionDelegate = delegate
+    }
+
+    var completeActionDelegate: GroupViewModelCompleteActionDelegate? {
+        get {
+            return completionDelegate
+        }
+        set { }
```
designatednerd: does this need an empty setter?

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+    
+    open override var canBecomeFirstResponder: Bool {
+        return true
+    }
+
+    open override func viewDidLoad() {
+        super.viewDidLoad()
+
+        if #available(iOS 11.0, *) {
+            self.navigationController?.navigationBar.prefersLargeTitles = false
```
designatednerd: isn't there some kind of extension for this?

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import SweetUIKit
+import SweetFoundation
+
+open class GroupViewController: UIViewController, KeyboardAdjustable, UINavigationControllerDelegate {
```
designatednerd: Why `open`?

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+        if #available(iOS 11.0, *) {
+            self.navigationController?.navigationBar.prefersLargeTitles = false
+        }
+
+        view.backgroundColor = Theme.lightGrayBackgroundColor
+        title = viewModel.viewControllerTitle
+
+        navigationItem.rightBarButtonItem = UIBarButtonItem(title: viewModel.rightBarButtonTitle, style: .plain, target:
+            viewModel, action: viewModel.rightBarButtonSelector)
+        navigationItem.rightBarButtonItem?.setTitleTextAttributes([.font: Theme.bold(size: 17.0), .foregroundColor: Theme.tintColor], for: .normal)
```
designatednerd: Is there a way to do this without hard-coding the font size - like do we have an enum of font sizes?

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+
+        addSubviewsAndConstraints()
+    }
+
+    open override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        if #available(iOS 11.0, *) {
+            navigationController?.navigationBar.prefersLargeTitles = false
```
designatednerd: same comment on the convenience method for this

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    func addSubviewsAndConstraints() {
+        view.addSubview(tableView)
+        view.addSubview(self.activityIndicator)
+
+        self.activityIndicator.set(height: 50.0)
+        self.activityIndicator.set(width: 50.0)
+        self.activityIndicator.centerXAnchor.constraint(equalTo: view.centerXAnchor).isActive = true
+        self.activityIndicator.centerYAnchor.constraint(equalTo: view.centerYAnchor).isActive = true
```
designatednerd: probably my fault for bring my `self`-ish ways to the team, but unnecessary `self` on all these

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+        return activityIndicator
+    }()
+
+    private func adjustDoneButton() {
+        navigationItem.rightBarButtonItem?.isEnabled = viewModel.isDoneButtonEnabled
+    }
+
+    private func showUserInfo(with userId: String) {
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
+        let users = appDelegate.contactsManager.tokenContacts.filter { $0.address == userId }
```
designatednerd: Is there a better way to get this than going to the app delegate?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: i use group info as a placeholder set of values in the process of creation/editing. it has more information we need to handle UI and user input, like switches state changes etc. in the end, when we hit Update /  Create - we pull all we need from `GroupInfo` and create TSGroupModel signal object which has only avatar / name, id and membersIds properties , with needed values and send relevant message with it.

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: nope but i cant omit it, can i ?

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: 🤔  not i am aware of ...

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: mistake

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: i pulled that piece from already existing controller to be consistent with titles style. it is used in this way in some places already, lets change that in separate pr?

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: not now 😞  i have a branch which refactors and pulls all those manages out of `AppDelegate`, but due to time limits i have not finished it yet. but we should address it next thing for sure !

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,361 @@
...
+    func didChangeSwitchState(_ cell: BasicTableViewCell, _ state: Bool) {
+        guard let indexPath = tableView.indexPath(for: cell) else { return }
+
+        let sectionData = viewModel.sectionModels[indexPath.section]
+        let cellData = sectionData.cellsData[indexPath.row]
+
+        guard let itemType = cellData.tag else { return }
+
+        switch itemType {
+        case GroupItemType.notifications.rawValue:
```
designatednerd: Even though you'd still need a `default` this seems like a good place to also use `GroupItemType` when you pull out the tag

Toshi/Controllers/GroupChat/GroupViewModelProtocol.swift
```diff
@@ -0,0 +1,61 @@
...
+    var title: String = ""
+    var avatar = UIImage(named: "avatar-edit")!
+    var isPublic = false
+    var notificationsOn = true
+    var participantsIDs: [String] = []
+}
+
+protocol GroupViewModelCompleteActionDelegate: class {
+
+    func groupViewModelDidFinish()
```
designatednerd: what did the group view model finish when this fires?

Toshi/Controllers/GroupChat/NewGroup/NewGroupConfigurator.swift
```diff
@@ -0,0 +1,46 @@
...
+
+import Foundation
+import UIKit
+
+final class NewGroupConfigurator: CellConfigurator {
+
+    override func configureCell(_ cell: UITableViewCell, with cellData: TableCellData) {
+        guard let cell = cell as? BasicTableViewCell else { return }
+
+        cell.titleTextField.placeholder = "New group"
```
designatednerd: 🚨 non-localized 🚨

Toshi/Controllers/GroupChat/NewGroup/NewGroupConfigurator.swift
```diff
@@ -0,0 +1,46 @@
...
+        cell.leftImageView.image = cellData.leftImage
+        cell.switchControl.isOn = cellData.switchState == true
+
+        cell.leftImageView.layer.cornerRadius = 20
+        cell.leftImageView.layer.masksToBounds = true
+
+        guard let itemType = cellData.tag else { return }
+
+        switch itemType {
+        case GroupItemType.addParticipant.rawValue:
```
designatednerd: same deal on `GroupItemType` here

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
```diff
@@ -0,0 +1,133 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+import UIKit
+
+final class NewGroupViewModel {
```
designatednerd: Why are `NewGroupViewModel` and `GroupViewModel` separate objects?

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: the thing is - we need to have tag as Int. it would be much better to have it as an enum value not raw value, but remember that tag is used in the generic cell and tag is very contextual. each controller / viewModel has its own set of tags / cell types which are processed on higher abstract level on CellTableData. We cant have enums hierarchy unfortunately, thus we deal with Int-s . ..

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        let participantsHeaderTitle = String(format: Localized("group_participants_header_title"), members.count)
+        let participantsSectionData = TableSectionData(cellsData: participantsCellData, headerTitle: participantsHeaderTitle)
+
+        models = [avatarTitleSectionData, settingsSectionData, participantsSectionData]
+    }
+
+    @objc private func createChat() {
+        let membersIds =  NSMutableArray(array: groupInfo.participantsIDs)
```
designatednerd: You could also do `var membersIDs = groupInfo.participantsIDs` and then you can `append` the next thing.

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: so 1 - we need to extract not just contactcManager but all of those from appDelegate (which were there from the very beginning 😞 )
2 - when we do so - it will include lots of rector throughout the app.

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    func updateAvatar(_ image: UIImage) {
+        groupInfo.avatar = image
+    }
+
+    func updateTitle(_ title: String) {
+        groupInfo.title = title
+    }
+
+    func updatePublicState(_ isPublic: Bool) {
```
designatednerd: maybe `to` instead of `_` here and below? otherwise it could read as something which would be performing these updates under the hood or not rather than setting the state.

Toshi/Controllers/Profiles/ProfilesDataSource.swift
```diff
@@ -2,68 +2,233 @@ import Foundation
...
+
+        return dbConnection
+    }()
+
+    private lazy var customFilterDatabaseConnection: YapDatabaseConnection = {
+        let database = Yap.sharedInstance.database
+        let dbConnection = database!.newConnection()
+
+        return dbConnection
+    }()
```
designatednerd: why pull these into separate connections?

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
yuliaveres: why is that better ?

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
yuliaveres: because they serve for different cases. datasource, view setup differ per NewGroup / Group Info. Keeping them separate is clear.

Toshi/Controllers/GroupChat/NewGroup/NewGroupConfigurator.swift
yuliaveres: dnt understand what do you mean.

Toshi/Controllers/Profiles/ProfilesDataSource.swift
```diff
@@ -2,68 +2,233 @@ import Foundation
...
+                return contact1.username.compare(contact2.username)
+            }
+
+            return .orderedAscending
+        }
+
+        return viewSorting
+    }
+
+    private func seltupAllProfilesCollection() {
```
designatednerd: excess `L` in `seltup`

Toshi/Controllers/GroupChat/GroupViewModelProtocol.swift
yuliaveres: updating or creating a group

Toshi/Controllers/Profiles/ProfilesDataSource.swift
```diff
@@ -73,40 +238,47 @@ class ProfilesDataSource: NSObject {
...
-    func reloadData() {
-        if #available(iOS 11.0, *) {
-            // Must perform batch updates on iOS 11 or you'll get super-wonky layout because of the headers.
-            tableView?.performBatchUpdates({
-                self.tableView?.reloadData()
-            }, completion: nil)
-        } else {
-            tableView?.reloadData()
+
+    func rightBatButtonEnabled() -> Bool {
```
designatednerd: 🦇  (I think you mean 📊 )

Toshi/Controllers/Profiles/ProfilesDataSource.swift
yuliaveres: those are different, one is for search filter, another is for optional filter, u can use id or not to filter out list if needed (one case is updating group, we filter out group members).
updating filter works with different connections

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -158,6 +158,79 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
+        guard let profile = dataSource.profile(at: indexPath) else {
+            assertionFailure("Could not get profile at indexPath: \(indexPath)")
+            return cell
+        }
+
+        cell.avatarPath = profile.avatarPath
+        cell.name = profile.name
+        cell.displayUsername = profile.displayUsername
+
+        if type == .newGroupChat || type == .updateGroupChat {
```
designatednerd: maybe use the computed var below?

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -298,53 +309,37 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
-        let alert = UIAlertController(title: "Coming soon!", message: "Group chat is under development. Soon you can talk to:\n\(usernames)", preferredStyle: .alert)
-        let ok = UIAlertAction(title: "OK", style: .cancel, handler: nil)
-        alert.addAction(ok)
-        self.present(alert, animated: true)
-    }
-}
 
-// MARK: - Profile Selection Delegate
+        let membersIdsArray = dataSource.selectedProfiles.sorted { $0.username < $1.username }.map { $0.address }
+        let membersIds = NSMutableArray(array: membersIdsArray)
```
designatednerd: same question on using NSMutableArray instead of a `var` here

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -298,53 +309,37 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
+
+    func reloadData() {
+        if #available(iOS 11.0, *) {
+            // Must perform batch updates on iOS 11 or you'll get super-wonky layout because of the headers.
+            tableView?.performBatchUpdates({
+                self.tableView?.reloadData()
+            }, completion: nil)
         } else {
-            doneButton.isEnabled = false
+            tableView?.reloadData()
```
designatednerd: out of curiosity, why move this out of the data source?

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: signal need to accept `NSMutableArray`

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: why datasource should work with UI ? UIViewController handles everything UITableView related related.

Toshi/Resources/Localizable.strings
```diff
@@ -253,6 +254,18 @@
...
 "error_title" = "Error";
+
+/* Group chat */
+"new_group_title" = "New Group";
+"group_info_title" = "Group Info";
+"create_group_button_title" = "Create";
+"update_group_button_title" = "Update";
+"new_group_settings_header_title" = "Group settings";
+"new_group_add_participants_action_title" = "Add Participants";
+"group_participants_header_title" = "%d Participants";
```
designatednerd: We should probably have a `Plurals.stringsdict` to handle pluralization for zero/one/many cases. https://www.raizlabs.com/dev/2017/03/localized-pluralization-with-stringsdict/ 

I'm cool if you want to put that in as a separate story for later, but it's definitely easier to do it at this point than later.

Toshi/Views/CellConfigurator/ToshiCellConfigurator.swift
```diff
@@ -0,0 +1,65 @@
...
+
+import UIKit
+
+protocol ToshiCellConfigurator: class {
+
+    func configureCell(_ cell: UITableViewCell, with cellData: TableCellData)
+    func cellIdentifier(for components: TableCellDataComponents) -> String
+}
+
+class CellConfigurator: ToshiCellConfigurator { // lets say we have this configurator responsible for
```
designatednerd: responsible for what?

Toshi/Views/Cells/ToshiAvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+final class AvatarTitleSubtitleCell: ToshiTableViewCell {
+
+    override func addSubviewsAndConstraints() {
+        leftImageView = UIImageView(frame: .zero)
```
designatednerd: might be good to do a `let leftImage = UIImageView(frame: .zero)` then set up all the constraints and then at the end do `leftImageView = leftImage` so that you don't need to have so many optionals during setup

Toshi/Views/CellConfigurator/ToshiCellConfigurator.swift
yuliaveres: i think it is not there anymore ? are u reviewing commit by commit? that is outdated. sorry

Toshi/Views/Cells/ToshiAvatarTitleSubtitleCell.swift
yuliaveres: i think this PR was approved and closed, related to unified cells and setup. ..

yuliaveres: not sure if i should make these kind of changes here ...

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -154,4 +155,8 @@ extension BasicTableViewCell: UITextFieldDelegate {
     func textFieldDidEndEditing(_ textField: UITextField) {
         self.actionDelegate?.didFinishTitleInput(self, text: textField.text)
     }
+
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
```
designatednerd: are these still necessary with the extension on `ToshiTableViewCell`?

yuliaveres: yes

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
designatednerd: if you just want this to be a computed property you can get rid of `get` and then you won't need `set`

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
designatednerd: OK, thank you for the clarification

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: oh right ! yes we do !

designatednerd: `UIViewController+NavigationBar` -> `preferLargeTitleIfPossible`

Toshi/Controllers/GroupChat/GroupViewController.swift
designatednerd: Sounds like a plan

Toshi/Controllers/GroupChat/GroupViewController.swift
designatednerd: Yeah it would probably be best to do this piece by piece so we can get little bits off the delegate as we need them

Toshi/Controllers/GroupChat/GroupViewController.swift
designatednerd: Right but you can do `GroupItemType(rawValue: cellData.tag)` and then that'll still allow you to have the int as the tag, but do the switch more cleanly

Toshi/Controllers/GroupChat/GroupViewModelProtocol.swift
designatednerd: I'd say either a) Update the name to indicate that or b) Add a docu-comment.

Toshi/Controllers/GroupChat/NewGroup/NewGroupConfigurator.swift
designatednerd: Pulling out `GroupItemType(rawValue: cellData.tag)` instead of just the tag itself

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
designatednerd: Is it something where it could be similar data but with a `newGroup` vs `editGroup` enum? Or is the data too different?

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
designatednerd: because it's using all swift stuff and not having to bridge to ObjC unnecessarily

Toshi/Controllers/Profiles/ProfilesDataSource.swift
designatednerd: ok

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: ahhhh that would explain the thing above too - I can't remember if you can do implicit conversion or not but my guess is no

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
yuliaveres: i really would not prefer this as we will end up with massive class :/
this is clearer and more flexible in term of introducing changes

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: also good god signal whyyyyyy it should be a copy of whatever people have been handing you

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
yuliaveres: table data is different yes

Toshi/Resources/Localizable.strings
yuliaveres: as a perfect case we should not end up with 1 member group (only if someone leaves 😢   )

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: Yes and no - the data source object handles everything related to setting up the UI of the tableview itself. I'd actually been thinking about handling the show/hide of the empty view to it since that's where all that information lives. 

Basically I tend to find that putting the data source object in charge of all bullshit related to the tableview makes the UIVC a lot cleaner.

Toshi/Views/CellConfigurator/ToshiCellConfigurator.swift
designatednerd: Nope looking at the big long thing in GitHub under "files changed"

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: is that too odd to retrieve all we need for setting up table view, an UI object ? It is not "not-clean" is it ?.

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: i will look carefully to convert it at final step only 👍

Toshi/Views/Cells/ToshiAvatarTitleSubtitleCell.swift
designatednerd: I feel like you've got the changes from master though - so I'm not sure why this would still be here.

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: i didnt know it is one of ways you prefer handling it, so sorry for changing it @designatednerd.
lets try that approach next time ?

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
designatednerd: Fair enough

Toshi/Resources/Localizable.strings
designatednerd: ah yeah good point

Toshi/Views/CellConfigurator/ToshiCellConfigurator.swift
designatednerd: Doesn't look like this got removed - this is still in the 2000+ lines of changes

Toshi/Views/Cells/ToshiAvatarTitleSubtitleCell.swift
yuliaveres: i am not sure why github shows this diff, looks like old stuff. i dnt have it locally. i have all cells-related things from `feature/group-chat`

designatednerd:
OK, through all this but in the future, I would suggest breaking some of this stuff up into smaller PRs into the `feature/group-chat` branch - there's a lot of different things happening here that need to happen for group chat, but it's difficult to review all of it at once.

Toshi/Views/Cells/ToshiAvatarTitleSubtitleCell.swift
designatednerd: I'm not sure what to tell you - maybe try merging it one more time?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: but that is protocol variable, it must be whether `{ get }` or `{ get set }`

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: so i checked, it gets converted to `NSMutableArray` only in two places just before creating or updating a group in both view models (`createChat()` and `updateGroup()`)

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: eh, fine by me to punt for next time. And it's basically a question of "how dumb do we want the view controller to be?"

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: Mind just doing it in the very last step just to make it clear that you're only doing it to make Signal's API happy?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
designatednerd: Does the protocol need to be `{ get set }` or just `{ get }`

Toshi/Controllers/Profiles/ProfilesViewController.swift
Elland: Sounds like a great place to add a comment explaining intent. 👍

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: it can be just { get } but i also set it ...

designatednerd: if you set it then you probably ought to pass it through to this underlying property

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: it is done in the very last step, just before creating `TSGroupModel` . the very last i think

yuliaveres: screenshot at dec 11 16-18-14

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: not sure i follow.. what should be in set block ?

designatednerd: Since this is being passed through to another variable from the getter, it also ought to be on the setter

```swift 
set {
    completionDelegate = newValue
}
```

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: Right what I'm suggesting is not pulling that out as a separate var where the first arrow is and only using `NSMutableArray` where you're creating the `TSGroupModel` where the second arrow is - makes it clearer that you're using an `NSMutableArray` because the `TSGroupModel` API requires it

yuliaveres: ahhhaaaa

Toshi.xcodeproj/project.pbxproj
```diff
@@ -51,9 +51,9 @@
...
 		2B0EF82B1F948931001A76CA /* CrashlyticsClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B0EF8281F948931001A76CA /* CrashlyticsClient.swift */; };
-		2B1D66961FC2E00E00EB166C /* BasicTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B1D66951FC2E00E00EB166C /* BasicTableViewCell.swift */; };
-		2B1D66971FC2E00E00EB166C /* BasicTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B1D66951FC2E00E00EB166C /* BasicTableViewCell.swift */; };
-		2B1D66981FC2E00E00EB166C /* BasicTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B1D66951FC2E00E00EB166C /* BasicTableViewCell.swift */; };
+		2B1D66961FC2E00E00EB166C /* ToshiTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B1D66951FC2E00E00EB166C /* ToshiTableViewCell.swift */; };
+		2B1D66971FC2E00E00EB166C /* ToshiTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B1D66951FC2E00E00EB166C /* ToshiTableViewCell.swift */; };
+		2B1D66981FC2E00E00EB166C /* ToshiTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B1D66951FC2E00E00EB166C /* ToshiTableViewCell.swift */; };
```
designatednerd: Does this need to be put back after the files you removed are gone?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoConfigurator.swift
```diff
@@ -0,0 +1,48 @@
...
+        cell.leftImageView.image = cellData.leftImage
+        cell.switchControl.isOn = (cellData.switchState == true)
+
+        cell.leftImageView.layer.cornerRadius = 20
+        cell.leftImageView.layer.masksToBounds = true
+
+        guard let tag = cellData.tag, let itemType = GroupItemType(rawValue: tag) else { return }
+
+        switch itemType {
+        case GroupItemType.participant:
```
designatednerd: you should be able to just use `.participant` etc here

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,354 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import SweetUIKit
+import SweetFoundation
+
+final class GroupViewController: UIViewController, KeyboardAdjustable, UINavigationControllerDelegate {
```
designatednerd: is it possible to pull the bits which are conforming to the `KeyboardAdjustable` and `UINavigationControllerDelegate` protocols out into individual extensions?

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -0,0 +1,354 @@
...
+    public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
+        let sectionData = viewModel.sectionModels[indexPath.section]
+        let cellData = sectionData.cellsData[indexPath.row]
+
+        guard let tag = cellData.tag, let itemType = GroupItemType(rawValue: tag) else { return }
+
+        switch itemType {
+        case GroupItemType.participant:
```
designatednerd: this can also be just `.participant` etc

Toshi/Controllers/GroupChat/NewGroup/NewGroupConfigurator.swift
```diff
@@ -0,0 +1,46 @@
...
+        cell.leftImageView.image = cellData.leftImage
+        cell.switchControl.isOn = (cellData.switchState == true)
+
+        cell.leftImageView.layer.cornerRadius = 20
+        cell.leftImageView.layer.masksToBounds = true
+
+        guard let tag = cellData.tag, let itemType = GroupItemType(rawValue: tag) else { return }
+
+        switch itemType {
+        case GroupItemType.addParticipant:
```
designatednerd: same with dropping to `.addParticipant` here

Toshi/Controllers/Profiles/ProfilesDataSource.swift
```diff
@@ -2,68 +2,233 @@ import Foundation
...
     
     var isEmpty: Bool {
         let currentItemCount = mappings.numberOfItems(inSection: 0)
-        return (currentItemCount > 0)
+        return currentItemCount == 0
+    }
+
+    var searchText: String = "" {
+        didSet {
+            searchDatabaseConnection.readWrite { [weak self] transaction in
```
designatednerd: why does this database connection need to be `readWrite`?

Toshi/Controllers/Profiles/ProfilesDataSource.swift
```diff
@@ -2,68 +2,233 @@ import Foundation
...
+            let tag = Date().timeIntervalSinceReferenceDate
+            filterTransaction.setFiltering(strongSelf.customFiltering, versionTag: String(describing: tag))
+        }
+    }
+
+    @discardableResult
+    private func registerTokenContactsDatabaseView() -> Bool {
+        guard let database = Yap.sharedInstance.database else { fatalError("couldn't instantiate the database") }
+        // Check if it's already registered.
+        //guard database.registeredExtension(TokenUser.viewExtensionName) == nil else { return true }
```
designatednerd: can this commented out code be removed?

Toshi/Controllers/Profiles/ProfilesDataSource.swift
```diff
@@ -73,40 +238,47 @@ class ProfilesDataSource: NSObject {
...
-        if #available(iOS 11.0, *) {
-            // Must perform batch updates on iOS 11 or you'll get super-wonky layout because of the headers.
-            tableView?.performBatchUpdates({
-                self.tableView?.reloadData()
-            }, completion: nil)
-        } else {
-            tableView?.reloadData()
+
+    func rightBarButtonEnabled() -> Bool {
+        let isSingleSelectionSetup = (type == .newChat || type == .favorites || type == .updateGroupChat)
```
designatednerd: this seems like it might be a good place to use a `switch` statement instead of a giant boolean

Toshi/Controllers/Profiles/ProfilesDataSource.swift
yuliaveres: because u set filtering on read write transaction which modifies results

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -30,49 +31,37 @@ public enum ProfilesViewControllerType {
...
+
 // MARK: - Profiles View Controller
 
 final class ProfilesViewController: UITableViewController, Emptiable {
     
     let type: ProfilesViewControllerType
-    
+
     var scrollViewBottomInset: CGFloat = 0
+    private(set) var output: ProfilesListCompletionOutput?
```
designatednerd: should this be `weak`?

yuliaveres: no since controller is not kept by presenting one

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -158,6 +162,79 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
+            didSelectProfile(profile: profile)
+        case .newGroupChat, .updateGroupChat:
+            dataSource.updateSelection(at: indexPath)
+            updateHeaderWithSelections()
+            reloadData()
+            navigationItem.rightBarButtonItem?.isEnabled = dataSource.rightBarButtonEnabled()
+        }
+    }
+
+    public override func numberOfSections(in tableView: UITableView) -> Int {
```
designatednerd: wait, I think I missed this in the first round - why is this now implementing UITableViewDelegate and UITableViewDataSource again rather than just letting the data source handle this?

Toshi/Controllers/Profiles/ProfilesDataSource.swift
yuliaveres: as u wish

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: (goes for all the UITVDS methods here)

yuliaveres: yes, and it retrieves everything needed from datasource to handle UITableView related part

yuliaveres: as i mentioned controller is tableView delegate and datasource

Toshi/Controllers/Profiles/ProfilesDataSource.swift
designatednerd: but it shouldn't actually *write* anything to the database, correct? does simply modifying the results require that?

Toshi/Controllers/Profiles/ProfilesDataSource.swift
designatednerd: Does it make sense why I think that's better?

Toshi/Controllers/Profiles/ProfilesDataSource.swift
yuliaveres: it does, as all filter changes are considered as this view changes (db change)

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: not in configuring the cells - that's just the same logic as was in the data source

yuliaveres: smth wrong with this ?

Toshi/Controllers/Profiles/ProfilesDataSource.swift
designatednerd: But the underlying contents of the database don't actually change, right? (Just trying to wrap my head around Yap's reasoning here)

Toshi/Controllers/Profiles/ProfilesViewController.swift
yuliaveres: i just dont get what is the issue . could you explain ?

designatednerd: The basic idea of a separate `DataSource` object is that you hand it the table view and a little bit of information about what it's supposed to display, and it (the data source Object) becomes the delegate/data source for the table view, then handles all configuration related to the table view itself. This leaves the view controller free to do whatever the hell it needs to. 

This allows a bunch of things: 
a) Not having UITableViewDataSource/Delegate gunk in the view controller
b) Reusable table views for situations where the same data needs to be displayed in multiple places
c) Ability to compose an existing table view and other things that go with that table view into a new controller without having to subclass. 

If you think it'd be better to call it something other than Data Source that's cool, but basically the idea is to isolate all the logic around the table view outside of the view controller so that it can be reused like a lego block.

designatednerd: Perhaps we can punt on this discussion for now but does that at least make more sense in terms of why I was confused about pulling all of this back out of the data source?

yuliaveres: sure, i see your point. what would be nice in that case is separate database related responsibility and UI setup related responsibility between different classes ... i guess data fetch should be handled separately, wdyt ?

yuliaveres: then we might want to have datasource / or configurator or whatever which does tableview setup , and actually view model or smth which does work with DB

Toshi/Controllers/Profiles/ProfilesDataSource.swift
yuliaveres: not underlying , but filtered view content changes. it is a child of main view and its content changes. main view's content stays same of course

Toshi/Controllers/Profiles/ProfilesDataSource.swift
yuliaveres: better readability?

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: In theory you could set something up that takes an already set up database connection, but in general I find these types of object to be useful for setting up the bindings between the database and the UI. which requires both database connections and UI code. 😄

Toshi/Controllers/Profiles/ProfilesDataSource.swift
designatednerd: Partly - it also allows you to isolate the return values in the statement below, and it makes it easier to see exactly which cases are handled in a given way.

Toshi/Controllers/GroupChat/GroupViewController.swift
yuliaveres: no way to extract `NavigationControllerDelegate` thought, as we don't use any of its methods actually, it is needed for `UIImagePickerController` instead.

designatednerd: Maybe move the conformance down with that extension, then?

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -273,6 +264,18 @@ extension GroupViewController: UITableViewDelegate {
...
+    private func presentImagePicker(sourceType: UIImagePickerControllerSourceType) {
+        let imagePicker = UIImagePickerController()
+        imagePicker.sourceType = sourceType
+        imagePicker.allowsEditing = true
+        imagePicker.delegate = self
+
+        present(imagePicker, animated: true)
+    }
+}
+
```
designatednerd: This works too but what I meant was move the `UINavigationControllerDelegate` conformance bit so it's down where `extension GroupViewController: UIImagePickerControllerDelegate` bit was. But I suppose this works to since it's isolating the bit where we're actually using this vc as a UINavigationController delegate.

PR #597, opened 2017-11-28
Whittle down long typecheck warnings

designatednerd:
So the main thing happening here is that I fixed a whole lot of long type check warnings. Fun fact: This is super time-consuming because they don't always update unless you super-clean (ie, clean build folder). 

The other thing is that I set our default timeout to 200ms - this is what it was on Debug, and all other targets were on way lower. If we're gonna ignore things, let's ignore them consistently so that our warnings are the same on all targets (and ideally, are gone).


Toshi/Models/Cereal.swift
```diff
@@ -43,6 +43,25 @@ public class Cereal: NSObject {
...
+        return mnemonic.keychain
+            .derivedKeychain(at: 0, hardened: true)
+            .derivedKeychain(at: 1)
+            .derivedKeychain(at: 0)
+    }
+    
+    private static func walletKeychain(from mnemonic: BTCMnemonic) -> BTCKeychain {
+        // wallet path: 44H/60H/0H/0 and then 0 again. Metamask root path, first key.
+        // Metamask allows multiple addresses, by incrementing the last path. So second key would be: 44H/60H/0H/0/1 and so on.
+        return mnemonic.keychain
```
yuliaveres: nice!

Toshi/Controllers/Sign In/Views/PaymentRequestInfoView.xib
```diff
@@ -28,7 +28,7 @@
                         <constraint firstAttribute="height" constant="44" id="l68-40-FNK"/>
                     </constraints>
                 </imageView>
-                <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" horizontalCompressionResistancePriority="1000" text="" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontForContentSizeCategory="YES" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="deq-gQ-KiS">
```
yuliaveres: can't get neither the purpose of `adjustsFontForContentSizeCategory ` being removed, nor commit message :/

Toshi/Controllers/Settings/Models/ProfileEditConfigurator.swift
```diff
@@ -29,8 +29,8 @@ final class ProfileEditConfigurator {
         cell.textField.text = item.detailText
         cell.switchControl.isOn = item.switchMode
 
-        cell.switchControl.isHidden = (item.type != .visibility)
-        cell.textField.isHidden = (item.type == .visibility)
+        cell.switchControl.isHidden = (item.type != ProfileEditItemType.visibility)
```
yuliaveres: i think we tended to avoid it previously.. ? 🤔  not sure if we had a guideline for it, but we preferred `.visibility`. 
Is `ProfileEditItemType.visibility` better in your point of view ?

Toshi/Controllers/Browse/Views/ReputationView.swift
```diff
@@ -65,25 +65,27 @@ final class ReputationView: UIView {
...
+        firstGuide.left(to: self)
+        firstGuide.bottom(to: self)
         
-        guides[1].top(to: self)
-        guides[1].leftToRight(of: guides[0])
-        guides[1].bottom(to: self)
-        guides[1].right(to: self)
+        let secondGuide = guides[1]
+        secondGuide.top(to: self)
+        secondGuide.leftToRight(of: firstGuide)
```
yuliaveres: 🌟

Toshi/Views/ChatCell.swift
```diff
@@ -21,43 +21,9 @@ import SweetFoundation
...
-                    unreadView.isHidden = false
-                    lastMessageDateLabel.textColor = Theme.tintColor
-                } else {
-                    unreadLabel.text = nil
-                    unreadView.isHidden = true
-                    lastMessageDateLabel.textColor = Theme.greyTextColor
-                }
-            }
+            self.configureLastVisibleMessage()
+            self.configureLastMessageDate()
```
yuliaveres: also we tend no using explicit `self.` .. do u have thoughts on this ?

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -49,17 +49,20 @@ class SOFAWebController: UIViewController {
...
         var userScript: WKUserScript = WKUserScript(source: js, injectionTime: .atDocumentStart, forMainFrameOnly: false)
-
+        
         configuration.userContentController.add(self, name: Method.getAccounts.rawValue)
         configuration.userContentController.add(self, name: Method.signTransaction.rawValue)
         configuration.userContentController.add(self, name: Method.publishTransaction.rawValue)
         configuration.userContentController.add(self, name: Method.approveTransaction.rawValue)
-
+        
         configuration.userContentController.addUserScript(userScript)
```
yuliaveres: can we use newline before `return`s ?

Toshi/Models/Notifications/EthereumNotificationHandler.swift
```diff
@@ -48,15 +48,15 @@ class EthereumNotificationHandler: NSObject {
...
-                completion(.newData)
-            }
-
             guard let sofa = SofaWrapper.wrapper(content: body) as? SofaPayment else {
                 completion(.noData)
 
                 return
             }
+            
+            defer {
```
yuliaveres: perfect 😄

Toshi/Views/ChatCell.swift
designatednerd: I personally prefer to use `self` because it's easier to tell during code review if you're calling a local variable or a property, but a) That is a WAY longer discussion and b) good catch on this being out of the code standards.

Toshi/Controllers/Sign In/Views/PaymentRequestInfoView.xib
designatednerd: There was a warning about this being enabled on a text view set up to use auto-shrinking font. This is more for Dynamic Type, auto-shrinking just makes the font smaller if there's no room.

Toshi/Controllers/Settings/Models/ProfileEditConfigurator.swift
designatednerd: So the main reason I added these is to give the compiler more of a clue about what's here - this method was taking over 200ms to compile before I gave it that hint.

Toshi/SOFAWebview/SOFAWebController.swift
designatednerd: yeah sorry that was reflexive cleanup on my part. putting back..

Toshi/Controllers/ScannerController.swift
```diff
@@ -93,30 +93,38 @@ extension ScannerController: PaymentPresentable {
...
-                self?.presentPaymentError(withErrorMessage: error?.localizedDescription ?? ToshiError.genericError.description)
+            self?.sendSignedTransaction(signedTransaction, originalTransaction: transaction)
+        }
+    }
+    
+    private func createSignedTransaction(from unsignedTransaction: String) -> String {
+        return "0x\(Cereal.shared.signWithWallet(hex: unsignedTransaction))"
+    }
+    
+    private func sendSignedTransaction(_ signedTransaction: String, originalTransaction: String) {
```
yuliaveres: would be nice to keep same order of signed/original transactions parameters as in `EthereumAPIClient.shared.sendSignedTransaction` just for readability purpose :)

yuliaveres:
Oh also could we name commits in a bit more descriptive way.
f.e. instead of "there should be newlines before return statements" we tend use "Add newlines before return statements"

designatednerd:
Sure - I was just phrasing that as if it was a command in our code standards. 

SLGTY (still looks good to you)?

yuliaveres:
What exactly does still look good to me ? 😄

designatednerd:
The code I changed after you originally approved it :)

yuliaveres:
Yup 👍

PR #598, opened 2017-11-29
Dependency organization/Adding log utility

designatednerd:
Depends on #597 - Please do not review until that one is merged!

Did a bunch of organization and clarification on our dependency management in both Carthage and CocoaPods.

I was hoping to be able to get rid of our dependency on CocoaLumberjack but alas, not so. In any case, I left the logging utility I wrote up while still hopeful in there so we can use it elsewhere. I'll hunt NSLogs in the project in a separate pull.


designatednerd:
OK this is now ready for review - It's being obnoxious about not removing the old commits because of squashing, but the file diff is the actual difference. I'll rebase next time.

Cartfile
```diff
@@ -1,17 +1,25 @@
...
+
+# General
+
 github "Anakros/WebRTC"
-github "UseSweet/SweetFoundation" ~> 1.2.0
-github "UseSweet/SweetSwift" "master"
-github "UseSweet/SweetUIKit" "master"
+github "ashleymills/Reachability.swift" "a17b2882db4803aaea08d7eb1c894304d744be64"
 github "bakkenbaeck/CameraScanner" "01708dc3f1b440177173b48377ca18f6748fa58e"
-github "bakkenbaeck/EtherealCereal" ~> 0.1
```
marijnschilling: We dont use this framework? Isnt this the framework that sets up public private key stuff? Or do we just have it inside the project now?

Podfile
```diff
@@ -3,13 +3,20 @@ platform :ios, '10.0'
 inhibit_all_warnings!
 
 abstract_target 'CocoaPods' do
+  # Things which can't be installed via Carthage because LOL C
```
marijnschilling: I would suggest `# Things which can't be installed via Carthage because of C` I know its more boring 🙈  but its also clearer 😬

Podfile
```diff
@@ -3,13 +3,20 @@ platform :ios, '10.0'
...
   pod 'AxolotlKit',  git: 'https://github.com/WhisperSystems/SignalProtocolKit.git'
   pod 'SignalServiceKit', git: 'https://github.com/toshiapp/Signal-iOS.git', commit: "ff7b1ec7feb876aa7c5cdb34d8fbacf86821c3cd"
+  
+  # Grabs from the head of this repo, which allows SignalServiceKit to actually compile
```
marijnschilling: Same here but for: `# Grabs from the head of this repo, which allows SignalServiceKit to compile`

Cartfile
designatednerd: This is now in the `# Ethereum` section

designatednerd:
Addressed your suggestions - yes, I do intend for `DLog` to be used instead of `NSLog`/`print` statements so that all the log barf gets turned off when we're not debugging, but that's gonna be a separate PR. 

I mostly made it to try and replace CocoaLumberjack's functionality of setting a particular log level, but it turns out *we* aren't actually using that, it's used by some of the stuff we pulled out of SignalServiceKit and altered.

marijnschilling:
Yes sounds good! And what about the ethereal cereal? Don't we use that?

designatednerd:
Yes - I replied to your comment, it is now in a different section of the Cartfile.

PR #599, opened 2017-11-29
Add Earl Grey via Carthage for future UI testing

designatednerd:
Depends on #598 - please do not review until after that has been merged. 

Doing this as a separate PR so the review is not so bad - Only changes to our actual code should be at the bottom in the `.pbxproj`.


designatednerd:
yah! gonna start using it for some group chat stuff i'm going to work on, so hopefully that will help

PR #600, opened 2017-11-29
Remove references to VOIP

designatednerd:
...since we're not actually using it.


Pods/SignalServiceKit/SignalServiceKit/src/Account/TSAccountManager.h
```diff
@@ -103,10 +103,9 @@ extern NSString *const kNSNotificationName_LocalNumberDidChange;
  *  @param pushToken Apple's Push Token
  */
 - (void)registerForPushNotificationsWithPushToken:(NSString *)pushToken
-                                        voipToken:(NSString *)voipToken
                                           success:(void (^)())successHandler
                                           failure:(void (^)(NSError *error))failureHandler
-NS_SWIFT_NAME(registerForPushNotifications(pushToken:voipToken:success:failure:));
+NS_SWIFT_NAME(registerForPushNotifications(pushToken:success:failure:));
```
yuliaveres: @designatednerd so we did not update this in SSK ?
I think that is the proper place ;)

designatednerd: Yep, that's what [this pull](https://github.com/toshiapp/Signal-iOS/pull/2) is - once that gets approved i'll pull it in

yuliaveres: ohh gotcha   .. didnt notice this one 😬  on it

PR #601, opened 2017-11-29
Update to fixed ssk which uses proper keychain service name

yuliaveres:


Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -566,7 +566,7 @@ - (NSData *)databasePassword
 
     NSError *keyFetchError;
     NSString *dbPassword =
-    [SAMKeychain passwordForService:keychainService account:keychainDBPassAccount error:&keyFetchError];
+    [SAMKeychain passwordForService:self.accountName account:keychainDBPassAccount error:&keyFetchError];
```
designatednerd: so the account name is the name of the keychain service here?

yuliaveres: yes

yuliaveres: as we used in previous versions

Pods/SignalServiceKit/SignalServiceKit/src/Util/OWSDispatch.h
```diff
@@ -24,6 +24,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
-void AssertIsOnMainThread(void);
+void AssertIsOnMainThread();
```
designatednerd: these two you should just be able to revert manually since i'm gonna update SSK for the other stuff

yuliaveres: so should i do it manually after each `pod install ` ? :/

Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
designatednerd: ahhhhh right I forgot about the way SAMKeychain is set up. 👍

Pods/SignalServiceKit/SignalServiceKit/src/Util/OWSDispatch.h
yuliaveres: oh you mean in ssk repo ? i can do it, just confirm 😸

yuliaveres: hmm in their repo it is without `void` ... 😨

yuliaveres:
@designatednerd  if that looks ok, could you approve ?

Pods/SignalServiceKit/SignalServiceKit/src/Util/OWSDispatch.h
designatednerd: yeah - it's mostly a silly thing causing compilation warnings on our end. I may look for another way to suppress those

PR #602, opened 2017-11-30
Layout all the cells

marijnschilling:
I layed out all the cells, to make it work with dynamic type as well 💅 

cells_be_dynamic


marijnschilling:
I will need to remove the testtableview before merging! But will do that once it is approved.

Toshi/Controllers/TabBarController.swift
```diff
@@ -86,7 +86,10 @@ open class TabBarController: UITabBarController, OfflineAlertDisplaying {
 
     @objc public func setupControllers() {
         browseViewController = BrowseNavigationController(rootViewController: BrowseViewController())
-        favoritesController = FavoritesNavigationController(rootViewController: FavoritesController())
+
+        //WARNING: to test the cells I create a TestTableViewController here instead of Favorites
+//        favoritesController = FavoritesNavigationController(rootViewController: FavoritesController())
```
designatednerd: 🚨 reminder to remove the test controller and use the real one 🚨

Toshi/Views/Cells/AvatarTitleCell.swift
```diff
@@ -15,26 +15,28 @@
...
+        leftImageView.contentMode = .scaleAspectFill
+
+        contentView.addSubview(leftImageView)
+        contentView.addSubview(titleTextField)
+
+        leftImageView.size(CGSize(width: BasicTableViewCell.imageSize, height: BasicTableViewCell.imageSize))
+        leftImageView.centerY(to: contentView, priority: .defaultHigh)
+        leftImageView.left(to: contentView, offset: BasicTableViewCell.horizontalMargin)
+        leftImageView.top(to: contentView, offset: 6, relation: .equalOrGreater)
+        leftImageView.bottom(to: contentView, offset: -6, relation: .equalOrGreater)
```
designatednerd: @roberthein is this something that could be easier with TinyConstraints?

designatednerd: (@marijnschilling if it's already using that feel free too say so)

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,49 @@
...
+
+import UIKit
+
+final class AvatarTitleSubtitleCell: BasicTableViewCell {
+
+    override func addSubviewsAndConstraints() {
+
+        let leftImageView = UIImageView(frame: .zero)
+        let titleTextField = UITextField(frame: .zero)
+        let subtitleLabel = UILabel(frame: .zero)
```
designatednerd: do these all need to have `.translatesAutoresizingMaskIntoConstraints` set to `false`?

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+final class AvatarTitleSubtitleCell: BasicTableViewCell {
+
+    override func addSubviewsAndConstraints() {
+
```
designatednerd: looks like a spare newline snuck in

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,49 @@
...
+        titleTextField.right(to: contentView, offset: -BasicTableViewCell.horizontalMargin)
+
+        subtitleLabel.topToBottom(of: titleTextField)
+        subtitleLabel.leftToRight(of: leftImageView, offset: BasicTableViewCell.verticalMargin)
+        subtitleLabel.right(to: contentView, offset: -BasicTableViewCell.horizontalMargin)
+        subtitleLabel.bottom(to: contentView, offset: -BasicTableViewCell.verticalMargin)
+
+        self.leftImageView = leftImageView
+        self.titleTextField = titleTextField
+        self.subtitleLabel = subtitleLabel
```
designatednerd: would it help readability/code swappability here to separate these out into three separate private methods - `setupImageView`, `setupTitleTextField`, and `setupSubtitle`?

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -66,9 +72,13 @@ class ToshiTableViewCell: UITableViewCell {
     private func setupTextStyles() {
         titleTextField?.font = Theme.preferredRegular()
         titleTextField?.isUserInteractionEnabled = false
+        titleTextField?.adjustsFontForContentSizeCategory = true
```
designatednerd: 👍 neat

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        cell.titleTextField?.text = cellData.title
+        cell.subtitleLabel?.text = cellData.subtitle
+        cell.detailsLabel?.text = cellData.details
+        cell.leftImageView?.image = cellData.leftImage
+        cell.switchControl?.isOn = cellData.switchState == true
+    }
+}
+
+class TestViewModel {
```
designatednerd: worth pulling any of this test data into actual tests?

designatednerd:
Same questions on a bunch of the other files as on `BasicTableViewCell`

- Separating per-item setup out into individual methods
- TinyConstraints
- Lack of `.translatesAutoresizingMaskIntoConstraints = false`

Toshi/Views/Cells/AvatarTitleCell.swift
```diff
@@ -15,26 +15,28 @@
...
+
+        titleTextField.centerY(to: contentView)
+        titleTextField.leftToRight(of: leftImageView, offset: BasicTableViewCell.verticalMargin)
+        titleTextField.right(to: contentView, offset: -BasicTableViewCell.horizontalMargin)
+
+        self.leftImageView = leftImageView
```
yuliaveres: :/ optionals ... is that the reason why u assign like this ? 🤔

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        contentView.addSubview(leftImageView)
+        contentView.addSubview(titleTextField)
+        contentView.addSubview(subtitleLabel)
+
+        leftImageView.size(CGSize(width: BasicTableViewCell.imageSize, height: BasicTableViewCell.imageSize))
+        leftImageView.centerY(to: contentView)
+        leftImageView.left(to: contentView, offset: BasicTableViewCell.horizontalMargin)
+        leftImageView.top(to: contentView, offset: 6, relation: .equalOrGreater, priority: .defaultLow)
+        leftImageView.bottom(to: contentView, offset: -6, relation: .equalOrGreater, priority: .defaultLow)
```
yuliaveres: why sometimes we have a constant and sometimes no ? .. maybe lets keep consistent ? .. 😄

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        titleTextField.top(to: contentView, offset: BasicTableViewCell.horizontalMargin)
+        titleTextField.leftToRight(of: leftImageView, offset: BasicTableViewCell.verticalMargin)
+        titleTextField.right(to: contentView, offset: -BasicTableViewCell.horizontalMargin)
+
+        subtitleLabel.topToBottom(of: titleTextField)
+        subtitleLabel.leftToRight(of: leftImageView, offset: BasicTableViewCell.verticalMargin)
```
yuliaveres: `vertical` ?

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+    override public func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let configurator = TestCellConfigurator()
+
+        let sectionData = viewModel.sectionModels[indexPath.section]
+        let cellData = sectionData.cellsData[indexPath.row]
+
+        let reuseIdentifier = configurator.cellIdentifier(for: cellData.components)
+
+        let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath)
+        configurator.configureCell(cell, with: cellData)
```
yuliaveres: 😋

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        return sectionData.cellsData.count
+    }
+
+    override public func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let configurator = TestCellConfigurator()
+
+        let sectionData = viewModel.sectionModels[indexPath.section]
+        let cellData = sectionData.cellsData[indexPath.row]
+
+        let reuseIdentifier = configurator.cellIdentifier(for: cellData.components)
```
yuliaveres: 😋

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath)
+        configurator.configureCell(cell, with: cellData)
+
+        return cell
+    }
+}
+
+class TestCellConfigurator: ToshiCellConfigurator {
```
yuliaveres: should be `final`, even if `Test` 😏

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+    func configureCell(_ cell: UITableViewCell, with cellData: TableCellData) {
+        guard let cell = cell as? BasicTableViewCell else { return }
+        guard let tag = TestItemType(rawValue: cellData.tag) else { return }
+
+        cell.titleTextField?.text = cellData.title
```
yuliaveres: hmm i guess same part will repeat in each single `Configurator` class.... inside `configureCell`... 😞 
**wdyt** if we could extract all content setup into separate method and provide default implementation which is basically same for all configurators? (by content setup i mean `cell.titleTextField?.text = ...`, `cell.subtitleLabel?.text = ...` etc.). So in same manner we do for `reuseIdentifier(for components: ...)` ?

yuliaveres: lets leave `configureCell` for another optional setup. not each configurator will need to implement it. (f.e. centered title cell will need additional setup , saynig - hey your title should be aligned by center..). see ?

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        configurator.configureCell(cell, with: cellData)
+
+        return cell
+    }
+}
+
+class TestCellConfigurator: ToshiCellConfigurator {
+    func configureCell(_ cell: UITableViewCell, with cellData: TableCellData) {
+        guard let cell = cell as? BasicTableViewCell else { return }
+        guard let tag = TestItemType(rawValue: cellData.tag) else { return }
```
yuliaveres: u dnt use it here right ?

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
marijnschilling: Njope! Because I use `TinyConstraints` that happens automatically 🎉

Toshi/Views/Cells/AvatarTitleCell.swift
marijnschilling: Yup it is 💛

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
marijnschilling: Hmmm might be interesting, but since the setup of the different views are depending on eachother (title goes to the left of the image view, and the subtitle below the title etc) I feel that there is also something to say for doing it all in one method. What do you think?

Toshi/Views/Cells/AvatarTitleCell.swift
marijnschilling: Yeah, so I don't want to force unwrap the views since that could crash the app if this code get's changed or rewritten. 

I looked into using `lazy vars` for the views but this would take some rewriting to make sure that there wont be views created that aren't necessary for that specific cell. 

So we could either keep it this way, which is a little... weird. Or make the views lazy and do the setup more in the cells themselves than in the `BasicTabeViewCell`

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
marijnschilling: yeah... damn designers using the vertical margin value for a horizontal spacing 🙈  I can better just use `10.0` here I guess.

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
designatednerd: yeah what i usually do is something like this: 

```swift
func setupAllViews() {
    setupView1()
    setupView2()
    setupView3()
}

private func setupView1() {
    // actually set up view 1
}

private func setupView2() {
    // actually set up view 2
}

private func setupView3() {
    // actually set up view 3
}
```

You still get the advantage of being able to set things up in a given order (you can even pass in a given view to any of the setup methods to make sure it's set up properly rather than calling the property directly), but you also get the clarity of having bits broken out into methods. 

Not required - it's not a pattern I see used in too many other places in Toshi (YET) but I have found it to be very helpful when working with views created entirely in code.

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
designatednerd: I have found it very helpful to create a constant, even if it's local, to provide some context to the next developer.

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
designatednerd: I would still advise a constant - you can have it have a different name but the same value if you want

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
marijnschilling: I <3 breaking up things :) So i'm completely down with this!

Toshi/Views/Cells/TestTableViewController.swift
marijnschilling: I find it hard to imagine what i would test....

Toshi/Views/Cells/AvatarTitleSubtitleCell.swift
marijnschilling: yes `interItemSpacing`!

PR #603, opened 2017-11-30
Prepare password in keychain in advance

yuliaveres:
Sometimes saving password in keychain is not completed while db being opened.
we should try accessing it beforehand.


Pods/SignalServiceKit/SignalServiceKit/src/Storage/TSStorageManager.m
```diff
@@ -582,6 +583,19 @@ - (void)backgroundedAppDatabasePasswordInaccessibleWithErrorDescription:(NSStrin
                 format:@"%@", errorDescription];
 }
 
+- (void)prepareDatabasePasswordIfNeededForKey:(NSString *)key
+{
+    [SAMKeychain setAccessibilityType:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly];
+
+    NSError *keyFetchError;
+    NSString *dbPassword =
+    [SAMKeychain passwordForService:key account:keychainDBPassAccount error:&keyFetchError];
```
marijnschilling: why is this on two lines?

yuliaveres: long line .. 🤷‍♂️

PR #604, opened 2017-11-30
Cleanup via AppCode

designatednerd:
Was trying to find some unused methods/files and ran things in AppCode. Took forever for it to actually finish the code inspections so in the meantime I fixed a ton of typos. Turned out there weren't really any files that aren't completely used. `¯\_(ツ)_/¯`


designatednerd:
Yes, the code inspections do spellcheck, but it's a little over-aggressive - it gave me hundreds and hundreds of errors around things like Lorem Ipsum text and hex strings 😝

roberthein:
hehe

PR #605, opened 2017-12-01
Use DLog/OCDLog instead of print/NSLog

designatednerd:
This swap guarantees a couple things: 

1. Whether or not logs will be printed is decided in one and only one place
2. Logs will always have the same format in terms of date / file / line where they're firing
3. Logs will only be printed when debugging. 

I'd say going forward it's fine to use `print`/`NSLog` for debugging, but if there's anything we want to leave in long-term, we should wrap it in one of these to ensure it never prints out in production. 

I'm not a *giant* fan of how it looks from Obj-C, but it's probably good that it's ugly because it'll discourage us from adding permanent logs. 😛


marijnschilling:
I have actually wondered myself why we print out so much errors and statusses and stuff. I would only look at that if i'm actually debugging those functionalities, and that I would add them myself and remove before merigng a PR. 

And I would say most of these errors are now caught automatically with the unit tests (or at least they should be!) So it might be a good time to just remove most of these logs...

designatednerd:
@marijnschilling I gotta fix some merge conflicts anyway - if you want to point out logs you know can definitely go I'll get rid of them.

marijnschilling:
I think we should have @yuliaveres go over this as well. I honestly don't see the immediate need for any of these logs. But that might have something to do with the fact that im mostly involved in the UI part of the app 😬

Toshi/Controllers/Favorites/FavoritesController.swift
```diff
@@ -47,7 +47,7 @@ open class FavoritesController: SweetTableController, KeyboardAdjustable, Emptia
 
         let searchText = searchController.searchBar.text?.lowercased() ?? ""
 
-        let filteringBlock: YapDatabaseViewFilteringWithObjectBlock = { transaction, group, _, key, object in
+        let filteringBlock: YapDatabaseViewFilteringWithObjectBlock = { transaction, group, colelction, key, object in
```
yuliaveres: `collection` ?

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -432,17 +435,26 @@ - (void)registerForRemoteNotifications {
...
 - (void)updateRemoteNotificationCredentials {
-    NSLog(@"\n||--------------------\n||\n|| --- Account is registered: %@ \n||\n||--------------------\n\n", @([TSAccountManager isRegistered]));
+    [OCDLog dlog:[NSString stringWithFormat:@"\n||--------------------\n||\n|| --- Account is registered: %@ \n||\n||--------------------\n\n", @([TSAccountManager isRegistered])]
+        filePath:__FILE__
+        function:__FUNCTION__
+            line:__LINE__];
 
     [[TSAccountManager sharedInstance] registerForPushNotificationsWithPushToken:self.token voipToken:nil success:^{
-        NSLog(@"\n\n||------- \n||\n|| - TOKEN: chat PN register - SUCCESS: token: %@\n", self.token);
+        [OCDLog dlog:[NSString stringWithFormat:@"\n\n||------- \n||\n|| - TOKEN: chat PN register - SUCCESS: token: %@\n", self.token]
```
yuliaveres: i would recommend using `ALog` here. This is pretty important message and as often we use Distribution while testing / developing, lets always display.

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -432,17 +435,26 @@ - (void)registerForRemoteNotifications {
...
+            function:__FUNCTION__
+                line:__LINE__];
 
         [[EthereumAPIClient shared] registerForMainNetworkPushNotifications];
 
         [[EthereumAPIClient shared] registerForSwitchedNetworkPushNotificationsIfNeededWithCompletion:nil];
 
     } failure:^(NSError *error) {
-        NSLog(@"\n\n||------- \n|| - TOKEN: chat PN register - FAILURE: %@\n||------- \n", error.localizedDescription);
+        [OCDLog dlog:[NSString stringWithFormat:@"\n\n||------- \n|| - TOKEN: chat PN register - FAILURE: %@\n||------- \n", error.localizedDescription]
```
yuliaveres: same here

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -465,7 +477,10 @@ - (void)application:(UIApplication *)application didRegisterForRemoteNotificatio
 }
 
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error {
-    NSLog(@"Failed to register for remote notifications. %@", error);
+    [OCDLog dlog:[NSString stringWithFormat:@"Failed to register for remote notifications. %@", error]
```
yuliaveres: let use `ALog` for this message ?

Toshi/APIClients/ChatAPIClient.swift
```diff
@@ -79,16 +79,16 @@ public class ChatAPIClient: NSObject {
...
                     guard response.statusCode == 204 else {
-                        print("Could not register user. Status code \(response.statusCode)")
+                        DLog("Could not register user. Status code \(response.statusCode)")
                         completion(false)
                         return
                     }
 
                     TSAccountManager.sharedInstance().storeServerAuthToken(parameters.password, signalingKey: parameters.signalingKey)
-                    print("Successfully registered chat user with address: \(cereal.address)")
+                    DLog("Successfully registered chat user with address: \(cereal.address)")
```
yuliaveres: could u also print this in all targets ?

Toshi/Controllers/Favorites/FavoritesController.swift
designatednerd: It wasn't used so I just set it to be ignored rather than fixing the spelling.

designatednerd:
@yuliaveres I broke out the logic of when to debug print (which was a good suggestion and caught an issue with my Objc dlog code) so that it can be overridden temporarily.

yuliaveres:
looks like what we need, thx @designatednerd

PR #606, opened 2017-12-01
Lazy var for cell views

marijnschilling:
Experiment branching from `feature/cell-layout` to test working with lazy var views, that only get accessed when necessary.


Toshi/Views/CellConfigurator/CellConfigurator.swift
```diff
@@ -21,25 +21,42 @@ protocol ToshiCellConfigurator: class {
     func cellIdentifier(for components: TableCellDataComponents) -> String
 }
 
-class CellConfigurator: ToshiCellConfigurator { // lets say we have this configurator responsible for
-
+extension ToshiCellConfigurator {
```
marijnschilling: here i added a default implementation for the cofigureCell method, that switches over the tags to set the data accordingly.

Toshi/Views/CellSectionData/TableViewData.swift
```diff
@@ -65,13 +65,14 @@ public final class TableCellData {
 
     private(set) var components: TableCellDataComponents = []
 
-    init(title: String? = nil, isPlaceholder: Bool = false, subtitle: String? = nil, leftImage: UIImage? = nil, details: String? = nil, switchState: Bool? = nil) {
+    init(title: String? = nil, isPlaceholder: Bool = false, subtitle: String? = nil, leftImage: UIImage? = nil, details: String? = nil, switchState: Bool? = nil, tag: Int) {
```
marijnschilling: So initialising `TableCellData` now requires a raw tag value.

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -17,6 +17,17 @@ import Foundation
 import UIKit
 import TinyConstraints
 
+enum ItemType: Int {
```
marijnschilling: This seems to make more sense to be a inside of `BasicTableViewCell`

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -36,29 +47,65 @@ class BasicTableViewCell: UITableViewCell {
     static let verticalMargin: CGFloat = 10.0
     static let imageSize: CGFloat = 38.0
 
-    var actionDelegate: BasicCellActionDelegate? {
-        didSet {
-            setupActions()
```
marijnschilling: the actions are now setup in the lazy vars of the properties that need actions

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -101,7 +127,13 @@ class BasicTableViewCell: UITableViewCell {
     override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
 
-        setupTextStyles()
+        titleTextField.font = Theme.preferredRegular()
```
marijnschilling: This is the only place that i haven't found a solution for :( We need to re-set the fonts whent he traintcolleciton changes to make resizing fonts work instantly. But we cant be sure at this point which of the labels are visible on this cell :( 

Only solution I know is calling a method in each subclass, but thats very redundant.

designatednerd: It's redundant, but I think it's necessary - since otherwise you may have to map the appropriate font sizes in a dicitonary with the labels. That seems like WAY too much work vs. what a subclass is generally designed to do

Toshi/Views/CellSectionData/TableViewData.swift
designatednerd: why not just pass in the `ItemType` instead of the raw value of the tag?

Toshi/Controllers/Recent/RecentViewController.swift
```diff
@@ -21,7 +21,7 @@ public extension NSNotification.Name {
     public static let ChatDatabaseCreated = NSNotification.Name(rawValue: "ChatDatabaseCreated")
 }
 
-open class RecentViewController: SweetTableController, Emptiable {
+open class RecentViewController: SweetTableController, Emptiable, ToshiCellConfigurator {
```
yuliaveres: we should not mix controller with configurator .... 😞

Toshi/Views/Cells/BasicTableViewCell.swift
yuliaveres: 👍

Toshi/Views/CellSectionData/TableViewData.swift
marijnschilling: Because cells shouldn't know about the model. We might want to use different sets up itemTypes throughout the app, this way we can just map them to ints.

Or we could also just do set the ItemType there, but ☝️ was my reasoning. What do you think?

Toshi/Controllers/Recent/RecentViewController.swift
marijnschilling: Why not? Isnt the Controller the one who configurates the cells in the end because that is also the DataSource.

marijnschilling: Like before we would call `configurator.configure` from the Controller anyway in the delegate method implementations of TableViewDelegate or TableViewDataSource. While that configurator is nothing more than just a porotocol that suffices with a default implementation.

marijnschilling:
Oke added the split up methods to setup the views in each ell in this PR, cause it will get quite ugly 🍝 when we work with optionals there.

Toshi/Views/CellSectionData/TableViewData.swift
designatednerd: I mean if it has to know the tag, then doesn't it need to know the `itemType` it maps to? Seems like it might be easy to pass a value for a tag that doesn't exist otherwise.

Toshi/Views/Cells/BasicTableViewCell.swift
yuliaveres: i do not think so .
it should be in models for each controller as each might have unique and specific cell item type.  + cell never knows anything about type or tag. Configurator does only.
And then each configurator works with cells and might use tag at some point to do additional configuration of particular cell if it is needed. that is why we should keep Item type in ViewModels as they differ per each.

Toshi/Controllers/Recent/RecentViewController.swift
yuliaveres: no, the whole point of developing predefined cells and configurators was to remove any configurations from controller and let configurator instance when he should do the job: `cellForRow: ...` -> `configurator.configure(cell: ...)`

Toshi/Views/Cells/TitleCell.swift
```diff
@@ -21,11 +21,8 @@ final class TitleCell: BasicTableViewCell {
     override open func addSubviewsAndConstraints() {
         accessoryType = .disclosureIndicator
 
-        let titleTextField = UITextField(frame: .zero)
         contentView.addSubview(titleTextField)
 
         titleTextField.edgesToSuperview(insets: TinyEdgeInsets(top: 10, left: 16, bottom: 10, right: 0))
```
yuliaveres: also didn't notice but why `right: 0` ?

Toshi/Views/CellConfigurator/CellConfigurator.swift
```diff
@@ -21,25 +21,42 @@ protocol ToshiCellConfigurator: class {
...
     func configureCell(_ cell: UITableViewCell, with cellData: TableCellData) {
         guard let cell = cell as? BasicTableViewCell else { return }
+        guard let tag = ItemType(rawValue: cellData.tag) else { return }
+
+        switch tag {
+        case .title:
+            break
+        case .titleSubtitle:
```
yuliaveres: wait a moment. now or am I mistaken or we have `ItemType` and `TableCellDataComponents`which are identical ? 
the initial idea of `itemType` - was semantic meaning of an item (lets say in some controller you will have a `Leave group` cell with centered title which is of red color. and only in this controller and this type. then ItemType would be specific - `.leaveGroup` and group info configurator will use additional logic in case of this item type - making it red.) . 
but title, subtitle, details - all of those are just components ...

yuliaveres: thus - we should think of item types range - as specific per controller.
and here you might want to configure cell based not on `ItemType`  but `cellData.components`.

Toshi/Views/Cells/BasicTableViewCell.swift
marijnschilling: Actually realised this as well! 

So the enum for cell type will be connected to the Controller? And then the different Controllers in toshi will have different enums?

I think that sounds like a pretty good approach, and also answers @designatednerd s question about why the BasicTableViewCell's `tag` property is an Int and not just the enum 🎉

Toshi/Controllers/Recent/RecentViewController.swift
marijnschilling: I don't really understand, i haven't changed where we configure the cells. Just removed an unused object. We let the protocol deal with the actual configuration.

How i set it up right now, with a default implementation of the `configure(cell: ...)` method in the protocol the `Configurator` class would just be an empty class implementing the protocol, that doesn't make sense right? Or how would you propose solving this?

Toshi/Views/Cells/TitleCell.swift
marijnschilling: Because the disclosure indicator has an inset, so for the textField we can take all the space on the right: 

screenshot 2017-12-04 14 46 05

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -45,13 +38,15 @@ class TestTableViewController: UITableViewController, ToshiCellConfigurator {
...
+
         let sectionData = viewModel.sectionModels[indexPath.section]
         let cellData = sectionData.cellsData[indexPath.row]
 
-        let reuseIdentifier = cellIdentifier(for: cellData.components)
+        let reuseIdentifier = configurator.cellIdentifier(for: cellData.components)
 
         let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath)
-        configureCell(cell, with: cellData)
+        configurator.configureCell(cell, with: cellData)
```
yuliaveres: 🌟

Toshi/Views/Cells/TestTableViewController.swift
```diff
@@ -1,17 +1,10 @@
...
 enum ItemType: Int {
-    case title
-    case titleSubtitle
-    case titleSwitch
-    case titleSubtitleSwitch
-    case avatarTitle
-    case avatarTitleSubtitle
-    case avatarTitleSubtitleDetails
-    case avatarTitleSubtitleSwitch
+    case redTextCell
```
yuliaveres: ideally it would be happy to reside in `viewModel`, as she creates `TableCellData`-s and sets tag if needed (`cellData.tag = .redTextCell`). 
But it is not that important in test controller. :-)

PR #607, opened 2017-12-01
Add wrapper for type-safe access to User Defaults

designatednerd:
This PR adds a wrapper for type-safe access to `NSUserDefaults`. It also cuts down on the code it takes to access any individual value at the call site, and will eventually allow us to pass in a mocked set of defaults during UI testing or use defaults in a container WAY more easily. 

This also eliminates all calls to `synchronize`, which are basically voodoo at this point anyway given that the [documentation for that method](https://developer.apple.com/documentation/foundation/nsuserdefaults/1414005-synchronize) ends with "this method is unnecessary and shouldn't be used."

This also adds a handy little protocol with a default implementation I've used a bunch of times which allows you to access all the cases in a `String` enum. 

Edit: This now also removes a whole bunch of files from the test target and uses the `@testable import` functionality instead, which allows testing internal methods within a module.


designatednerd:
retest this please

Toshi/Protocols/StringCaseListable.swift
```diff
@@ -0,0 +1,40 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+/// A protocol to allow grabbing an array of all the cases of a string enum.
+/// Inspired by: http://stackoverflow.com/a/32429125/681493
+public protocol StringCaseListable {
```
roberthein: Nice! ✨

PR #611, opened 2017-12-04
Force chat registration update in case prekey registration failed

yuliaveres:
Currently there is an issue, user who experienced db loss - is not registered with chat.
Log out / in fixes this. Otherwise - user will never be able to chat.
Unfortunately there is no way to check if user is registered with chat. Also it is not a good way to register on each launch triggering `PUT` request with `UserBootstrapParameters`.
We can use a flag which checks prekeys once for each user (already logged in) and in case we receive `unauthorized` response - we want to force chat registration update.


Toshi/Supporting Files/AppDelegate.m
```diff
@@ -366,6 +366,31 @@ - (void)configureForCurrentSession
...
+
+- (void)tryRegisteringPrekeysIfNeeded
+{
+    if ([UserDefaultsWrapper chatRegistrationUpdateTriggered] == NO) {
+
+        [UserDefaultsWrapper setChatRegistrationUpdateTriggered:YES];
+
+        [TSPreKeyManager registerPreKeysWithMode:RefreshPreKeysMode_SignedAndOneTime
+                                         success:^{
+
```
designatednerd: should this just be `nil` rather than empty?

yuliaveres: there is no check inside if it is nil, it is just called and in case `nil` - crashes

yuliaveres:
@designatednerd if you have any better suggestion how can we handle this - let me know !

Toshi/Utilities/UserDefaultsWrapper.swift
```diff
@@ -30,7 +30,8 @@ enum UserDefaultsKey: String, StringCaseListable {
     selectedApp = "Restoration::SelectedApp",
     selectedContact = "Restoration::SelectedContact",
     selectedThreadAddress = "Restoration::SelectedThread",
-    tabBarSelectedIndex = "TabBarSelectedIndex"
+    tabBarSelectedIndex = "TabBarSelectedIndex",
+    chatRegistrationUpdateTriggered = "chatRegistrationUpdateTriggered"
```
designatednerd: Since this is a string based enum, by default the raw value is the name of the case, so in theory you wouldn't need to set it again here. 

However, please follow the request in the comments to use a `Restoration::` prefix.

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -366,6 +366,31 @@ - (void)configureForCurrentSession
...
+                                         } failure:^(NSError *error) {
+                                             [CrashlyticsLogger log:@"Failed registering prekeys - triggering Chat register" attributes:nil];
+
+                                             if (error.code == 401) {
+                                                 [ChatAPIClient.shared registerUserWithCompletion:^(BOOL success) {
+                                                     if (success) {
+                                                         [CrashlyticsLogger log:@"Successfully registered user with chat service after forced trigger" attributes:nil];
+                                                     }
```
designatednerd: do you also want to log failures on this?

yuliaveres: prolly yes

Toshi/Supporting Files/AppDelegate.m
designatednerd: Ah - I'd at least throw a comment in here to that effect then.

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -366,6 +366,31 @@ - (void)configureForCurrentSession
...
     [SignalNotificationManager updateUnreadMessagesNumber];
+
+    [self tryRegisteringPrekeysIfNeeded];
+}
+
+- (void)tryRegisteringPrekeysIfNeeded
+{
+    if ([UserDefaultsWrapper chatRegistrationUpdateTriggered] == NO) {
+
+        [UserDefaultsWrapper setChatRegistrationUpdateTriggered:YES];
```
designatednerd: Does this ever need to get set back to `NO`?

yuliaveres: i think NO. but maybe we should set YES only in case of already registered successfully (no need to retry later) , or registered successfully after forced trigger (also no need to retry)..

Toshi/Utilities/UserDefaultsWrapper.swift
```diff
@@ -100,6 +101,15 @@ class UserDefaultsWrapper: NSObject {
             setValue(newValue, for: .addressChangeAlertShown)
         }
     }
+
+    @objc static var chatRegistrationUpdateTriggered: Bool {
```
designatednerd: Mind throwing in some docu-comments here to explain what this is looking at? I know I hadn't added them on other stuff yet but I plan to.

designatednerd:
I'm still getting more familiar with the codebase so I don't have any better ideas right now - this is a bit brute-force, but if it works, it works.

Toshi/Utilities/UserDefaultsWrapper.swift
```diff
@@ -100,6 +101,18 @@ class UserDefaultsWrapper: NSObject {
             setValue(newValue, for: .addressChangeAlertShown)
         }
     }
+
+    /// Looks at whether user's prekeys were checked once after db loss issue
+    /// and in case of failure chat registration was executed
+
```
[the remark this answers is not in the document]
yuliaveres: ok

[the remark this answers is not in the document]
yuliaveres: done @designatednerd

PR #613, opened 2017-12-05
Fix avatars on Browse being not cached at first launch

yuliaveres:
We call `cachedImage` on avatar manager but on first launch there is no cached images yet. We should call `download avatar` which inside also checks for cached and if there is any - returns it, otherwise - downloads it.


Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -300,10 +300,13 @@ extension BrowseViewController: UISearchBarDelegate {
...
 
                 cacheQueue.async {
-                    let image = AvatarManager.shared.cachedAvatar(for: item.avatarPath)
-                    DispatchQueue.main.async {
-                        let cell = sectionedCollectionView.cellForItem(at: indexPath) as? BrowseEntityCollectionViewCell
-                        cell?.avatarImageView.image = image
+                    AvatarManager.shared.avatar(for: item.avatarPath) { image, path in
+                        DispatchQueue.main.async {
+                            if item.avatarPath == path {
+                                let cell = sectionedCollectionView.cellForItem(at: indexPath) as? BrowseEntityCollectionViewCell
```
designatednerd: what happens if the cell has already been recycled by the time it hits this?

yuliaveres: nothing . 😄

designatednerd: It looks like you're comparing the captured `item.avatarPath` rather than anything on the current cell - which means that even if a new `item` has been set on the cell, the check you're using here isn't going to see that. 

Does `cellForItem(at:)` give you back a new cell even if the current cell has been recycled?

yuliaveres: @designatednerd as we check for `avatarPath` being same as `completion`'s one, guess no need to retrieve `item` here once more at the given `indexPath` 🤔

designatednerd: So the `item` is initially grabbed at line 295, and then captured by the `DispatchQueue` closure - no matter what, that's always going to be the same `item`, even if the cell has been recycled. 

Can you think of a way to check if the current cell still has the same index path it did when the load started?

yuliaveres: lets see

yuliaveres: regarding `cellForItem(at:` - it should ritgh ?

PR #614, opened 2017-12-05
Group chat invite cell

marijnschilling:
have some layout warnings that i do not understand, about the height of the labels.

✨ it already looks great though ✨ 
screenshot 2017-12-05 11 21 32


Toshi/Controllers/Recent/RecentViewController.swift
```diff
@@ -83,7 +83,7 @@ open class RecentViewController: SweetTableController, Emptiable {
 
         addSubviewsAndConstraints()
 
-        tableView.separatorStyle = .none
+//        tableView.separatorStyle = .none
```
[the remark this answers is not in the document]
marijnschilling: Yeah, I changed the setup of this table to show the cell i'm working on. But will change it back to the original before i merge. Just commited so that @roberthein can take a look at the constraints ✌️

Toshi/Views/Cells/DoubleAvatarTitleSubtitleCell.swift
```diff
@@ -0,0 +1,60 @@
...
+    override func addSubviewsAndConstraints() {
+        accessoryType = .disclosureIndicator
+
+        contentView.addSubview(doubleImageView)
+        contentView.addSubview(titleTextField)
+        contentView.addSubview(subtitleLabel)
+
+        setupDoubleImageView()
+        setupTitleTextField()
+        setupSubtitleLabel()
```
designatednerd: 👍 nice! For what it's worth I also add the subviews in the individual setup methods to keep it all together, but defintiely not a requirement

Toshi/Views/DoubleImageView.swift
```diff
@@ -0,0 +1,86 @@
...
+import TinyConstraints
+
+class DoubleImageView: UIView {
+    static let imageSize: CGFloat = 38.0
+    static let borderWidth: CGFloat = 2.0
+
+    private lazy var firstImageView: UIImageView = {
+        let view = UIImageView()
+
+        view.layer.cornerRadius = DoubleImageView.imageSize * 0.5
```
designatednerd: do we have some kind of utility extension either in this or in SweetUIKit to make it easy to make something into a circle? Seems like it might be worthwhile.

Toshi/Views/Cells/DoubleAvatarTitleSubtitleCell.swift
marijnschilling: But than you get a big fat crash if you connect the constraint to a view that hasnt been added yet :/

designatednerd: Ah, fair point - I usually set them up in the order where they're gonna be added, but that's probably a bit magic.

Toshi/Views/DoubleImageView.swift
marijnschilling: Would be cool, but i don't see how except for subclassing UIView which is not cool. :)

designatednerd: I've had an extension/category for a while that you call after autolayout fires: 

```width
func circleify() {
    self.layer.cornerRadius = self.frame.width / 2
}
```

You could probably set it up to take optional parameters for border width and color too.

marijnschilling: aha yes! And i call that in layoutSubViews and then it works! i added to local extensions, once its used in more places (so tested to be awesome) we'll add it to SweeeeetUI!

PR #615, opened 2017-12-05
Point to master SSK

yuliaveres:
Nothing basically changed, but PR on SSK for was merge so needed to reinstall pods.


yuliaveres:
thx guys ! 👏

PR #616, opened 2017-12-05
Pull password validation logic into a singleton and add tests

designatednerd:
I ran into the issue on my dev account where valid words from the word list in my passphrase which would generate errors, so I decided to take a stab at fixing it. 

This at least managed to fix my error, and gives us some assurance that we haven't screwed something up if we ever need to monkey with this under the hood.


Toshi/Utilities/PasswordValidator.swift
```diff
@@ -0,0 +1,49 @@
...
+
+import Foundation
+
+class PasswordValidator {
+    
+    private init() {}
+    static let shared = PasswordValidator()
+    
+    lazy var passwords: [String] = {
+        guard let path = Bundle.main.path(forResource: "passwords-library", ofType: "txt") else {
```
designatednerd: Loading from the path is the main reason I decided to make this a singleton - this way the list of passwords is only loaded once no matter what needs to use them. If you think it's better to just make this an instance, lmk.

marijnschilling:
Ah cool improvement! I think it looks way cleaner to have this logic separated. Way way way. 

I am just wondering if you figured out why it wasn't working before. Was that because the words where loaded multiple times as you mention in your changes? 

I don't really have strong opinions about this password generator stuff, since RH is the one who cerated it! I just ran in to the bug 😇

Tests/PasswordValidatorTests.swift
```diff
@@ -0,0 +1,32 @@
...
+    
+    func testLoadingWords() {
+        XCTAssertEqual(PasswordValidator.shared.passwords.count, 2049)
+    }
+    
+    func testValidatingKnownProblematicWords() {
+        let you = "you"
+        let youValidationResult = PasswordValidator.shared.validateWord(for: you)
+        XCTAssertEqual(youValidationResult.match, you)
+        XCTAssertEqual(youValidationResult.isSingleOccurrence, false)
```
marijnschilling: What do you test here exactly?

designatednerd:
I think it was more around a lot of stuff being optional and that eventually coming to bite us in the ass. I think it might have been OK if I'd left it in the VC and just made a bunch of stuff that was unnecessarily async synchronous. 

I didn't wind up digging too deep because I did the refactor to pull everything out first and once tests were passing, I dropped it in and it worked.

designatednerd:
@roberthein thoughts?

Toshi/Utilities/PasswordValidator.swift
```diff
@@ -0,0 +1,49 @@
...
+        }
+        
+        return longestPassword.count
+    }()
+    
+    func validateWord(for text: String) -> (match: String?, isSingleOccurrence: Bool) {
+        guard !text.isEmpty else { return (nil, false) }
+        
+        let filtered = passwords.filter {
+            $0.range(of: text, options: [.caseInsensitive, .anchored]) != nil
```
marijnschilling: I kinda like it better when $0 values are just a var name that defines what it is. Because I tend to find it hard to follow. 

so then it would be 
```
let filtered = passwords.filter { password in
             password.range(of: text, options: [.caseInsensitive, .anchored]) != nil
}
```

marijnschilling: maybe better to rename `filtered` to `matches` (as in multiple of match)

Tests/PasswordValidatorTests.swift
designatednerd: "you" is the prefix for a couple other things, so "you" should not be the single occurence of a validation result.

Toshi/Utilities/PasswordValidator.swift
designatednerd: Good call - kept the original name but that makes more sense

Tests/PasswordValidatorTests.swift
marijnschilling: oooh, might be good to test this in a different test with an explaining name `testMultipleMatchesForPrefix` 🤔  or something... I don't know.

or maybe add comments? 

So that when this test is failing we know immediately what the dealio is.

designatednerd:
@marijnschilling changes pushed - still looks good to you?

marijnschilling:
Still 👌  even better! 👏

Elland:
Coming in late, but do we really need this in-memory for the life-cycle of the app, instead of a one-off object that we deallocate when the sign in controller dies?

designatednerd:
i think the main reason i left it as a singleton was to avoid the hit of reloading the passwords from disk. i'm fine if you want to make it an instance on each VC though.

PR #617, opened 2017-12-05
iOS-415: Rename Browse to Search

designatednerd:
Note that I only updated the 1x asset in the tab bar to match the other tabs.

simulator screen shot - iphone 8 - 2017-12-05 at 15 58 57


Toshi/Resources/Assets.xcassets/Tabbar/tab1.imageset/Contents.json
```diff
@@ -2,7 +2,7 @@
   "images" : [
     {
       "idiom" : "universal",
-      "filename" : "tab1.pdf",
+      "filename" : "1@1x.png",
```
roberthein: I think this should be a `.pdf` like the rest of the tab bar images.

designatednerd: Ah, I missed that, mostly because the other images weren't showing up as universal size. Will update those as well.

PR #618, opened 2017-12-05
Update Circle config to use CI 2.0

designatednerd:
Mostly done, just trying to make sure Danger actually runs with the PR title


.circleci/config.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+    build-and-test:
+        macos:
+            xcode: "9.1.0"
+        steps:
+            - checkout
+            - run:
+                name: Install with bundler
+                command: bundle install
```
Elland: Is this because of danger?

designatednerd: yep, although i'd want it for some fastlane stuff I want to do anyway

PR #623, opened 2017-12-06
Remove redundant access control keywords

designatednerd:
This PR removes a bunch of redundant access control statements, since all our code is in the same module: 

- `internal` is no longer necessary since that is the default access for all items in the same module. 
- `public` is not necessary on anything since all items within the same module can see each other, and our whole app is in the same module.
- `open` is not necessary because anything not marked `final` within a module can be subclassed within that module. 

If we decide to move anything into a framework we should revisit what needs to be `public`/`open` then but for now: None of it needs to be.


Toshi/Models/Users/TokenUser.swift
```diff
@@ -195,22 +195,23 @@ public class TokenUser: NSObject, NSCoding {
...
+    required convenience init?(coder aDecoder: NSCoder) {
         guard let jsonData = aDecoder.decodeObject(forKey: "jsonData") as? Data else { return nil }
         guard let deserialised = try? JSONSerialization.jsonObject(with: jsonData, options: []), let json = deserialised as? [String: Any] else { return nil }
 
         self.init(json: json)
     }
 
-    @objc(encodeWithCoder:) public func encode(with aCoder: NSCoder) {
+    @objc(encodeWithCoder:)
+    func encode(with aCoder: NSCoder) {
```
marijnschilling: What is the reason you put this on two lines?

Toshi/Utilities/DLog.swift
```diff
@@ -29,7 +29,7 @@ import Foundation
 /// NOTE: This can be overridden temporarily on a local basis, but it's not recommended that said changes be checked in
 ///
 /// - Returns: True if debug logs should fire, false if not.
-private func shouldDebugLog() -> Bool {
+func shouldDebugLog() -> Bool {
```
marijnschilling: Shouldn't we keep `private` function to stay private?

Toshi/Models/Users/TokenUser.swift
designatednerd: Mostly habit - If there's a different method signature for Obj-C, I find it easier to parse the objective-C name of something and how it relates to the Swift method signature if it's on the line above. I leave it just at the front if all it's doing is making a func or a var available without changing the name. 

I can put it back if you'd prefer though

Toshi/Utilities/DLog.swift
designatednerd: Whoopsies, good catch

Toshi/Models/Users/TokenUser.swift
Elland: Indeed, I too find it easier to parse.

Toshi/Utilities/DLog.swift
```diff
@@ -29,7 +29,7 @@ import Foundation
 /// NOTE: This can be overridden temporarily on a local basis, but it's not recommended that said changes be checked in
 ///
 /// - Returns: True if debug logs should fire, false if not.
-func shouldDebugLog() -> Bool {
+private func shouldDebugLog() -> Bool {
```
marijnschilling: Yay, now I'm eligible for a badge of `being still sharp when reviewing a pr with around 600 identical changes`!

designatednerd: https://www.youtube.com/watch?v=I_izvAbhExY

PR #624, opened 2017-12-06
Add tests

marijnschilling:
- [ ] test for `transactionSkeleton`
- [ ] test for `estimatedEthereumNetworkFeeString`


Tests/EthereumConverterTests.swift
```diff
@@ -25,6 +25,14 @@ class EthereumConverterTests: QuickSpec {
                     expect(ethereumValueString).to(equal("3.5000 ETH"))
                 }
 
+                it("returns the string representation of the ethereum network fees") {
+                    let ethereumNetworkFeeString = EthereumConverter.estimatedEthereumNetworkFeeString(for: wei, exchangeRate: exchangeRate)
+
+                    let dollarSting = String(format: "$100%@00 USD", TokenUser.current?.cachedCurrencyLocale?.decimalSeparator ?? ".")
+                    let feeSting = String(format: "The estimated Ethereum network fees are %@ (1.0000 ETH)", dollarSting)
```
yuliaveres: no localization needed ?

Tests/EthereumAPIClientTests.swift
```diff
@@ -46,6 +46,28 @@ class EthereumAPIClientTests: QuickSpec {
...
 
+                it("transactionSkeleton") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: EthereumAPIClientTests.self), mockFilename: "transactionSkeleton")
+                    mockTeapot.overrideEndPoint("timestamp", withFilename: "timestamp")
+                    subject = EthereumAPIClient(mockTeapot: mockTeapot)
+
+                    let parameters: [String: Any] = [
+                        "from": "0x011c6dd9565b8b83e6a9ee3f06e89ece3251ef2f",
+                        "to": "0x011c6dd9565b8b83e6a9ee3f06e89ece3251ef2f",
+                        "value": 1000
```
yuliaveres: still works even if value should be hex string ?

Tests/EthereumAPIClientTests.swift
```diff
@@ -46,6 +46,28 @@ class EthereumAPIClientTests: QuickSpec {
...
+                    ]
+
+                    waitUntil() { done in
+                        subject.transactionSkeleton(for: parameters) { skeleton, error in
+                            expect(skeleton.gas).toNot(beNil())
```
designatednerd: ☠️ 💨 

... I am twelve.

Tests/EthereumConverterTests.swift
designatednerd: Even though this is a test, this should be localized - that way we only have to make sure the string is changed in one place. 

Also, although fees do sting, I think you mean `feeString` 🐝

marijnschilling: not in tests

Tests/EthereumAPIClientTests.swift
designatednerd: Also, is there any way to validate that the parameters getting passed in are being parsed correctly?

Tests/EthereumConverterTests.swift
marijnschilling: Oke! I thought in tests you should always be as explicit as possible. So no localizations

designatednerd: That's generally true but there's also the factor of not wanting to have to change things twice, which can lead to failed tests pretty easily. The other thing is that it means any time you want to run your tests in another language, you have to write totally separate tests for each language, which scales...poorly. 

I've used something like [this](https://github.com/designatednerd/GoCubs/blob/master/GoCubs/DeStringers/LocalizedString.swift) to both a) Contain my localized strings in one place and b) Facilitate UI testing in the last few projects I've done, and I like it. 

Certainly doesn't need to be set up now but it's why I think in a localized app you ought to be testing the localized string and not a literal

Tests/EthereumAPIClientTests.swift
marijnschilling: yeah apparantly... i can maybe change it

Tests/EthereumAPIClientTests.swift
marijnschilling: nope, it just returns the mocked data since it overrides the `post` method in `Teapot`

designatednerd: right but do you want to verify that those are, in fact, the values getting set for those values? Basically, to validate that the parsing isn't doing something silly like assigning the `value` to `skeleton.transaction` or something

marijnschilling: Maybe, but that should be something to add to `MockTeapot`, which I made to be able to mock the tests for the APICAlls we make with Teapot. For now this is not part of the functionality, so this is what we'll have to be satisfied with right now :)

designatednerd: Talked IRL with @marijnschilling and I think I was unclear here - I was trying to make sure the values passed in via the JSON being returned by `Teapot` were actually what was coming through -she's gonna take a shot at making that work.

marijnschilling:
@designatednerd Also added the test for when we do an unauthorized call. Wonder about your opinion about these tests though. I guess i made them to make sure nothing crashes when we get an invalid response, but it also feels a bit redundant in a way to add these tests. Not sure how i feel about them :)

yuliaveres:
@marijnschilling  maybe change base branch to master now as `feature/ethereum-network-fee` is merged ?

Tests/EthereumAPIClientTests.swift
```diff
@@ -119,6 +119,28 @@ class EthereumAPIClientTests: QuickSpec {
                     }
                 }
 
+                it("fetches the transaction Skeleton") {
```
designatednerd: this probably should be different than the section above since it's testing how a 401 is handled, not the transaction skeleton itself

designatednerd:
@marijnschilling I personally err to the side of "test expected errors are handled properly", which I think this falls into. 

There's a big big big rabbit hole you can go down testing that every error you get back is handled properly, but it's reasonable to make sure a particular API doesn't blow up if it gets an error.

PR #629, opened 2017-12-07
[IOS-437]: Group chat feature

yuliaveres:


Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -126,6 +126,32 @@ class BrowseViewController: SearchableCollectionController {
...
+            guard let info = notification.userInfo else { return }
+            let duration = info[UIKeyboardAnimationDurationUserInfoKey] as? TimeInterval ?? 0.2
+            let curveRaw = info[UIKeyboardAnimationCurveUserInfoKey] as? UInt ?? 0
+            let curve = UIViewAnimationOptions(rawValue: curveRaw)
+
+            UIView.animate(withDuration: duration, delay: 0, options: UIViewAnimationOptions(rawValue: UInt(curve.rawValue)), animations: {
+                self?.searchResultView.contentInset.bottom = 0
+                self?.collectionView.contentInset.bottom = self?.bottomInset ?? 0.0
+            })
+        }
```
designatednerd: is this the stuff that's supposed to be replaced or am I thinking of something else

yuliaveres: replaced? i think it is one of the ui improvements prs. merged into here for some reason. see in blame

yuliaveres:
Remove the odd BrowseViewController UI change

Toshi/Models/SofaWrappers/SofaWrapper.swift
```diff
@@ -53,6 +55,8 @@ enum SofaType: String {
             self = .paymentRequest
         } else if sofa.hasPrefix(SofaType.payment.rawValue) {
             self = .payment
+        } else if sofa.hasPrefix(SofaType.payment.rawValue) {
+            self = .status
```
designatednerd: ^^ looks like think this slipped through - you can fix after merge if you want though

PR #631, opened 2017-12-07
Get rid of some warnings

yuliaveres:


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -269,7 +269,7 @@ extension SOFAWebController: WKScriptMessageHandler {
                 return
             }
 
-            self.sendSignedTransaction(signedTransaction, with: callbackId, completion: { [weak self] callBackId, payload in
+            self.sendSignedTransaction(signedTransaction, with: callbackId, completion: { [weak self] callbackId, payload in
```
designatednerd: do you want to use a different variable name to make it clear that the variable being passed in below is the updated callback ID?

yuliaveres: any idea how to name the one from completion ? it is not updated, it is just "FOR SURE SAME to the one we received in parameter". (before there was a variable on the controller storing that `callbackId` and if you had a lot of events you might have ended up with wrong callbackID. now we just decided to pass it along the flow so for sure in completion we have callbackId associated with current event).

yuliaveres: @designatednerd  ☝️

designatednerd: I usually do something like `returnedCallbackId` just to differentiate it

PR #581, opened 2017-11-15
Prepare New-Chat for Group-Chat

roberthein:
This does a bunch of refactoring from Contacts -> Profiles and adds the ability for group chats to be started. 

simulator screen shot - iphone 8 - 2017-12-06 at 17 42 08
simulator screen shot - iphone 8 - 2017-12-06 at 17 42 12
simulator screen shot - iphone 8 - 2017-12-06 at 17 42 14
simulator screen shot - iphone 8 - 2017-12-06 at 17 42 15

PR #633, opened 2017-12-07
Initial swipe at iOS 10 Group chat

designatednerd:
Picking up from where I left off at #581: 

- [x] Actually have search working in iOS 10 instead of disappearing on click
- [ ] Figure out how users added to the list of selected users for a group chat should be listed
- [ ] UI polish


designatednerd:
Just talked to Marek, apparently I misunderstood the entire design here, so fixing it in iOS 10 isn't going to be that helpful. This is at least working for now so I'd like to get this in and rip everything apart on a new branch.

Toshi/Controllers/Profiles/ProfilesAddGroupHeader.swift
```diff
@@ -29,6 +29,7 @@ class ProfilesAddGroupHeader: UIView {
         let view = LeftAlignedButton()
         view.icon = UIImage(color: .lightGray, size: CGSize(width: 48, height: 48))
         view.title = Localized("profiles_new_group")
+        view.icon = UIImage(named: "navigation_bar_add")
```
yuliaveres: `navigation_bar..` ?

Toshi/Controllers/Profiles/ProfilesAddedToGroupHeader.swift
```diff
@@ -32,15 +32,21 @@ class ProfilesAddedToGroupHeader: UIView {
         return view
     }()
     
+    private var margin: CGFloat = 0
+    
     convenience init(margin: CGFloat) {
         self.init(withAutoLayout: true)
+        self.margin = margin
```
yuliaveres: not sure why we need this margin? if it is zero we can use `edges(to: ...)` ?

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
```diff
@@ -28,8 +31,8 @@ final class ProfilesHeaderView: UIView {
             if #available(iOS 11, *) {
                 // Do nothing, search bar should already be set up
             } else {
-                height(searchBar?.frame.height ?? 0)
-                addSearchBar(searchBar, bottomPinnedTo: nil)
+                self.frame = CGRect(origin: .zero, size: (searchBar?.frame.size ?? .zero))
```
yuliaveres: do you want to leave `self`  ?

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
```diff
@@ -41,9 +44,10 @@ final class ProfilesHeaderView: UIView {
                 height(groupHeaderHeight)
                 addGroupHeader.edgesToSuperview()
             } else {
-                height(groupHeaderHeight + (searchBar?.frame.height ?? 0))
-                addSearchBar(searchBar, bottomPinnedTo: addGroupHeader)
+                frame = CGRect(origin: .zero, size: CGSize(width: (searchBar?.frame.size.width ?? 0), height: (searchBar?.frame.size.height ?? 0) + groupHeaderHeight))
+                addSearchBar(searchBar)
```
yuliaveres: in general we should not mix frames with layout.. but since stuff is broken on iOS 10 and i guess u tried to handle it with just layout, we should go with this solution 😿

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
```diff
@@ -53,26 +57,25 @@ final class ProfilesHeaderView: UIView {
...
-        addSubview(searchBar)
+        searchBar.frame.origin = .zero
         
-        if let viewToPinTo = view {
-            searchBar.edgesToSuperview(excluding: .bottom)
-            searchBar.bottomToTop(of: viewToPinTo)
-        } else {
-            searchBar.edgesToSuperview()
-        }
+        // Note: Autolayout does not play nicely with the search bar.
```
yuliaveres: 😢

Toshi/Models/Cereal.swift
```diff
@@ -107,11 +107,11 @@ class Cereal: NSObject {
             mnemonic = BTCMnemonic(entropy: entropy, password: nil, wordListType: BTCMnemonicWordListType.english)!
         }
 
-        let idKeychain = Cereal.idKeychain(from: mnemonic)
+        let idKeychain: BTCKeychain = Cereal.idKeychain(from: mnemonic)
```
yuliaveres: why u needed to specify it ?

yuliaveres: and how is it related to the PR ? 😛

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -42,7 +43,7 @@ class BasicTableViewCell: UITableViewCell {
     static let doubleImageMargin: CGFloat = 16.0
     static let largeVerticalMargin: CGFloat = 22.0
 
-    var actionDelegate: BasicCellActionDelegate?
+    weak var actionDelegate: BasicCellActionDelegate?
```
yuliaveres: oh good catch again

Toshi/Controllers/Profiles/ProfilesAddGroupHeader.swift
designatednerd: yes that is the name of the icon in the Sketch file

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
designatednerd: whoops, nope

Toshi/Controllers/Profiles/ProfilesAddedToGroupHeader.swift
designatednerd: The margin gets set as part of this initializer. I just have it as zero above so it doesn't need to be optional.

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
designatednerd: Yeah basically the long and short of it is that table header views and autolayout haaaate each other on iOS 10. This is way, way better on 11.

Toshi/Models/Cereal.swift
designatednerd: giving hints to the compiler so it stops warning about this, that's all.

Toshi/Views/Cells/BasicTableViewCell.swift
designatednerd: that was SwiftLint, not me 😄

Toshi/Controllers/Profiles/ProfilesAddedToGroupHeader.swift
yuliaveres: do we even need this margin not 0 ? just asking if we can get rid of this

PR #642, opened 2017-12-11
Use toshi id to sign all requests to eth service

tristan:
Currently the wallet private key is being used to sign eth service requests.
Toshi id private key should be used for all http request signing (see PN registration was using the correct key but deregistration was using the wrong one 🙈 ).


Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -129,11 +129,11 @@ class EthereumAPIClient: NSObject {
                 return
             }
 
-            let hashedPayload = cereal.sha3WithWallet(string: payloadString)
-            let signature = "0x\(cereal.signWithWallet(message: "POST\n\(path)\n\(timestamp)\n\(hashedPayload)"))"
+            let hashedPayload = cereal.sha3WithID(string: payloadString)
```
Elland: 🦅👀

PR #643, opened 2017-12-11
[IOS-422]: Retrieve prekeys, regester user with chat service if 401 returned

yuliaveres:
Also:
remove not needed key from UserDefaults wrapper;
do not store network requests password in database, create new one each login;
refresh prekeys if there are less then desirable amount left (defines by constant of `20`, same as used in bots code).

### DIFF
--- a/Toshi/Supporting Files/AppDelegate.m
+++ b/Toshi/Supporting Files/AppDelegate.m
@@ -18,12 +18,11 @@
 #import <SignalServiceKit/OWSDispatch.h>
 #import <SignalServiceKit/ProfileManagerProtocol.h>
 #import "ProfileManager.h"
-#import "PrekeysRequest.h"
 
 #import <AxolotlKit/SessionCipher.h>
+#import "PreKeyHandler.h"
 
 NSString *const ChatSertificateName = @"token";
-NSUInteger const PREKEY_MINIMUM_COUNT = 20;
 
 @import WebRTC;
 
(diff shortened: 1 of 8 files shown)


Toshi/Supporting Files/AppDelegate.m
```diff
@@ -380,32 +382,58 @@ - (void)configureForCurrentSession
...
-                                                         [CrashlyticsLogger log:@"Successfully registered user with chat service after forced trigger" attributes:nil];
-                                                     } else {
-                                                         [CrashlyticsLogger log:@"Failed to register user with chat service after forced trigger" attributes:nil];
-                                                     }
-                                                 }];
-                                             }
-                                         }];
-    }
+    NSURL *url = [NSURL URLWithString:[NSString stringWithFormat:@"%@",
+                                       textSecureKeysAPI]];
```
designatednerd: if this is a string you should just be able to pass it in directly without the `stringWithFormat`

Toshi/Supporting Files/AppDelegate.m
```diff
@@ -380,32 +382,57 @@ - (void)configureForCurrentSession
...
+                                         if (error.code == 401) {
+                                             [ChatAPIClient.shared registerUserWithCompletion:^(BOOL success) {
+                                                 if (success) {
+                                                     [CrashlyticsLogger log:@"Successfully registered user with chat service after forced trigger" attributes:nil];
+                                                 } else {
+                                                     [CrashlyticsLogger log:@"Failed to register user with chat service after forced trigger" attributes:nil];
+                                                 }
+                                             }];
+                                         }
+                                     }];
```
designatednerd: Is it worth pulling this out into some kind of separate service for testing (or at least getting it out of the app delegate)?

yuliaveres: i wont do it now. so won;t close pr until i do

Toshi/Supporting Files/PreKeyHandler.m
```diff
@@ -0,0 +1,78 @@
...
+#import "PrekeysRequest.h"
+#import <SignalServiceKit/TSNetworkManager.h>
+#import <SignalServiceKit/TSPreKeyManager.h>
+#import "Toshi-Swift.h"
+
+NSUInteger const PREKEY_MINIMUM_COUNT = 20;
+
+@implementation PreKeyHandler
+
++ (void)tryRetrievingPrekeys
```
designatednerd: is there any way to test any of this?

yuliaveres: there is always way to test this. we might be back to it later when there is more time

designatednerd: but what if by adding tests now we prevent more mistakes with this and save ourselves even more time? 🤔 🤔 🤔

PR #644, opened 2017-12-12
[IOS-424]: Leave group support

yuliaveres:
So we agreed with Marek to use Leave instead of Exit:
img_0024
img_0025
img_0027


Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -79,8 +79,15 @@ final class ChatInteractor: NSObject {
 
         messageSender?.sendAttachmentData(datasource, contentType: "image/jpeg", sourceFilename: "image.jpeg", in: outgoingMessage, success: {
```
designatednerd: Good grief, that's obnoxious that this doesn't come back on the main thread by default 😠

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -360,13 +367,51 @@ final class ChatInteractor: NSObject {
...
-                interactor.send(image: groupAvatar, in: outgoingMessage)
+                interactor.send(image: groupAvatar, in: outgoingMessage, completion: completion)
             } else {
-                interactor.send(outgoingMessage)
+                interactor.send(outgoingMessage, completion: completion)
             }
         }
     }
 
+    static func deleteThread(_ thread: TSThread) {
```
designatednerd: What about making this an extension method on `TSThread`? Might help us get some of the database logic out of the VC

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
```diff
@@ -51,11 +59,23 @@ final class GroupInfoViewModel {
...
+        models = [avatarTitleSectionData, settingsSectionData, participantsSectionData, exitGroupSectionData]
+    }
+
+    private func setupParticipantsSection() -> TableSectionData {
         let addParticipantsData = TableCellData(title: Localized("new_group_add_participants_action_title"))
         addParticipantsData.tag = GroupItemType.addParticipant.rawValue
 
         var participantsCellData: [TableCellData] = [addParticipantsData]
-        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { fatalError("Can't access app delegate") }
```
designatednerd: should this crash the app at runtime even in production? And if it should, should we fire a log off to Crashlytics before we do?

Toshi/Resources/Localizable.strings
```diff
@@ -266,6 +266,11 @@
 "new_group_public_settings_title" = "Public";
 "new_group_notifications_settings_title" = "Notifications";
 
+"group_info_exit_group" = "Exit group";
+"group_info_exit_confirmation_message" = "Do you want to exit group?";
```
designatednerd: Maybe "Do you want to exit **this** group?" or "Do you want to exit [group name]?" - this seems worded a little oddly.

Toshi/Resources/Localizable.strings
```diff
@@ -266,6 +266,11 @@
...
+"group_info_exit_action_title" = "Exit";
+"group_info_leave_group_failure_message" = "Failed to leave group";
```
designatednerd: Is there anything the user can do once they've failed to leave the group?

Toshi/Controllers/GroupChat/GroupViewModelProtocol.swift
```diff
@@ -28,6 +29,8 @@ struct GroupInfo {
 protocol GroupViewModelCompleteActionDelegate: class {
 
     func groupViewModelDidFinishCreateOrUpdate()
+    func groupViewModelDidStartCreateOrUpdate()
+    func groupViewModelaDidRequireReload(_ viewModel: GroupViewModelProtocol)
```
designatednerd: looks like a spare `a` snuck in before `Did`

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: lets log to Crashlytics ! (it should)

Toshi/Controllers/Chat/ChatInteractor.swift
yuliaveres: 🤔

Toshi/Resources/Localizable.strings
yuliaveres: this group ..  i will check with Marek right now

Toshi/Controllers/Chat/ChatInteractor.swift
yuliaveres: that is not controller though. i try to keep all messages / group related logic just in the interactor

designatednerd: Oh yeah derp, my bad.

Toshi/Resources/Localizable.strings
yuliaveres: i do not think so. I would guess dismiss the alert and go back or press exit once more ..

designatednerd: maybe add something along the lines of "Please try again" then

yuliaveres: ok. lets do this

PR #646, opened 2017-12-12
Status cells for chat

marijnschilling:
Issue: https://toshiapp.atlassian.net/browse/IOS-421

screenshot 2017-12-12 16 15 21

So basically what did is added a SOFAWrapper type for the Status messages, those status messages still have to be created from SignalMessages that give a `QUIT` (leave) or `UPDATE` (other changes to the group). 

The SofaStatus can be created with a content string just as the other SofaWrappers. I've added tests for this. Right now it's created with 'Strings' for the names, this will probably change to userID's or something... but we can do that later when we actually connect the signal messages.

This SofaWrapper has a `statusType` and an `object`. So for example if the `statusType` is `.leave` and the `object` is 'Robert' we'll need a message saying `Robert left the group` Some status types also use the `subject` var. For example `.added` with object `Daniel` and subject `Robert` should be `Daniel added Robert to the group`. Or status type `.rename` with object `Daniel` and subject `Hamsters` should be `Daniel renamed the group to Hamsters`.
- [ ] I think this might be a bit unclear though, and i'm thinking about just adding more optional var's for `newGroupname` or `addedGroupMember` that are only used for specific `statusType`'s. This will clutter the code a bit though, so not sure. What do you think?

So the `SofaStatus` also created this attributed string. I've added tests for this. 

I've only added one of the possible status messages so far, but the logistics for the others in in place.

So the next step would be implementing the other status messages. I will continue with that.

Another next step is compressing the statusMessages. So for example when 6 people are added to the group it should fold into one message saying: `Robert added you, Robert-Hein, Daniel, Simone, Adam, Andrew, Crystal, Simon Pegg, Marijn Schilling, Sam Wood, Diana Drew and Tristan King to B&B Amsterdam.` and maybe even just says `Robet added 6 people to the group` that is tapable and expands into the first message.
This is a lot of work though. We should make something like a `StatusMessageParser` that acts upon the `mesageModels` in the `ChatViewModel` and analyses wether there are status messages that can be compressed together. And then me should create new `StatusType`'s for these compressed messages ánd expandable states. And that keeps trach of wether we should show them expanded or not.

I think it would be smart to wait with this until a later stage maybe. I think group chat can already go live without these compressed status messages.


Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -446,10 +448,35 @@ extension ChatViewController: UITableViewDataSource {
     }
 
     func tableView(_: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        var cell = tableView.dequeueReusableCell(withIdentifier: UITableViewCell.reuseIdentifier, for: indexPath)
 
-        let message = viewModel.messageModels[indexPath.item]
+        //WARNING: Don't merge this! It's commented for testing purposes only!
```
marijnschilling: This should not be merged! I just create a Status message here and overwrite the actual message models.

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        if let statusJSON = json["type"] as? String {
+            statusType = StatusType(rawValue: statusJSON) ?? .none
+        }
+
+        return statusType
+    }
+    // The types of the object and subjects of the action will probably be user id's int he end
+    var subject: String {
+        return json["subject"] as? String ?? ""
```
marijnschilling: I don't really know how to handle this optionality here. it shouldnt be an empty string when it isn't here, it should just be here!

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,91 @@
...
+    var object: String? {
+        return json["object"] as? String
+    }
+
+    var attributedText: NSAttributedString? {
+       switch statusType {
+       case .leave:
+        break
+       case .added:
+           if let object = object {
```
marijnschilling: Same here. I the status type is `.added` subject should not be optional. And this if else statement doesnt make sense at all. It should never say `Robert added to the group` 🙈

Tests/SofaStatusTest.swift
```diff
@@ -0,0 +1,74 @@
...
+                expect(status.object).to(equal("Marek"))
+            }
+
+            it("creates an attributed string with bold parts for the names") {
+                guard let attributedText = status.attributedText else {
+                    XCTFail("no attributed text set on sofa status")
+                    return
+                }
+
+                let fontAttribute = [NSAttributedStringKey.font: Theme.preferredFootnote()]
```
designatednerd: is this used?

Tests/SofaStatusTest.swift
```diff
@@ -0,0 +1,74 @@
...
+                expect(status.object).to(equal("New group name"))
+            }
+
+            it("creates an attributed string with bold parts for the name and the new group name") {
+                guard let attributedText = status.attributedText else {
+                    XCTFail("no attributed text set on sofa status")
+                    return
+                }
+
+                let fontAttribute = [NSAttributedStringKey.font: Theme.preferredFootnote()]
```
designatednerd: again, is this used?

Tests/SofaStatusTest.swift
```diff
@@ -0,0 +1,74 @@
...
+
+                let robertRange = attributedText.string.nsRange(forSubstring: "Robert")
+                let boldAttribute = attributedText.attributes(at: robertRange.location, effectiveRange: nil)
+                expect(boldAttribute[NSAttributedStringKey.font] as? UIFont).to(equal(Theme.bold(size: 13)))
```
designatednerd: or is this where you wanted to use `fontAttribute`? Seems like it'd be really easy to break this test without breaking the actual functionality if the the font being used for bolding ever changes

Tests/String+nsRangeTests.swift
```diff
@@ -0,0 +1,39 @@
+import XCTest
+import UIKit
+import Quick
+import Nimble
+@testable import Toshi
+
+class String_nsRangeTests: QuickSpec {
```
designatednerd: ❤️ ❤️ ❤️ Tests! ❤️ ❤️ ❤️

Toshi/Controllers/Chat/ChatViewController.swift
designatednerd: You should probably keep this PR classified as WIP until this is removed then

Toshi/MessagesUI/Cells/StatusCell.swift
```diff
@@ -0,0 +1,45 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+class StatusCell: UITableViewCell {
+
+    static let reuseIdentifier = "StatusCell"
```
designatednerd: Given that `SweetUIKit` has the ability to dequeue without the reuse identifier, do we need to be setting these?

Toshi/MessagesUI/Cells/StatusCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        setTextLabelStyle()
```
designatednerd: This needs to be re-set to support dynamic type, right? Might help to throw that in as a comment

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    override var type: SofaType {
+        return .status
+    }
+
+    var statusType: StatusType {
+        var statusType = StatusType.none
+
+        if let statusJSON = json["type"] as? String {
```
yuliaveres: do we need to have it computed ? 
If no lets set status type in `init` once ?

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,99 @@
...
+            statusType = StatusType(rawValue: statusJSON) ?? .none
+        }
+
+        return statusType
+    }
+    // The types of the object and subjects of the action will probably be user id's int he end
+    var subject: String? {
```
yuliaveres: same regarding moving set of this `var` to `init` ?

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,99 @@
...
+        return json["subject"] as? String
+    }
+    
+    // object is optional since this is only used in added "Marek added Robert" (where Robert is the subject)
+    var object: String? {
```
yuliaveres: and here?

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,99 @@
...
+        return json["object"] as? String
+    }
+
+    var attributedText: NSAttributedString? {
+        guard let subject = subject else { return nil }
```
yuliaveres: i would not make it computed until u have a reason for.. 
we might want to adjust `attributedText` as soon as we have `statusType` ( in `didSet {}` ? )

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,99 @@
...
+            guard let object = object else { return nil }
+            return attributedStringFor(Localized("status_type_rename"), with: [subject, object])
+        case .setToPublic:
+            return attributedStringFor(Localized("status_type_make_public"), with: [subject])
+        case .setToPrivate:
+            return attributedStringFor(Localized("status_type_make_private"), with: [subject])
+        default:
+            return nil
+        }
+
```
yuliaveres: 🙈

Toshi/Models/SofaWrappers/SofaStatus.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private func attributedStringFor(_ string: String, with boldStrings: [String]) -> NSAttributedString {
```
yuliaveres: the one i am going to ask now is not really important, but if u have a minute could u please use label for function instead of attaching it to the name.  sometimes we need to do in this ☝️  way but here i dnt see any reason ..
`private func attributedStringFor(_ string: String, with boldStrings: [String]) -> NSAttributedString` vs
`private func attributedString(for string: String, with boldStrings: [String]) -> NSAttributedString`.

Toshi/Models/SofaWrappers/SofaWrapper.swift
```diff
@@ -88,6 +92,8 @@ class SofaWrapper: SofaWrapperProtocol {
             return SofaPaymentRequest(content: content)
         case .payment:
             return SofaPayment(content: content)
+        case .status:
+            return SofaPayment(content: content)
```
yuliaveres: i guess `SofaStatus` ?

yuliaveres:
that can go directly to master, but lets keep things clear and have it along group chat for now? (`feature/group-chat` as base branch)

yuliaveres:
i assume the `IOS-421` is ticket to be attached to this PR ? if yes, please do 👍  👍

Tests/SofaStatusTest.swift
marijnschilling: Ah yeah, forgot to update!

Tests/SofaStatusTest.swift
marijnschilling: nope :) old copied stuff :)

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -446,10 +448,35 @@ extension ChatViewController: UITableViewDataSource {
...
+    }
+
+    private func dequeueStatusCell(message: MessageModel, indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: StatusCell.reuseIdentifier, for: indexPath)
+        cell.textLabel?.attributedText = message.attributedText
+
+        return cell
+    }
+
+    private func dequeueMessageBasicCell(message: MessageModel, indexPath: IndexPath) -> UITableViewCell {
```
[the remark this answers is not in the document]
marijnschilling: since at `cellForRowAtIndexPath` it will assign in to cell which is a `UITableViewCell` But i guess it doesnt really matter. And it makes more sense for this message to return a `MessagesBasicCell`. I will adjust!

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -463,25 +465,42 @@ extension ChatViewController: UITableViewDataSource {
     }
 
     func tableView(_: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        var cell = tableView.dequeueReusableCell(withIdentifier: UITableViewCell.reuseIdentifier, for: indexPath)
```
designatednerd: You should be able to just use `tableView.dequeue(UITableViewCell.self, for: indexPath)` with SweetUIKit

designatednerd: also you could just `let cell: UITableViewCell` since it's assigned in both the if and else cases below

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -463,25 +465,42 @@ extension ChatViewController: UITableViewDataSource {
...
+    private func dequeueStatusCell(message: MessageModel, indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: StatusCell.reuseIdentifier, for: indexPath)
+        cell.textLabel?.attributedText = message.attributedText
 
-            updateMessageState(messageModel, in: cell)
+        return cell
+    }
+
+    private func dequeueMessageBasicCell(message: MessageModel, indexPath: IndexPath) -> MessagesBasicCell {
+        guard let cell = tableView.dequeueReusableCell(withIdentifier: message.reuseIdentifier, for: indexPath) as? MessagesBasicCell else { fatalError("Couldn't deqeueu MessagesBasicCell")}
```
designatednerd: again with Sweet UIKit sugar I think you can do this as `tableView.dequeue(MessagesBasicCell.self, for: indexPath)` and skip having to do the guard statement

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -463,25 +465,42 @@ extension ChatViewController: UITableViewDataSource {
...
+        cell.transform = self.tableView.transform
+
+        return cell
+    }
 
-            cell.isOutGoing = messageModel.isOutgoing
-            cell.positionType = positionType(for: indexPath)
-            cell.delegate = self
+    private func dequeueStatusCell(message: MessageModel, indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: StatusCell.reuseIdentifier, for: indexPath)
```
designatednerd: also here with the sweet UIKit helper

PR #648, opened 2017-12-13
Prevent matching usernames inside URLs with @ signs.

Elland:


Toshi/MessagesUI/Cells/MessagesTextCell.swift
```diff
@@ -99,18 +99,31 @@ class MessagesTextCell: MessagesBasicCell {
             let range = NSRange(location: 0, length: text.string.utf16.count)
             text.addAttributes([.kern: -0.4], range: range)
 
+            // Do a link detector first-pass, to avoid creating username links inside URLs that contain an @ sign.
+            let linkDetector = try? NSDataDetector(types: NSTextCheckingResult.CheckingType.link.rawValue)
```
designatednerd: can you not just pass in `[.link]`?

Elland: It should work, but it doesn't. 🤦🏽‍♂️

designatednerd: :sad-trombone:

Toshi/MessagesUI/Cells/MessagesTextCell.swift
```diff
@@ -86,47 +84,4 @@ class MessagesTextCell: MessagesBasicCell {
         textView.adjustsFontForContentSizeCategory = true
         textView.text = nil
     }
-
-    private func detectUsernameLinksIfNeeded() {
-        guard frame.isEmpty == false else { return }
```
designatednerd: does removing this rather than just putting the text detector in after the `if let` statement cause side effects?

Elland: This is a very odd guard, and since there's no comment, it's hard to understand why it was there to begin with… Any insights appreciated, but so far I couldn't see anything. Only case I can see frame being `0` also requires the text to be `nil`.

designatednerd: I've got no idea, but that's part of why my suggestion is to only isolate the bits that are actively changing the text - basically, this is a place where you can pull out a little chunk, test it, but not really change the way stuff individual methods are called.

It's piecemeal, but I think you're probably finding there's a biiiiig rabbit hole here once you start messing around with things that you're not sure why they're there 😄

Elland: I stopped myself from rewriting the whole cell, so that's an improvement. 👍 But I did end up removing the method and cleaning up the `textView` interactions a bit.

PR #650, opened 2017-12-14
[WIP] Chat preview

marijnschilling:
Issue: https://toshiapp.atlassian.net/browse/IOS-431

accept

1. I created `PreviewChatViewController` a subclass of the `ChatViewController` where only the tableview is added but the cells aren't interactive, and I introduced the approve and decline buttons on this view. I did this cause i dint want to mess around with different states in the `ChatViewController` itself because if this is not handled super carefully it could easily lead to features being disabled in the `ChatViewContoller` itself, and we wouldn't want that now would we?

2. When the user approves the chat i do a little trick though: I insert the real `ChatVieController` underneath before I pop the `PreviewChatViewController`. We can add a custom transition animation to this later to make it look sassy.

3. I extracted these sassy rounded-cornered ChatButtons into a UIButton so we can use it here and in the Chat buttons. 

For now it works, but it's not nice yet. I plan to do that in a branch from this branch, so it won't block further development of this flow.

In that branch I will address the folowing issues:
- [ ] Add the ❌  and ✔️  to the buttons. Thereby also making their use as approve/decline buttons and as Chatbuttons a bit easier and straightforwarder.
- [ ] Probably give the Approve/decline view a blurry blackground (will approve with marek)
- [ ] make the transition to the real ChatViewController nicer.


Toshi/Controllers/Chat/ChatButton.swift
```diff
@@ -0,0 +1,82 @@
+import UIKit
+
+class ChatButton: UIButton {
```
yuliaveres: `final` ?

Toshi/Controllers/Chat/ChatButtonsCollectionViewCell.swift
```diff
@@ -5,84 +5,39 @@ import TinyConstraints
 final class ChatButtonsViewCell: UICollectionViewCell {
     
     static let reuseIdentifier = "ChatButtonsViewCell"
-    
+
     var title: String? {
         didSet {
-            guard let title = title else {
```
yuliaveres: perfect 👍

Toshi/Controllers/Chat/PreviewChatViewController.swift
```diff
@@ -0,0 +1,72 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+final class PreviewChatViewController: ChatViewController {
```
yuliaveres: Wondering regarding the decision of making everything disabled from interactions. Does it come from design?
If yes:
lets think if there is a reason to subclass a controller for the sake of disabling cell ? ..
wdyt ?

I would avoid this. This controller does not have less responsibility or logic. it takes everth from his parent, the only really different what he does is overriding `cellForRow` ?
Would not that be fair enough to add top transparent disabled from interaction view on top of chat controller if thread is `pendingAccept` ? and those bottom buttons view?

Toshi/Views/AcceptDeclineButtonsView.swift
```diff
@@ -0,0 +1,77 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+protocol AcceptDeclineButtonsViewDelegate: class {
+    func didSelectAccept()
+    func didSelectDecline()
+}
+
+class AcceptDeclineButtonsView: UIView {
```
yuliaveres: `final` ?

Toshi/Controllers/Chat/PreviewChatViewController.swift
marijnschilling: Yes, disabling interactions is from design (in the issue you can read what the specs are)

It's not only disabling the cells in the subclass it's also not adding a lot of views, balance, textinput, network view

I don't want to add this preview state handling in the chatview controller because it will have impact in a lot of different places and the consequenses when handled wrong are too big. 

When it's in a subclass it's way easier too maintain the separate state without fear of getting them mixed up.

yuliaveres: hmm, what is your plan for moving to interactive chat after accept ? 🤔

marijnschilling: Have you read my explanation? What is not clear about it?

yuliaveres: yes, i have. i am still wondering if the transition tricks as u call it are better than having those views on a regular chat controller ("a lot of views, balance, textinput, network view") ?

marijnschilling: What do you mean `having those views on a regular chat controller ("a lot of views, balance, textinput, network view") ?`

yuliaveres: referring to one of the the reasons you mention:
"It's not only disabling the cells in the subclass it's also not adding a lot of views, balance, textinput, network view"

Toshi/Controllers/Chat/PreviewChatViewController.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    private lazy var acceptDeclineButtons: AcceptDeclineButtonsView = {
+        let acceptDeclineButtons = AcceptDeclineButtonsView()
+        acceptDeclineButtons.delegate = self
+
+        return acceptDeclineButtons
+    }()
+
+    override func viewDidLoad() {
+        view.backgroundColor = Theme.viewBackgroundColor
```
yuliaveres: u should always call `super.viewDidLoad()`

Toshi/Controllers/Chat/ChatButtonsCollectionViewCell.swift
```diff
@@ -5,84 +5,39 @@ import TinyConstraints
...
     }
-    
+
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-    
+
     override func prepareForReuse() {
         super.prepareForReuse()
```
yuliaveres: i guess we should add some relayout. what might happen is:
but that happens only after approval...
img_0047

marijnschilling: 🙈  Oh no, I'll check if i can reproduce this bug!

PR #652, opened 2017-12-15
[IOS-425]: Thread requests

yuliaveres:
Adding actually approve / decline actions and seems all we need.
img_0029
img_0030
img_0031
screenshot at dec 15 11-32-55


Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+    private var idAPIClient: IDAPIClient {
+        return IDAPIClient.shared
+    }
+
+    override init(style: UITableViewStyle) {
+        super.init(style: style)
+
+        title = dataSource.title
+
+        loadViewIfNeeded()
```
designatednerd: why does this need to be called here - wouldn't this always be needed in an init?

yuliaveres: @designatednerd  i would say the cleanup part is not done yet , which will touch things like this.

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+        cell.actionDelegate = self
+
+        return cell
+    }
+}
+
+extension MessagesRequestsViewController: BasicCellActionDelegate {
+
+    func didTapFirstActionButton(_ cell: BasicTableViewCell) {
+        // accept thread at given indexpath
```
designatednerd: are these `TODO` to get done in the next PR?

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+        preferLargeTitleIfPossible(true)
+        tabBarController?.tabBar.isHidden = false
+    }
+
+    private func recipient(for thread: TSThread) -> TokenUser? {
+        guard let recipientAddress = thread.contactIdentifier() else { return nil }
+
+        var recipient: TokenUser?
+
+        if let userData = (Yap.sharedInstance.retrieveObject(for: recipientAddress, in: ThreadsDataSource.nonContactsCollectionKey) as? Data),
```
designatednerd: would it be helpful to try to isolate some of this logic in the data source so you're not actively accessing the database within the view controller?

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+        navigationController?.pushViewController(chatViewController, animated: true)
+    }
+
+    func tableView(_: UITableView, editActionsForRowAt indexPath: IndexPath) -> [UITableViewRowAction]? {
+        let action = UITableViewRowAction(style: .destructive, title: "Delete") { _, indexPath in
+            if let thread = self.dataSource.acceptedThread(at: indexPath) {
+
+                TSStorageManager.shared().dbReadWriteConnection?.asyncReadWrite { transaction in
+                    thread.remove(with: transaction)
+                }
```
designatednerd: again this seems like it might be something better handled by the data source since it's directly interacting with the database

yuliaveres: @designatednerd  please i have not cleaned this up yet. i will let u know when u can start review

Toshi/Controllers/Recent/RecentViewController.swift
```diff
@@ -83,25 +53,26 @@ class RecentViewController: SweetTableController, Emptiable {
 
         addSubviewsAndConstraints()
 
-        tableView.dataSource = self
         tableView.delegate = self
+        tableView.dataSource = self
+
         tableView.register(ChatCell.self)
+        BasicTableViewCell.register(in: tableView)
```
designatednerd: why are these set up in different orders? seems like internal consistency would be nice if there's not a specific reason for it

Toshi/Controllers/Recent/RecentViewController.swift
```diff
@@ -324,15 +250,24 @@ extension RecentViewController: UITableViewDelegate {
     }
 
     func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
-        if let thread = self.thread(at: indexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
+        switch indexPath.section {
+        case 0:
```
designatednerd: is there an enum or other constant you can use here which can better describe which section this is for?

Toshi/Controllers/Recent/RecentViewModel.swift
```diff
@@ -0,0 +1,131 @@
...
+        let filteringBlock: YapDatabaseViewFilteringWithObjectBlock = { transaction, group, collection, key, object in
+            guard let thread = object as? TSThread else { return true }
+            return thread.isPendingAccept == true
+        }
+
+        return YapDatabaseViewFiltering.withObjectBlock(filteringBlock)
+    }
+
+    private func threadsSorting() -> YapDatabaseViewSorting {
+        let viewSorting = YapDatabaseViewSorting.withObjectBlock { (_, _, _, _, _, _, _, _) -> ComparisonResult in
```
designatednerd: I'm gonna call this the "Crew rowing operator" from now on because it reminds me of: 
sportssergio-acuna-garcia-982x654

Toshi/Controllers/Recent/ThreadCellConfigurator.swift
```diff
@@ -0,0 +1,99 @@
...
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.lineSpacing = 2.5
+        paragraphStyle.paragraphSpacing = -4
+        paragraphStyle.lineBreakMode = .byTruncatingTail
+
+        return [
+            .kern: -0.4,
+            .font: Theme.preferredRegularSmall(),
+            .foregroundColor: Theme.greyTextColor,
+            .paragraphStyle: paragraphStyle
```
designatednerd: Is this some kind of style that we should be defining in the `Theme`?

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -0,0 +1,253 @@
...
+        case .unacceptedThreadRequests:
+            return Localized("messages_requests_title")
+        }
+    }
+
+    func title(for section: Int) -> String? {
+        guard self == .recent else { return nil }
+
+        switch section {
+        case 1:
```
designatednerd: If you make this an `Int` enum, you should be able to `switch self` and then return the appropriate title there without having to hard code the section index. 

Also: See `CountableIntEnum` for more fun party tricks

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
designatednerd: D'oh! Sorry

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
yuliaveres: yea

Toshi/Controllers/Recent/ThreadCellConfigurator.swift
yuliaveres: we do not have those defined and agreed on yet

Pods/SignalServiceKit/SignalServiceKit/src/Messages/Interactions/TSOutgoingMessage.m
```diff
@@ -485,9 +485,6 @@ - (OWSSignalServiceProtosDataMessageBuilder *)dataMessageBuilder
                 [groupBuilder setType:OWSSignalServiceProtosGroupContextTypeUpdate];
                 break;
             }
-            case TSGroupMetaMessageRequestInfo:
-                [groupBuilder setType:OWSSignalServiceProtosGroupContextTypeRequestInfo];
-                break;
```
designatednerd: I'm not sure what's happening here, but this probably shouldn't be re-deleted

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    private lazy var dataSource: ThreadsDataSource = {
+        let dataSource = ThreadsDataSource(target: .unacceptedThreadRequests)
+        dataSource.output = self
+
+        return dataSource
+    }()
+
+    private var idAPIClient: IDAPIClient {
+        return IDAPIClient.shared
```
designatednerd: why not do this as a `let` and just assign the shared instance?

Toshi/Views/ChatCell.swift
```diff
@@ -1,293 +0,0 @@
...
-//
-// You should have received a copy of the GNU General Public License
-// along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-import UIKit
-import SweetUIKit
-import SweetSwift
-import SweetFoundation
-
-class ChatCell: UITableViewCell {
```
designatednerd: Did you intend to delete this? I don't see any place where usage of this class has been removed, just the deletion of the class.

yuliaveres: https://github.com/toshiapp/toshi-ios-client/pull/652/files#diff-5d41b94642688ea191300fd86b0ac469L351

designatednerd: aha! thank you :)

Toshi/Resources/Localizable.strings
```diff
@@ -28,6 +28,10 @@
 "messages_sent_error_action_delete" = "Delete";
 "messages_sent_error_action_resend" = "Resend";
 
+"recent_messages_section_header_title" = "Messages";
+"messages_requests_title" = "Message Requests";
+"message_requests_description" = "Some people want to start a chat";
```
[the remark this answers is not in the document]
yuliaveres: basically we need to have a number of people here. will change now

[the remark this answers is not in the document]
designatednerd: spoke to Yuliia and I'm gonna steal this one and make a plurals plist in another branch.

Toshi/Controllers/Recent/RecentViewController.swift
```diff
@@ -125,179 +97,137 @@ class RecentViewController: SweetTableController, Emptiable {
...
+
+        emptyView.isHidden = shouldHideEmptyState
     }
 
-    @objc func yapDatabaseDidChange(notification _: NSNotification) {
-        let notifications = uiDatabaseConnection.beginLongLivedReadTransaction()
+    private func messagesRequestsCell(for indexPath: IndexPath) -> UITableViewCell {
+        let cellConfigurator = CellConfigurator()
+        var cellData: TableCellData
+        var cell = UITableViewCell(frame: .zero)
```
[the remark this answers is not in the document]
yuliaveres: cause the return call is nonnull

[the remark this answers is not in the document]
designatednerd: ok. I'm gonna clean this up a little in my PR 😄

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,157 @@
...
+
+extension MessagesRequestsViewController: ThreadsDataSourceOutput {
+
+    func threadsDataSourceDidLoad() {
+        tableView.reloadData()
+        dismissIfNeeded()
+    }
+}
+
+extension MessagesRequestsViewController: UITableViewDelegate {
```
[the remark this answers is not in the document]
yuliaveres: regarding this, agreed- interesting to try, but this does not really fit my case here as all selections and display are quite quite different in requests and recent. those controllers handle table view differently and if i wanted to manage `tableViewDatasource` and `delegate` in data source object, i would prefer having separate datasources in that case. while having single datasource dealing with collections and database access for both controller has more value here as for me. and dealing with table view in each controller in a way it needs. not sure if u recall - we had discussion about that when i started that. :)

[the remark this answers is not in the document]
designatednerd: yeah although i think maybe i understood that conversation differently. This is good for now though.

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+}
+
+extension MessagesRequestsViewController: UITableViewDataSource {
+
+    func tableView(_: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return dataSource.unacceptedThreadsCount
+    }
+
+    func
+        tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
```
designatednerd: alignment is off here

PR #653, opened 2017-12-15
Chat preview by managing state in chatController

marijnschilling:
Instead of using a subclass I now manage state within the `ChatViewController`

Animation form preview to real chat controller goes like this:
accept animation

We can merge this into #650 and then I continue with my fancying-everything-up branch (to add the buttons)


Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -97,12 +105,49 @@ class ChatViewController: UIViewController, UINavigationControllerDelegate {
...
+        hiddenButtonsViewBottomConstraint?.isActive = false
+        buttonsViewBottomConstraint?.isActive = true
+
+        hiddenEthereumPromptViewHeightConstraint?.isActive = false
+        ethereumPromptViewHeightConstraint?.isActive = true
+
+        UIView.animate(withDuration: 0.2, animations: {
+            self.buttonsView.alpha = 1.0
+            self.previewButtonsView.alpha = 0.0
+            self.view.setNeedsLayout()
```
yuliaveres: the only thing i see here is that the `view.setNeedsLayout()` call should not be in animation block. 
Afaik, `setNeedsLayout` might be called after you modify constraints (if u need) , and inside animation `layoutIfNeeded()`.

PR #654, opened 2017-12-18
Small fixes for group chat

yuliaveres:
We need to reset filter view on favourites and New group targeted profiles picker.
Also current user is displayed as a group member, thus when tapping on it nothing happened cause we were filtering only contacts without taking into consideration current user. We need to count him/her as well.


Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -144,7 +144,7 @@ final class ProfilesViewController: UITableViewController, Emptiable {
         preferLargeTitleIfPossible(true)
         showOrHideEmptyState()
 
-        guard dataSource.type == .updateGroupChat else { return }
+        guard dataSource.type != .updateGroupChat else { return }
```
yuliaveres: we need to clear excluded profiles filtering in each case apart from `updateGroupChat`.

PR #656, opened 2017-12-18
iOS-426: Requesting group info

designatednerd:
I tested this by manually removing the group chat member IDs and having that force a refresh - worked just fine. 

Note that due to issues with pre-keys which @yuliaveres can explain better than I can, if you sign out or uninstall, you will not be able to test this. 

Also includes a little organization and cleanup of warnings.


Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -127,6 +127,8 @@ final class ChatViewController: UIViewController, UINavigationControllerDelegate
 
         title = thread.name()
         updateChatAvatar()
+        
+        requestGroupInfoIfNeeded()
```
yuliaveres: 🤔 
i guess here we process any updates coming from yap about given thread (if someone renamed , left , updated avatar etc. ). not sure if we might need requesting any missing info about it here ... maybe, but definitely not the only place.
lets say for some reason u might just open the chat and miss some info, maybe after relogin / out or whatever (avatar, members id . ? ) and at that point we want to request for it immediately. still wondering about the case of requesting in this methos ☝️ . . .

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -183,6 +185,16 @@ final class ChatViewController: UIViewController, UINavigationControllerDelegate
             avatarImageView.image = (thread as? TSGroupThread)?.groupModel.groupImage
         }
     }
+    
+    private func requestGroupInfoIfNeeded() {
+        guard let groupThread = thread as? TSGroupThread else { /* Not a group thread, doesn't matter. */ return }
+        
+        guard (groupThread.groupModel.groupMemberIds?.count ?? 0) <= 1 else { /* We already have group member IDs besides our own, not needed */ return }
```
yuliaveres: not only group members which can miss.. 🤔  anything apart from `groupId` in `TSGroupModel` if I am right .. i know  theoretically signal API allows creating a group without avatar but what is Toshi designed for is that neither a user nor a group can't miss avatar, means we always use placeholders. so i would model that to check if avatar is missing as well. in that case that should signal about missing info.

Toshi/Controllers/Chat/ChatViewController.swift
designatednerd: I also had it in `viewWillAppear` earlier for testing purposes - given what you're saying it could make sense to put it there as well.

yuliaveres: yeah probably in willAppear ..

Toshi/Extensions/TSGroupModel+Additions.swift
```diff
@@ -0,0 +1,31 @@
...
+
+import Foundation
+
+extension TSGroupModel {
+
+    // true if all data required to display the group is loaded, false if not.
+    var isFullyLoaded: Bool {
+        guard
+            (groupMemberIds?.count ?? 0) > 1, // There must be more than just the current user as a member of the group
+            groupImage != nil else { // Image must either be a real image or a placeholder
```
yuliaveres: what about `groupName ` ?

designatednerd: I thought group name could be nil - does it at least have to be an empty string?

yuliaveres: let me check how android handles this. at least on ios it is not possible at the stage of creation / editing.

yuliaveres: neither it is allowed on android. so missing name is a problem

Toshi/Controllers/Chat/ChatViewController.swift
yuliaveres: actually that should be more than enough to update only on open.. u cant lose group info after u opened a chat. moreover that update fires a lot of times if update came actually. so maybe it is overwhelming to even check if group info should be requested so many times here. let me know your thoughts.

designatednerd: 🤔 I suppose that makes sense. Will remove here.

yuliaveres:
not sure how u managed to get info message from other member and have your group updated. cause  i failed. i did clear name and avatar of a group, when opening chat group info was requested from my device and apparently received on other member device as push notification was received. so the results was - my group was not updated on the device from where i cleared data on purpose. 🤔

yuliaveres:
i have a feeling we need to implement response for group info request.

designatednerd:
So I cleared it in `requestGroupInfoIfNeeded` right before it hit that second guard statement - where were you clearing it?

yuliaveres:
i had my group created. then on second run i added clear data code right before chat is open. that code run, thread was save i made sure. so before next run i removed clear code, on Recent i had my saved thread showing without any name and image. when i opened the chat, info was sent from this device. it was received by other member as i seen push notification. i tries to re run app but still group stayed without name and image ...

yuliaveres:
@designatednerd i found the problem.
in `MessageManager` `handleGroupInfoRequest` is never called as by default outgoing message is of type `OWSSignalServiceProtosGroupContextTypeDeliver` instead of `OWSSignalServiceProtosGroupContextTypeRequestInfo` when needed.

so the missing part in signal side which we should add is in `TSOutgoingMessage.m 487`:
```
case TSGroupMetaMessageRequestInfo:
                [groupBuilder setType:OWSSignalServiceProtosGroupContextTypeRequestInfo];
                break;
```

then `handleGroupInfoRequest` is called on other member's device and group info is sent back (group membership check is handled on their side 👍  ) and my group is updated.

designatednerd:
@yuliaveres https://github.com/toshiapp/Signal-iOS/pull/5- once that's approved i'll merge it in here.

yuliaveres:
@designatednerd please merge this

PR #658, opened 2017-12-20
iOS-432: Implement dapp featuring and dapp profiles

designatednerd:
In this PR we've replaced the featured Apps section with a featured Dapps section (distributed apps vs. centralized apps). 

simulator screen shot - iphone 8 - 2017-12-20 at 13 20 34

simulator screen shot - iphone 8 - 2017-12-20 at 13 20 27

Tapping the "enter" button will take the user to the Dapp's webpage in a SOFA browser. 

Also: 
- Added a protocol so we can more easily display different data types in the browse section
- Fun with `Codable`


Toshi/Controllers/Browse/BrowseAllViewController.swift
```diff
@@ -98,8 +101,8 @@ class BrowseAllViewController: UITableViewController {
         let cell = tableView.dequeue(SearchResultCell.self, for: indexPath)
 
         if let item = searchResults.element(at: indexPath.row) {
-            cell.nameLabel.text = item.name
-            cell.usernameLabel.text = item.isApp ? item.about : item.username
+            cell.nameLabel.text = item.nameForBrowsing
```
yuliaveres: not important but why we use both Browsing / search ?

Toshi/Controllers/Browse/DappViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+//
+//  Created by Ellen Shapiro (Work) on 12/19/17.
+//  Copyright © 2017 Bakken&Baeck. All rights reserved.
+//
+
+import SweetUIKit
+import UIKit
+import TinyConstraints
+
+class DappViewController: UIViewController {
```
yuliaveres: `final` 🙏

Toshi/Controllers/Browse/BrowseAllViewController.swift
```diff
@@ -85,8 +85,11 @@ class BrowseAllViewController: UITableViewController {
     }
 
     override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
-        if let contact = searchResults.element(at: indexPath.row) {
+
+        if let contact = searchResults.element(at: indexPath.row) as? TokenUser {
```
yuliaveres: could u please do same for `BrowseSearchResultView` in `didSelectRow` ?

Toshi/Models/Users/Dapp.swift
```diff
@@ -0,0 +1,67 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+/// An individual Dapp
+class Dapp: Codable {
```
yuliaveres: `final`

Toshi/Models/Users/Dapp.swift
```diff
@@ -0,0 +1,67 @@
...
+        return false
+    }
+    
+    var rating: Float? {
+        return nil
+    }
+}
+
+// Convenience class for decoding an array of Dapps with the key "results"
+class DappResults: Codable {
```
yuliaveres: `final`

Toshi/APIClients/idAPIClient.swift
```diff
@@ -582,4 +584,46 @@ import Teapot
...
+                    completion([], nil)
+                    return
+                }
+    
+                let dappResults: DappResults
+                do {
+                    let jsonDecoder = JSONDecoder()
+                    dappResults = try jsonDecoder.decode(DappResults.self, from: data)
+                } catch {
+                    completion(nil, .invalidPayload)
```
yuliaveres: do we want to do this on main thread ?

designatednerd: Yes - it looks like most of the rest of the errors in this class also need to be dispatched on the main queue (was following pattern but thinking about it, it doesn't make sense). I'll do that in a separate pull but fix this one now

Toshi/Controllers/Browse/BrowseAllViewController.swift
designatednerd: everything being assigned to that is Users - right now there's not a way to search Dapps

Toshi/Controllers/Browse/BrowseAllViewController.swift
designatednerd: Fair question - I should probably update the name of this property

PR #664, opened 2017-12-21
Change the constant of the height constraint

marijnschilling:
I got a constraint crash when opening a Dapp-view

It seems like the ActionButton set it's on constraint, I find it a bit weird how it works so if you think rewriting it would be better just let me know.

By setting the constant of the heightConstraint instead of a (nother) height constraint directly it doesnt crash at least.

I also implemented an extension method for the switch between safe area and layoutguides in ios 10 and 11.

### DIFF
--- a/Toshi/Controllers/Browse/DappViewController.swift
+++ b/Toshi/Controllers/Browse/DappViewController.swift
@@ -77,8 +77,8 @@ final class DappViewController: UIViewController {
         stackView.addArrangedSubview(descriptionLabel)
         stackView.addArrangedSubview(urlLabel)
         stackView.addArrangedSubview(enterButton)
-        
-        enterButton.height(44)
+
+        enterButton.heightConstraint.constant = 44
         
         return stackView
     }()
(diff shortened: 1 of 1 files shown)


Toshi/Controllers/Browse/DappViewController.swift
```diff
@@ -104,13 +104,7 @@ final class DappViewController: UIViewController {
...
         let topAnchor: NSLayoutYAxisAnchor
         
-        if #available(iOS 11.0, *) {
-            topAnchor = view.safeAreaLayoutGuide.topAnchor
-        } else {
-            topAnchor = view.topAnchor
-        }
-        
-        primaryStackView.top(to: view, topAnchor, offset: 16)
+        primaryStackView.top(to: layoutGuide(), offset: 16)
```
designatednerd: AH that's how this works. I was fighting with it and thought it only worked in certain circumstances which i was not in. Thank you!

Toshi/Controllers/Browse/DappViewController.swift
```diff
@@ -77,8 +77,8 @@ final class DappViewController: UIViewController {
         stackView.addArrangedSubview(descriptionLabel)
         stackView.addArrangedSubview(urlLabel)
         stackView.addArrangedSubview(enterButton)
-        
-        enterButton.height(44)
+
+        enterButton.heightConstraint.constant = 44
```
designatednerd: more out of curiosity than anything, why do you think this caused a crash?

marijnschilling: Because the ActionButton sets it's own heightConstraint. So then it will have two conflicting height constraints. 

So i only get the crash cause I have `UIViewAlertForUnsatisfiableConstraints` turned on.

designatednerd: ah, ok, i missed that with the action button height constraint. thank you!

PR #665, opened 2017-12-21
Switch to using a .stringsdict for plurals

designatednerd:
This allows us greater flexibility for future localization (and to not have to do all the math about how many items are in something just to get the proper pluralization). 

Also did a little cleanup to make getting a cell out of the request view controller simpler. 

Edit: Also moved the `Localizable.strings` to `Base.lproj` so it will provide the proper base localization when we do another localization.


Toshi/Controllers/Sign In/SignInFooterView.swift
```diff
@@ -36,16 +36,12 @@ final class SignInFooterView: UIView {
...
-        switch remaining {
-        case 0:
+        if remaining == 0 {
             return Localized("passphrase_sign_in_button")
-        case 1:
-            return String(format: Localized("passphrase_sign_in_button_placeholder_singular"), remaining)
-        default:
-            return String(format: Localized("passphrase_sign_in_button_placeholder"), remaining)
+        } else {
+            return LocalizedPlural("passphrase_sign_in_button_placeholder", for: remaining)
```
yuliaveres: hmm do you think it might be easier to use just `localizedStringWithFormat` to not distinguish here. i think localization can handle it 🤔

yuliaveres: oh nvm :)

designatednerd: Yeah - i think having the wrapper around it is useful as a hint to the reader that this is handled under the hood.

PR #668, opened 2017-12-21
Use 32 length group id

yuliaveres:
We need to match group id length to match android, which is 32 length hex string.


yuliaveres:
nope :)

PR #670, opened 2018-01-03
Clean up API tests (except IDAPI), centralize some stuff on Ethereum API

designatednerd:
In this pull I'm adding some parsing tests to make sure we're parsing things correctly as we're getting them. When we have changes to APIs, we should update these tests accordingly. Eventually I'd like to do a parallel set of live-pointed-at-staging tests to make sure we're actually getting the same JSON back from the server we're using locally, but this is at least a start. 

As I was working on the Ethereum tests it jumped out at me that a bunch of JSON parsing was being repeated to access the transaction hash, so I centralized it and the error handling around it in the actual API itself. 

I also disabled code coverage in the Pods so our code coverage better reflects what is covered in *our* code.


Tests/Mocks/AppsAPIClient/getFeaturedApps.json
```diff
@@ -8,17 +8,17 @@
     {
       "about": "It's all about tests",
       "avatar": "https://token-id-service-development.herokuapp.com/avatar/0x8f9bdb7f562ccdedf3c24cf25e9cece9df62138b.png",
-      "average_rating": 1,
+      "average_rating": 4.5,
       "is_app": 1,
-      "location": "",
-      "name": "TestName",
+      "location": "Hamsterdam",
```
roberthein: 💯

Podfile
```diff
@@ -33,4 +33,13 @@ abstract_target 'CocoaPods' do
   target 'Tests' do
 	
   end
+  
+  # Disable Code Coverage for Pods projects
+  post_install do |installer_representation|
+      installer_representation.pods_project.targets.each do |target|
+          target.build_configurations.each do |config|
+              config.build_settings['CLANG_ENABLE_CODE_COVERAGE'] = 'NO'
```
roberthein: 👌

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -310,19 +310,13 @@ extension SOFAWebController: WKScriptMessageHandler {
...
                 completion(callbackId, "{\\\"error\\\": \\\"\(error?.description ?? "Error sending transaction")\\\"}")
 
                 return
             }
 
-            guard let txHash = json["tx_hash"] as? String else {
-                completion(callbackId, "{\\\"error\\\": \\\"Error recovering transaction hash\\\"}")
-
-                return
-            }
```
designatednerd: @yuliaveres @marijnschilling just a heads up that this was changed so it could be centralized and tested

PR #671, opened 2018-01-03
Update ID client tests

designatednerd:
Added a whole mess of parsing tests and mock data, and then added a failure test for the Dapps API where i found it wasn't behaving as I thought. Whoops! Fixed that as well. 😛


Tests/Mocks/IDAPIClient/userAfterDataUpdate.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "Marijntje",
```
roberthein: 😄

designatednerd: I wish i could take credit for that, it was already in other user info :)

PR #674, opened 2018-01-05
iOS-439: Add endpoint to allow fetching details for multiple users

designatednerd:
Added API to allow group chat to easily fetch the details for a bunch of users in one chat + tests.


Toshi/APIClients/idAPIClient.swift
```diff
@@ -430,6 +430,58 @@ typealias DappCompletion = (_ dapps: [Dapp]?, _ error: ToshiError?) -> Void
...
+            // No addresses to actually fetch = no users to return.
+            completion([], nil)
+            
+            return
+        }
+        
+        if addresses.count >= 1000 {
+            // This is beyond the bounds of what this API is supposed to be able to handle:
+            // https://bakkenbaeck.slack.com/archives/C0PHFR45V/p1515150276000331
+            assertionFailure("Talk to the backend team about increasing the limit or paginating results")
```
tristan: The issue with the limit is a physical limit of processing the query string up to a certain length (2^16 characters). It's not physically possible to increase it server side.

Pagination server side also doesn't make sense as the result is exactly the number of values the client asks for, thus pagination would just be breaking up the `addresses` list into batches (and due to the above limitation, pagination server side wouldn't help).

I would also say it's probably better to make the batches even smaller at ~100 per batch, for both server and client performance reasons, but we'd have to do some testing around this.

Not sure where this batching should be done, but this assertion definitely isn't correct

designatednerd: fair enough, I'll update it.

designatednerd:
@yuliaveres @tristan ready for another look

PR #676, opened 2018-01-08
[IOS-440]: Status messages

yuliaveres:
We have already have SOFAStatus message type for handling status messages in a chat.
In this PR I processed Signal `TSInfoMessage` (which corresponds for status message in Signal) in `handleSignalMessage` inside `ChatInteractor` where we do process other signal messages and create SOFA message for each thus we can render a message in a chat.
Also I included SOFAStatus type into displayable types list as before no status could ever be shown in a chat.


Toshi/Avatars/AvatarManager.swift
```diff
@@ -128,6 +152,13 @@ final class AvatarManager: NSObject {
                 return
             }
 
+            guard !self.isDownloadInProgress(for: path) else {
+                completion?(nil, path)
```
designatednerd: So if the download is in progress, we return nil? How does the caller know when the other download finishes?

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -207,7 +207,8 @@ final class ChatInteractor: NSObject {
     /// - Parameters:
     ///   - interaction: the interaction to handle. Incoming/outgoing messages, wrapping SOFA structures.
     ///   - shouldProcessCommands: If true, will process a sofa wrapper. This means replying to requests, displaying payment UI etc.
-    func handleSignalMessage(_ signalMessage: TSMessage, shouldProcessCommands: Bool = false) -> Message {
+    ///   - shouldUpdateGroupMembers: If true will go over gorup members IDs and update with calling id service and download avatar
```
designatednerd: `gorup` -> `group` (I make this typo ALL THE TIME)

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -216,6 +217,69 @@ final class ChatInteractor: NSObject {
...
+
+            if let thread = infoMessage.thread as? TSGroupThread, let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+
+                if let author = appDelegate.contactsManager.tokenContact(forAddress: authorId) {
+
+                    var actionString = ""
+                    var object = "Marek"
+                    let subject = author.nameOrDisplayName
+
+                    if customMessage == GroupTitleChangedMessage {
```
designatednerd: you should be able to `switch` on the string here, though you'll have to add a `default` clause since it's not an enum. IMO it's a bit easier to read than a giant if-else statement

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -216,6 +217,69 @@ final class ChatInteractor: NSObject {
...
+            let customMessage = infoMessage.customMessage
+            let updateInfoString = infoMessage.additionalInfoString
+            let authorId = infoMessage.authorId
+
+            if let thread = infoMessage.thread as? TSGroupThread, let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+
+                if let author = appDelegate.contactsManager.tokenContact(forAddress: authorId) {
+
+                    var actionString = ""
+                    var object = "Marek"
```
designatednerd: should this still be here?

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -216,6 +217,69 @@ final class ChatInteractor: NSObject {
...
+
+                    if customMessage == GroupTitleChangedMessage {
+                        actionString = "rename"
+                        object = updateInfoString
+                    } else if customMessage == GroupAvatarChangedMessage {
+                        actionString = "changed avatar"
+                    } else if customMessage == GroupMemberLeftMessage {
+                        actionString = "leave"
+                    } else if customMessage == GroupMemberJoinedMessage {
+                        actionString = "added"
```
designatednerd: where are these action strings being used? Could it be worth making them a string enum?

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -216,6 +217,69 @@ final class ChatInteractor: NSObject {
...
+                                    } else {
+                                        resultNames.append(name.truncate(length: 10))
+                                    }
+                                } else {
+                                    resultNames.append(name)
+                                }
+                            }
+
+                            object = resultNames.joined(separator: ", ")
+                        }
```
designatednerd: Simplification alert: Everything in this `if` statement could probably be pulled out into its own method, which would make it easier to test.

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -79,30 +99,26 @@ - (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)newModel contactsManage
...
-        updatedGroupInfoString = [updatedGroupInfoString
-                                  stringByAppendingString:[NSString stringWithFormat:NSLocalizedString(@"GROUP_MEMBER_JOINED", @""),
-                                                           [newMembersNames componentsJoinedByString:@", "]]];
+        updateTypeString = [NSString stringWithFormat:NSLocalizedString(GroupMemberJoinedMessage, @""),
+                            [membersWhoJoined.allObjects componentsJoinedByString:@", "]];
+        updatedGroupInfoString = [membersWhoJoined.allObjects componentsJoinedByString:@", "];
     }
 
-    return updatedGroupInfoString;
+    return @{GroupUpdateTypeSting: updateTypeString,
```
designatednerd: I don't know if we have official style on this but when I do objc dictionary literals I usually do: 

```objc
@{
    "foo": true,
   "bar": "Something!",
};
```

Main reason is so that it's a little easier to look at diffs since only the actual content changes. 

Definitely not required but worth considering for your literals here.

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -52,24 +63,33 @@ - (BOOL)isEqualToGroupModel:(TSGroupModel *)other {
     return YES;
 }
 
-- (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)newModel contactsManager:(id<ContactsManagerProtocol>)contactsManager {
+- (NSDictionary *)getInfoAboutUpdateTo:(TSGroupModel *)newModel contactsManager:(id<ContactsManagerProtocol>)contactsManager {
+    NSString *updateTypeString = @"";
     NSString *updatedGroupInfoString = @"";
+
     if (self == newModel) {
```
marijnschilling: So if there are no changes we return an empty string for `GroupUpdateTypeSting` and `GroupInfoString`? 

We could put this in a guard maybe? 

```
let modelChanged = self != newModel
guard modelChanged else { //return the empty string dict thing } 
```

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -52,24 +63,33 @@ - (BOOL)isEqualToGroupModel:(TSGroupModel *)other {
...
-        return NSLocalizedString(@"GROUP_UPDATED", @"");
+        return @{GroupUpdateTypeSting: NSLocalizedString(GroupUpdatedMessage, @""),
```
marijnschilling: we can use the var `updateTypeString` here instead of creating a new empty string since we also do that for the `GroupInfoString` with `updatedGroupInfoString`

Toshi/Avatars/AvatarManager.swift
yuliaveres: I will revert changes related to this in the branch as it requires more work as I see now and i will do it in separate branch (initially that was just an improvement idea)

yuliaveres: that is a good point

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -52,24 +63,33 @@ - (BOOL)isEqualToGroupModel:(TSGroupModel *)other {
...
+                 GroupInfoString: updatedGroupInfoString
+                 };
     }
+
     if (![_groupName isEqual:newModel.groupName]) {
```
marijnschilling: Would like it if this was cut up in separate methods. And might be more clearer with a bool test with a clear name. 
```
let groupNameDidChange = ![_groupName isEqual:newModel.groupName]
if groupNameDidChange {
   
}

File: Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m [hunk 11]
```diff
@@ -52,24 +63,33 @@ - (BOOL)isEqualToGroupModel:(TSGroupModel *)other {
...
         !([UIImagePNGRepresentation(_groupImage) isEqualToData:UIImagePNGRepresentation(newModel.groupImage)])) {
-        updatedGroupInfoString =
-            [updatedGroupInfoString stringByAppendingString:NSLocalizedString(@"GROUP_AVATAR_CHANGED", @"")];
+        updateTypeString =
+        [updateTypeString stringByAppendingString:NSLocalizedString(GroupAvatarChangedMessage, @"")];
     }
-    if ([updatedGroupInfoString length] == 0) {
-        updatedGroupInfoString = NSLocalizedString(@"GROUP_UPDATED", @"");
+
+    if ([updateTypeString length] == 0) {
```
marijnschilling: I don't understand why this is happening.

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -52,24 +63,33 @@ - (BOOL)isEqualToGroupModel:(TSGroupModel *)other {
     return YES;
 }
 
-- (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)newModel contactsManager:(id<ContactsManagerProtocol>)contactsManager {
+- (NSDictionary *)getInfoAboutUpdateTo:(TSGroupModel *)newModel contactsManager:(id<ContactsManagerProtocol>)contactsManager {
```
marijnschilling: we don't use contactsmanager anymore, we can remove the parameter.

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
```diff
@@ -79,30 +99,26 @@ - (NSString *)getInfoStringAboutUpdateTo:(TSGroupModel *)newModel contactsManage
...
+        NSString *oldMembersString = [[membersWhoLeft allObjects] componentsJoinedByString:@", "];
+        updateTypeString = [updateTypeString
+                            stringByAppendingString:[NSString
+                                                     stringWithFormat:NSLocalizedString(GroupMemberLeftMessage, @""),
+                                                     oldMembersString]];
+        updatedGroupInfoString = oldMembersString;
     }
-    
+
     if ([membersWhoJoined count] > 0) {
```
marijnschilling: Would be great to have this whole members who left and joined code in a separate method.

Toshi/APIClients/idAPIClient.swift
```diff
@@ -116,6 +115,19 @@ typealias DappCompletion = (_ dapps: [Dapp]?, _ error: ToshiError?) -> Void
...
     }
 
+    func updateContacts(with identifiers: [String]) {
+        fetchUsers(with: identifiers) { users, _ in
+
+            guard let fetchedUsers = users else { return }
+            guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
+
+            for user in fetchedUsers {
+                Yap.sharedInstance.insert(object: user.json, for: user.address, in: TokenUser.storedContactKey)
```
marijnschilling: Shouldn't this only happen for the newly added users? Now we insert all the users from the group in the Yap database when one new user is added right?

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -216,6 +217,69 @@ final class ChatInteractor: NSObject {
             return Message(sofaWrapper: nil, signalMessage: invalidKeyErrorMessage, date: invalidKeyErrorMessage.dateForSorting(), isOutgoing: false)
         }
 
+        if let infoMessage = signalMessage as? TSInfoMessage {
+            let customMessage = infoMessage.customMessage
```
marijnschilling: Could we put all this InfoMessage related code in a separate method?

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
```diff
@@ -39,7 +39,16 @@ final class GroupInfoViewModel {
...
         groupInfo = GroupInfo()
         groupInfo.title = groupModel.groupName
-        groupInfo.participantsIDs = groupModel.groupMemberIds
+
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else {
+            CrashlyticsLogger.log("Failed to access app delegate")
+            fatalError("Can't access app delegate")
+        }
+        
+        let members = appDelegate.contactsManager.tokenContacts.filter { groupModel.groupMemberIds.contains($0.address) }
```
marijnschilling: Hmmm what do you exactly filter out here?

Toshi/Controllers/Recent/RecentViewController.swift
```diff
@@ -171,7 +171,9 @@ extension RecentViewController: UITableViewDataSource {
     }
 
     func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
-        if section == 0 && dataSource.hasUnacceptedThreads {
+        let isUnacceptedThredsSection = (section == 0 && dataSource.hasUnacceptedThreads)
```
marijnschilling: should be `isUnacceptedThreadsSection` right?

Toshi/Models/Message.swift
```diff
@@ -121,10 +121,12 @@ class Message: NSObject {
             guard !message.body.isEmpty else { return false }
         } else if let paymentRequest = sofaWrapper as? SofaPaymentRequest {
             guard EthereumAddress.validate(paymentRequest.destinationAddress) else { return false }
+        } else if let status = sofaWrapper as? SofaStatus {
```
marijnschilling: Should we add this case to the comments above here?

Toshi/Models/Message.swift
```diff
@@ -121,10 +121,12 @@ class Message: NSObject {
...
+            guard status.attributedText != nil else { return false }
         }
 
         // or not one of the types below
```
marijnschilling: I do not undesttand this comment. Its not displayable ifff it's these types below? ... I don't get htat.

Toshi/Avatars/AvatarManager.swift
```diff
@@ -174,6 +174,8 @@ final class AvatarManager: NSObject {
         completion(avatar, path)
     }
 
+    // Finds avatar for the KEY which may be token_id/address or resource url path in cache
+    // triggers download if not found
```
designatednerd: can you use swiftdoc style comments for this please?

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -356,6 +361,74 @@ final class ChatInteractor: NSObject {
...
+                                } else {
+                                    resultNames.append(name.truncate(length: 10))
+                                }
+                            } else {
+                                resultNames.append(name)
+                            }
+                        }
+
+                        object = resultNames.joined(separator: ", ")
+                    }
```
designatednerd: I still think this bit should probably be broken out into its own method so it can more easily tested separately from message parsing.

Pods/SignalServiceKit/SignalServiceKit/src/Messages/TSGroupModel.m
yuliaveres: we have OBJC here

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
yuliaveres: contacts for this group, should I use different syntax? `{ object in ... }` ?

Toshi/Models/Message.swift
yuliaveres: if the message type does not match any of those types, i will change this

Toshi/APIClients/idAPIClient.swift
```diff
@@ -122,7 +122,10 @@ typealias DappCompletion = (_ dapps: [Dapp]?, _ error: ToshiError?) -> Void
             guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
 
             for user in fetchedUsers {
-                Yap.sharedInstance.insert(object: user.json, for: user.address, in: TokenUser.storedContactKey)
+                if !Yap.sharedInstance.containsObject(for: user.address, in: TokenUser.storedContactKey) {
+                    Yap.sharedInstance.insert(object: user.json, for: user.address, in: TokenUser.storedContactKey)
+                }
```
designatednerd: would it make more sense to filter out the users that are already in the database before making the call to fetch the users? might save the server some time fetching data we already have.

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -429,6 +409,32 @@ final class ChatInteractor: NSObject {
...
+            let validDisplayName = appDelegate.contactsManager.displayName(forPhoneIdentifier: nameOrAddress)
+            if !validDisplayName.isEmpty {
+                return validDisplayName
+            } else {
+                return nameOrAddress.truncate(length: 10)
+            }
+        }
+
+        return nameOrAddress
+    }
```
designatednerd: 👏 👏 👏 🗜 .... you know what i'm just gonna leave that last one

Toshi/Avatars/AvatarManager.swift
```diff
@@ -151,7 +151,13 @@ final class AvatarManager: NSObject {
...
     }
 
-    // Downloads or finds avatar for the KEY which may be token_id/address or resource url path
+    /// Downloads or finds avatar for given key.
+    ///
+    /// - Parameters:
+    ///   - key: token_id/address or resource url path.
+    ///   - completion: The completion closure to fire when the request completes or image is found in cache.
+    ///                 - image: Found in cache or fetched image.
+    ///                 - path: Path for found in cache or fetched image.
```
marijnschilling: Love this! 👏  ❤️

Toshi/Controllers/GroupChat/GroupInfo/GroupInfoViewModel.swift
designatednerd: Is there any other way to query this? `tokenContacts` could be an enormous array.

yuliaveres: i do not think so. this should not cause any issue. what other way would u suggest ?

designatednerd: Thinking about this overnight, let's deal with this if it becomes an actual problem. Leave it for now

Toshi/Controllers/GroupChat/NewGroup/NewGroupViewModel.swift
```diff
@@ -64,6 +63,25 @@ final class NewGroupViewModel {
...
+        guard let currentUser = TokenUser.current else {
+            CrashlyticsLogger.log("Failed to access current user")
+            fatalError("Can't access current user")
+        }
+
+        var members = appDelegate.contactsManager.tokenContacts
+        members.append(currentUser)
+        members = members.filter { groupInfo.participantsIDs.contains($0.address) }
+
+        sortedMembers = members.sorted { $0.username < $1.username }
```
designatednerd: is there any way to centralize this logic between the group view model and the new group view model?

yuliaveres: my initial approach was to keep sorted members as private, thus it required to have them separately in two models. otherwise i might need to open `sortedMembers` and include it in the protocol. it should not be needed there though for public access in my opinion 🤔 wdyt ?

designatednerd: maybe make it a method with a default implementation on the protocol? I see what you mean about wanting to keep it private, but i think the benefits of centralizing the code (esp in terms of bug fixing/prevention) outweigh the risks.

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -81,15 +81,14 @@ extension MessagesRequestsViewController: UITableViewDataSource {
 
         guard let thread = dataSource.unacceptedThread(at: indexPath) else { return UITableViewCell(frame: .zero) }
 
-        var avatar: UIImage?
+        var avatar = thread.avatar() ?? UIImage(named: "avatar-placeholder")
         var subtitle = "..."
         var title = ""
 
         if thread.isGroupThread() {
             avatar = (thread as? TSGroupThread)?.groupModel.groupImage ?? UIImage(named: "avatar-placeholder")
```
designatednerd: can this line also be deleted with the change to using `thread.avatar` above?

PR #677, opened 2018-01-08
UI Tests + Robot Pattern Starter

designatednerd:
The good news is this PR is not **as** bad as it looks: Only 1415/~7400 additions and 51/~230 additions are actually adding tests. 

The bad news is...this still kind of got away from me a little, since in order to get something complete I needed to write a bunch of boilerplate. 

The biggest thing in this PR is the addition of UI Tests following the Robot pattern. This is a pattern I've already discussed at length with @roberthein and @marijnschilling and have used extensively in other projects. 

I really need to write a blog post on it so I don't have to keep making people look at slides or videos, but [here's a video of the short version](https://www.youtube.com/watch?v=flZDWc25paw) and [here are some slides from the long version](https://speakerdeck.com/designatednerd/better-ui-testing-with-swift-robots-cocoaconf-next-door-san-jose-june-2017) from my talk on this. 

The TL;DR is that it's a pattern using protocols to make it easier to read UI tests by abstracting away implementation details for what each action taken by a user is doing. It also allows easy swap-out of the underlying framework, which I talk about at length in the short video but you can see for yourself in the [test project](https://github.com/designatednerd/GoCubs) that I've been using along with that talk. 

Phew. OK.

In this PR: 
- Add a "Tests_UI" non-UI test bundle. I know that sounds weird, but what Apple considers as a non-UI tests bundle is anything that runs in the same process as the main app, which Earl Grey (and KIF) do. 
- Add that bundle to the Podfile and then `pod install`, which is where the vast majority of the changed lines come from. 
- Add `BasicRobot`, `EarlGreyRobot`, and a few other instances of the Robot Pattern to test stuff on the splash screen
- Add `SplashScreenUITests` to actually test those items. 
- Fix a bug with the notification stuff not being fired on the main thread, since that was causing weirdness with the tests. 
- Add utilities for determining whether tests are running and showing test-only alerts for things which Earl Grey isn't capable of testing (ie, system alerts and SFSafariViewController). 
- Swap out `main.m` for `main.swift` to take advantage of centralized utility. 
- Disable the test for creating a new account until issues with sign-out are addressed.


marijnschilling:
I'm super enthousiastic about this robot pattern! Would like to get into it. In this PR can I already try to see if i can apply it to centain ViewController I made? that might be a nice way to review this PR, since there's so much code to check! That might be a nice way to get to know it ✌️

designatednerd:
@marijnschilling yes, although there is some weirdness around logout that means you already have to have a logged in user to test it on anything beyond the login/create account barrier.

marijnschilling:
I will start a new feature and maybe we can discuss how i could use these robots 🤖  to test the feature.

Tests_UI/SplashScreenUITests.swift
```diff
@@ -0,0 +1,95 @@
...
+        self.splashRobot
+            .validateOnSplashScreen()
+            .select(button: .createNewAccount)
+            .validateTermsDialogShowing()
+            .select(termsOption: .read)
+            .validateTermsDialogGone()
+            .validateTestAlertShowing(withMessage: TestOnlyString.readTermsAlertMessage(termsURL: expectedTermsURL))
+            .dismissTestAlert()
+            .validateTestAlertGone()
+
```
yuliaveres: do we need to `validateOnSplashScreen` again ?

Toshi/Protocols/PaymentPresentable.swift
```diff
@@ -0,0 +1,93 @@
...
+
+extension PaymentPresentable where Self: UIViewController {
+
+    /// Displays payment confirmation for given user info and parameters
+    ///
+    /// - Parameters:
+    ///   - userInfo:
+    ///   - parameters: include from, to, and value in wei fields
+
+    func displayPaymentConfirmation(userInfo: UserInfo, parameters: [String: Any]) {
```
yuliaveres: this is probably post merge leftover.

Tests_UI/Robots/SignInRobot.swift
```diff
@@ -0,0 +1,85 @@
...
+        tapButtonWith(accessibilityLabel: button.accessibilityLabel,
+                      file: file,
+                      line: line)
+        
+        return self
+    }
+    
+    // MARK: - Validators
+    
+    @discardableResult
```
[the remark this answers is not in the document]
designatednerd: Yeah we can definitely take a look at that long term - this stuff isn't getting called from any tests quite yet.

Tests_UI/SplashScreenUITests.swift
designatednerd: Personally I don't think so since the test alerts never actually do anything

Tests_UI/Robots/MyProfileRobot.swift
```diff
@@ -0,0 +1,146 @@
...
+    qrCode,
+    signOut,
+    storePassphrase
+    
+    var accessibilityLabel: String {
+        switch self {
+        case .advanced:
+            return Localized("settings_cell_advanced")
+        case .balance:
+            assertionFailure("You need to set this up on the balance cell before you use it")
```
marijnschilling: Aha so because by default the accesibility label is the title in the view, we don't need to set in on the `advanced` cell because we can just point to the localized string used on that cell. 

But for the others we should set it manually in the `ProfileTableViewController`?

Would we use LocalizedStrings for that as well?

Tests_UI/Robots/SignInRobot.swift
```diff
@@ -0,0 +1,83 @@
...
+// MARK: - Robot to deal with the Sign In With Passphrase page of the app
+
+protocol SignInRobot: BasicRobot { }
+
+// MARK: The various buttons which are on the passphrase page
+
+enum SignInScreenButton {
+    case
+    back,
+    howWork,
```
marijnschilling: `howWork` 😂 I'm sorry but that sounds like an Amsterdam tourist is asking help for the trainticket machine.

but I guess it kinda makes sense to call `How does it work` that... But i wouldnt be opposed to just writing out the entire title. Thats why they invented autocomplete right? Not just because i think it sounds weird but also to have 0 chance for confusion (not in this case but maybe it could cause confusion in other cases)

It could become quite long, but i dont really see a problem with that myself.

Tests_UI/Robots/SignInRobot.swift
```diff
@@ -0,0 +1,83 @@
...
+    signIn,
+    wordsLeft(words: Int)
```
marijnschilling: I would be interested to see how a check for this would work, is there any reason you haven't wrote a test for this yet?

Tests_UI/SplashScreenUITests.swift
```diff
@@ -0,0 +1,95 @@
...
+            // TODO: Re-enable using full terms check when Earl Grey figures their shit out with SFSafariViewController.
```
marijnschilling: Re-enable what? 😬

Toshi/Controllers/Sign In/SplashViewController.swift
```diff
@@ -155,6 +156,11 @@ final class SplashViewController: UIViewController {
         
         let read = UIAlertAction(title: Localized("accept_terms_action_read"), style: .default) { [weak self] _ in
             guard let url = URL(string: "http://www.toshi.org/terms-of-service/") else { return }
+            guard !UIApplication.isUITesting else {
+                self?.showTestAlert(message: TestOnlyString.readTermsAlertMessage(termsURL: url))
```
marijnschilling: Ah is this showing an alert with a predefined string instead of getting the terms form the interwebs? Is it not working to test this with data from interwebs?

Toshi/Extensions/UIViewController+TestHelpers.swift
```diff
@@ -0,0 +1,40 @@
...
+
+// Methods which help out with testing.
+extension UIViewController {
+    
+    /// Shows a test alert with the given message, a defined title, and an OK button to dismiss it.
+    ///
+    /// - Parameter message: The message to display on the alert.
+    func showTestAlert(message: String) {
+        guard UIApplication.isUITesting else {
+            assertionFailure("DON'T CALL THIS IN PROD!")
```
marijnschilling: Do we need capitalisation? 😬 I'm a bit allergic to it, and think just an assertionfailure wiht an exclimation mark will do the job just fine.

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -199,6 +199,8 @@
 "sign-out-failure-title" = "Could not sign out";
 "sign-out-failure-message" = "Error attempting to unregister from chat service. Our engineers are looking into it.";
 "cancel_action_title" = "Cancel";
+"back_action_title" = "Back";
```
marijnschilling: maybe we can make a file for ui testing specific localized strings? Or maybe a separate section in this file? Easier to not get them mixed up and wornlgy used that way

Toshi/Utilities/AccessibilityIdentifier.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+// Accessibility identifiers which will never be visible to users or read to VoiceOver users - only used for testing purposes.
+enum AccessibilityIdentifier: String {
+    case
+    mainTabBar = "main_tab_bar"
```
marijnschilling: I don't get this 😬 In what case do we choose to use the label and in which the identifier?

Toshi/Utilities/TestOnlyString.swift
```diff
@@ -0,0 +1,31 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+// Strings which will only be used during UI testing and thus do not need to be localized or instantiated.
```
marijnschilling: hmmm maybe just add the `Back` and `Done` button strings here?

Tests_UI/Robots/BasicRobot.swift
```diff
@@ -0,0 +1,140 @@
...
+    // MARK: - Is it there?
+    
+    /// Runs an action where the implementation's test framework validates that a view is presently visible based on its accessibility label.
+    ///
+    /// - Parameters:
+    ///   - accessibilityLabel: The Accessibility Label of the view to look for. This would be read out loud to VoiceOver users.
+    ///   - file: The file from which this method is being called.
+    ///   - line: The line from which this method is being called.
+    func confirmViewVisibleWith(accessibilityLabel: String,
+                                file: StaticString,
```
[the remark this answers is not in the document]
designatednerd: So it's more for when you run them locally - this causes the failure error to show up at the file and line of the caller, so you can see where the failure occurred.

Tests_UI/Robots/BasicRobot.swift
```diff
@@ -0,0 +1,140 @@
...
+    // MARK: - Universal actions
+    
+    @discardableResult
+    func dismissTestAlert(file: StaticString = #file,
+                          line: UInt = #line) -> Self {
+        tapViewWith(accessibilityLabel: TestOnlyString.okButtonTitle,
+                    file: file,
+                    line: line)
+        
+        return self
```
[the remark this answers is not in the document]
designatednerd: to allow the DSL-like chaining that's used in the tests.

Tests_UI/Robots/BasicRobot.swift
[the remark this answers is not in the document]
marijnschilling: Oh really, i didnt see this happening when i was running the test when i was logged in, so when they failed then.

Tests_UI/Robots/EarlGreyRobot.swift
```diff
@@ -0,0 +1,133 @@
...
+import EarlGrey
+import UIKit
+@testable import Toshi
+import XCTest
+
+// MARK: - Concrete implementation of Robot pattern for Earl Grey
+
+class EarlGreyRobot {
+    
+    private func retriveElementOfClass<T>(_ clazz: T.Type, _ completion: @escaping (T?) -> Void) -> GREYActionBlock where T: AnyObject {
```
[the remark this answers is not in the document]
designatednerd: So `BasicRobot` is a protocol. `EarlGreyRobot` is a concrete implementation of that protocol. 

The methods lower in this file in the `BasicRobot` extension are implementations of the required methods in the `BasicRobot` protocol. These methods are exclusive to Earl Grey, which is why they're only in the `EarlGreyRobot`. 

The methods in the `EarlGreyRobot: BasicRobot` extension can call into these methods because they are within the same class.

Does that make more sense?

Tests_UI/Robots/MyProfileRobot.swift
designatednerd: Correct, for most things you can use the `accessibilityLabel` with the text on the view.

For these two, there's two labels that are of variable content, so we need to decide if we want to make a single combined accessibility label for the cell or if we want to just use an identifier. 

Since I'm not actually doing tests with this yet, I left this with TODOs

Tests_UI/Robots/SignInRobot.swift
designatednerd: I wanted to keep the initial PR to a single set of tests. I will definitely be adding tests for this once this PR is merged.

Tests_UI/Robots/SignInRobot.swift
designatednerd: Nah, that's fair, I'll update the name

Toshi/Controllers/Sign In/SplashViewController.swift
designatednerd: It is not, that's what the above TODO is around - the problem is that EarlGrey doesn't play super nice with SafariViewControlller at the moment. (Neither does KIF, for what it's worth). 

At this point, we can at least still check that the correct URL will be navigated to by showing an alert and checking its contents.

Toshi/Extensions/UIViewController+TestHelpers.swift
designatednerd: I wanted this to be an extra-obnoxious failure. :trollface:

Toshi/Resources/Base.lproj/Localizable.strings
designatednerd: There's another place where we actually use this within the application, IIRC. 

I'm working on something to try to generate code based on our localizable.strings file though, so hopefully that will help

Toshi/Utilities/AccessibilityIdentifier.swift
designatednerd: `accessibilityLabel`s are read out loud to voice-over users. They should be localized so, for example, a Dutch-speaking VoiceOver user doesn't get a bunch of random things read to them in English. 

`acccessibilityIdentifier`s are only available to tests, and will never be seen or read by voiceover users, so they don't need to be translated or have semantic meaning, they just need to be consistent. 

When I write tests, I mostly use `Labels` for anything that actually has to be tapped. I use `identifiers` as helpers to say "I need to look for the existence of this view" or "i need to look for a subview of this view"

Toshi/Utilities/TestOnlyString.swift
designatednerd: These are things which will be read out loud to voice over users, which is why I want to keep them localized.

Tests_UI/Robots/BasicRobot.swift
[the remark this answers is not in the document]
designatednerd: Yeah if you're logged in, the tests are going to fail since the only actual *tests* I wrote are against things that happen if you're not logged in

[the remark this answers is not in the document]
marijnschilling: oke, but then it doesnt print anything to the console?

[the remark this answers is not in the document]
designatednerd: It does whatever XCTest does by default - by passing all these things through to `XCTAssert` it allows it to fire as if it was actually being called directly in the test method. 

I honestly can't remember if XCTest does it by default or not, but whatever stock XCT would do, this will do.

designatednerd:
🤖 🎉 !!!

PR #678, opened 2018-01-09
Fix Pod Structure; Beginnings of Message Parsing tests

designatednerd:
In this PR: 
- Fix pod structure so we don't have duplicated copies of libraries in the test bundle (which led to [this fun adventure](https://twitter.com/designatednerd/status/950414376946798592) last night)
- Begin adding tests around parsing things into the `Message` object from all the various signal messages.


Tests/MessageParsingTests.swift
```diff
@@ -0,0 +1,122 @@
...
+    }()
+    
+    private lazy var textMessageBody: String = {
+        return "\(SofaType.message.rawValue){\"body\":\"o hai\"}"
+    }()
+    
+    // MARK: - Single User Threads
+    
+    func testHandlingInvalidKeyMessage() {
+        let interactor = ChatInteractor(output: nil, thread: self.normalThread)
```
marijnschilling: self? 😃  Is that necessary here or is it just habit?

Tests/MessageParsingTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        
+        let message = TSInvalidIdentityKeySendingErrorMessage(timestamp: 5, // Must be greater than 0
```
marijnschilling: maybe you can make the timestamp value a property of this test class, than you can put the comment there and it will probably prevent miss-use . Like if people would add a test they just use the property, and won't make the mistake to put in 0.

Tests/MessageParsingTests.swift
```diff
@@ -0,0 +1,122 @@
...
+                                                              in: self.normalThread,
```
marijnschilling: some more self

Tests/MessageParsingTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        let message = TSOutgoingMessage(timestamp: 5,
+                                        in: normalThread,
+                                        messageBody: textMessageBody,
+                                        attachmentIds: NSMutableArray(array: ["One", "Two"]))
+        
+        let parsed = interactor.handleSignalMessage(message)
+
+        XCTAssertNil(parsed.fiatValueString)
+        XCTAssertNil(parsed.ethereumValueString)
+        XCTAssertNil(parsed.attachment) // This should be nil since there arent' actually attachments saved
```
marijnschilling: I wonder, is there a way in XCTests to have a message printed to the console when the test fails? Would be cool if we could use or add that. So we would just see int he logs what it means when this test fails.

Tests/MessageParsingTests.swift
designatednerd: ah yeah good point

Tests/MessageParsingTests.swift
designatednerd: Old habits die hard 😝

Tests/MessageParsingTests.swift
designatednerd: Yes, a message prints to the console in addition to the failure that pops up in the tests. You can see this if you check out this branch and flip, for example, this line, to `XCTAssertNotNil`.

designatednerd:
@marijnschilling still good to go?

marijnschilling:
Even better ✌️

PR #681, opened 2018-01-10
Point to SSK commit making valid check for new group

yuliaveres:
For some weird reason related to YAP, even already existing object might have `uniqueId` as `nil`, which under normal conditions is valid only for fresh local variable object, not saved yet. The fix for this is checking for group name as well. Because actually, new group never has image or name. Thus checking for name being `nil` in addition to checking for `uniqueId` guarantees that group object is actually new and user is added to it, if s/he receives info message in the relevant thread. (As additional info: we need this when obtaining info about group model change, the fresh local variable with message info of group object is created first, that is why it never has `uniqueId` and `name` and already existing object is retrieved if present; both objects are compared and info is composed)


designatednerd:
@yuliaveres is this happening when the group chat is started by another user on another platform? Or just another user?

yuliaveres:
@designatednerd it does not matter, just another user

yuliaveres:
the issue was caught in the group created by current user who was affected (on iOS)

marijnschilling:
Ah so they got a message 'Group created' when they created the group themselves while it was not supposed to? Or what was the effect on the user? (if there was any)

yuliaveres:
@marijnschilling  yes the affected user seen the "U were added to the group" (like the group is new for him/her)

designatednerd:
So after further discussion it seems like this is a workaround for a race condition when you get a bunch of group-related messages coming in at once, and the one for the creation of the group has not resulted in an identifier being created yet because of a race condition, but other messages related to that group are still being processed. 

Whee.

PR #682, opened 2018-01-10
Actually show preview strings of messages on recent page

yuliaveres:
Currently on Recent view in thread cell there is not subtitle displaying message overview. Currently it works only for plain messages - then it displays part of a message. But when there is image or payment/payment request there is nothing (and used to be). With new generic cells which resize we have not very nice looking list of threads in this case. Moreover there should be description of message. Here I added strings for corresponding messages type, as it is done on Android.

Before:
screen shot 2018-01-10 at 17 43 37

After:
screen shot 2018-01-10 at 17 45 30


designatednerd:
@yuliaveres can you add before/after screenshots of this plz?

yuliaveres:
@marijnschilling  screenshots are here :)

PR #683, opened 2018-01-10
More message parsing tests

designatednerd:
Added a few more parsing tests around parsing and messages. 

Also removed the `senderId` property from `Message` since we weren't actually using it.


Tests/MessageParsingTests.swift
```diff
@@ -117,7 +126,219 @@ class MessageParsingTests: XCTestCase {
...
+        XCTAssertNil(parsed.text)
+        XCTAssertNil(parsed.attributedText)
+        
+        XCTAssertTrue(parsed.isOutgoing)
+        XCTAssertTrue(parsed.isDisplayable)
+        
+        // Not actionable despite message type since outgoing.
+        XCTAssertFalse(parsed.isActionable)
+        
+        guard let payment = parsed.sofaWrapper as? SofaPayment else {
```
yuliaveres: not sure if this is needed as soon as we have `XCTAssertEqual(parsed.sofaWrapper?.type, .payment)` ?

Tests/MessageParsingTests.swift
```diff
@@ -117,7 +126,219 @@ class MessageParsingTests: XCTestCase {
...
+        
+        XCTAssertNil(parsed.attachment)
+        XCTAssertNil(parsed.image)
+        XCTAssertNil(parsed.text)
+        XCTAssertNil(parsed.attributedText)
+        
+        XCTAssertTrue(parsed.isOutgoing)
+        XCTAssertTrue(parsed.isDisplayable)
+        
+        // Not actionable despite message type since outgoing.
```
yuliaveres: i think that actionable is only `PaymentRequest` 🤔

Tests/MessageParsingTests.swift
```diff
@@ -117,7 +126,219 @@ class MessageParsingTests: XCTestCase {
...
+        XCTAssertNil(parsed.ethereumValueString)
+        
+        XCTAssertNil(parsed.attachment)
+        XCTAssertNil(parsed.image)
+        XCTAssertNil(parsed.text)
+        XCTAssertNil(parsed.attributedText)
+        
+        XCTAssertFalse(parsed.isOutgoing)
+        
+        // Not actionable despite message type since payment is complete.
```
yuliaveres: same here, reason for it being not actionable - not a `PaymentRequest`

Tests/MessageParsingTests.swift
```diff
@@ -117,7 +126,219 @@ class MessageParsingTests: XCTestCase {
...
+        
+        XCTAssertEqual(parsed.messageType, "Actionable")
+        XCTAssertEqual(parsed.text, "")
+        XCTAssertEqual(parsed.signalMessage, message)
+        XCTAssertEqual(parsed.title, "Payment request")
+        XCTAssertEqual(parsed.attributedTitle?.string, "Payment request")
+        XCTAssertEqual(parsed.sofaWrapper?.type, .paymentRequest)
+        XCTAssertEqual(parsed.deliveryStatus, .attemptingOut)
+        
+        guard let paymentRequest = parsed.sofaWrapper as? SofaPaymentRequest else {
```
yuliaveres: and here, do we need it as we check for sofa wrapper type being `.paymentRequest` ? 
(`XCTAssertEqual(parsed.sofaWrapper?.type, .paymentRequest)`)

Tests/MessageParsingTests.swift
```diff
@@ -117,7 +126,219 @@ class MessageParsingTests: XCTestCase {
...
+                       "Ethereum value string for static hex code has changed")
+        XCTAssertTrue(parsed.subtitle?.contains(ethereumValueString) ?? false,
+                      "Subtitle \"\(parsed.subtitle ?? "(null)")\" does not contain \"\(ethereumValueString)\"")
+        XCTAssertTrue(parsed.subtitle?.contains(paymentRequest.fiatValueString) ?? false,
+                      "Subtitle \"\(parsed.subtitle ?? "(null)")\" does not contain \"\(paymentRequest.fiatValueString)\"")
+        XCTAssertTrue(parsed.attributedSubtitle?.string.contains(ethereumValueString) ?? false,
+                      "Attributed subtitle's underlying string does not contain \"\(ethereumValueString)\"")
+        XCTAssertTrue(parsed.attributedSubtitle?.string.contains(paymentRequest.fiatValueString) ?? false,
+                      "Attributed subtitle's underlying string does not contain \"\(paymentRequest.fiatValueString)\"")
+
```
yuliaveres: 🙈

Tests/MessageParsingTests.swift
designatednerd: It's mostly to pull it out so I can test that the ethereum value is correct and part of the message, and to access the fiat value that would change based on the exchange rate.

Tests/MessageParsingTests.swift
designatednerd: Any idea why the `messageType` string on those is still `"Actionable"` then?

Tests/MessageParsingTests.swift
designatednerd: Yeah it's not a check, it's to make sure the subtitle strings are being set up correctly

Tests/MessageParsingTests.swift
yuliaveres: i think it is just wrong :/

designatednerd: Seems like this would be a good time to fix that then - do you know what it's supposed to be?

yuliaveres: agree, we have to refactor it. there are afaik: text, image, actionable. it might be enough to use Payment in addition. wdyt ?

yuliaveres: and in general i believe we have overcomplicated messages models. we can definitely simplify those and make sure everything is used (f.e. this variable `messageType` which is `String` - not sure if it is used, as soon as there is `Bool` for it 🤷‍♂️  )

designatednerd: I [added a bug in Jira](https://toshiapp.atlassian.net/browse/IOS-456) and updated the comments here and at the site where it's set to indicate where things need to be fixed.

designatednerd:
@yuliaveres still good?

yuliaveres:
@designatednerd perfect !

PR #684, opened 2018-01-12
[IOS-455]: Add status messages for updated triggered by current user

yuliaveres:
We need to use `customInfo` on `TSOutgoingMessage` so while updating a group we can use `updateType` and additional info in dictionary and create a corresponding status message for this while rendering chat.
Deleting `customMessage` on `TSOutgoingMessage` has no effect as it was used previously for status messages (when composing `TSInfoMessage`) and now `customInfo` dictionary is used instead. When responding to group info request, custom message was not used as well. Removing it neither breaks group info request functionality. QAd properly.

screen shot 2018-01-12 at 12 02 02


Pods/SignalServiceKit/SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1340,14 +1340,14 @@ - (void)saveGroupMessage:(TSOutgoingMessage *)message inThread:(TSThread *)threa
         [[[TSInfoMessage alloc] initWithTimestamp:message.timestamp
                                          inThread:thread
                                       messageType:TSInfoMessageTypeGroupQuit
-                                    customMessage:message.customMessage] save];
+                                    customMessage:@"No custom message needed for current user leave message"] save];
```
designatednerd: this should probably be a comment rather than actually being in the code since if someone screws up this could get displayed to the user.

Pods/SignalServiceKit/SignalServiceKit/src/Messages/OWSMessageSender.m
```diff
@@ -1340,14 +1340,14 @@ - (void)saveGroupMessage:(TSOutgoingMessage *)message inThread:(TSThread *)threa
...
     } else {
         [[[TSInfoMessage alloc] initWithTimestamp:message.timestamp
-                                         authorId:@""
+                                         authorId:[[TSAccountManager sharedInstance] localNumber]
                                          inThread:thread
                                       messageType:TSInfoMessageTypeGroupUpdate
-                                    customMessage:message.customMessage
-                             additionalInfoString:thread.name] save];
+                                    customMessage:message.customInfo[GroupUpdateTypeSting]
+                             additionalInfoString:message.customInfo[GroupInfoString]] save];
```
designatednerd: Would it matter if either of these two properties of the dictionary became `nil`? If not, this is good. If so, might be good to Elvis-operator it with an empty string: `?: @""`

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -330,15 +330,20 @@ final class ChatInteractor: NSObject {
...
 
+            oldViewModel = thread?.groupModel
+
             guard thread != nil else { return }
             thread!.groupModel = groupModel
             thread!.save(with: transaction)
         }
 
-        sendGroupUpdateMessage(to: thread!, with: groupModel, completion: completion)
+        guard oldViewModel != nil else { return }
```
designatednerd: this should probably be a `guard let` statement so that you don't have to force-unwrap old `oldViewModel` below. 

Also probably couldn't hurt to unwrap the `thread` and log an error if either isn't present

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -290,6 +290,8 @@
 "status_type_make_public" = "%@ set the group as public.";
 "status_type_make_private" = "%@ set the group as private.";
 
+"current_user_vocative" = "You";
```
designatednerd: I had to google what the hell "vocative" meant - maybe just `current_user_pronoun`?

Pods/SignalServiceKit/SignalServiceKit/src/Messages/OWSMessageSender.m
yuliaveres: it would not matter, no. 👍

PR #685, opened 2018-01-12
[IOS-449]: Open thread after it is accepted from MessagesRequests

yuliaveres:
In this PR I push ChatViewController with accepted thread from Messages Requests View Controller.

Also small `var` <-> `let` warning fixed.

thread_accept


Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -81,7 +86,7 @@ extension MessagesRequestsViewController: UITableViewDataSource {
 
         guard let thread = dataSource.unacceptedThread(at: indexPath) else { return UITableViewCell(frame: .zero) }
 
-        var avatar = thread.avatar() ?? UIImage(named: "avatar-placeholder")
+        let avatar = thread.avatar() ?? UIImage(named: "avatar-placeholder")
```
designatednerd: :+1: thank you

PR #686, opened 2018-01-12
[IOS-397]: Hide payment buttons on a group chat

yuliaveres:
According to design decision, we need to hide payment related buttons from header on a group chat view.
screen shot 2018-01-12 at 15 04 44
screen shot 2018-01-12 at 15 04 34


Toshi/Controllers/Chat/ChatFloatingHeaderView.swift
```diff
@@ -17,6 +17,9 @@ import UIKit
 import SweetUIKit
 
 protocol ChatFloatingHeaderViewDelegate: class {
+    func shouldShowRequestButton() -> Bool
+    func shouldShowPayButton() -> Bool
```
designatednerd: why make these methods on the delegate rather than properties which can be set?

yuliaveres: no particular reason, both work fine for me.

yuliaveres: i will make variables though, seems more simple 👍

PR #687, opened 2018-01-12
iOS-453: Better handling of URLs and strings we think are URLs in Search

designatednerd:
This PR: 
- Removes a regex that was not allowing a large number of TLDs to be seen as valid links
- Adds ability to turn a search with a dot in it into a search for something with an `https://` prefix 
- Improves display of URLs to be all-lowercase. 
- Adds tests around validating whether something could be a URL or not. 
- Cleans up a couple Swift-Lint warnings in files I apparently hadn't previously touched

Before: 
simulator screen shot - iphone 8 - 2018-01-12 at 16 24 56

After: 
simulator screen shot - iphone 8 - 2018-01-12 at 16 23 57


Toshi/Extensions/String+Additions.swift
```diff
@@ -18,13 +18,23 @@ import Foundation
...
+        return UIApplication.shared.canOpenURL(url)
+    }
+    
+    var asPossibleURLString: String? {
+        let lowerSelf = self.lowercased()
 
-        if let url = URL(string: self), UIApplication.shared.canOpenURL(url) {
-            let urlPattern = "^(http|https|ftp)\\://([a-zA-Z0-9\\.\\-]+(\\:[a-zA-Z0-9\\.&amp;%\\$\\-]+)*@)*((25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9])\\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[1-9]|0)\\.(25[0-5]|2[0-4][0-9]|[0-1]{1}[0-9]{2}|[1-9]{1}[0-9]{1}|[0-9])|localhost|([a-zA-Z0-9\\-]+\\.)*[a-zA-Z0-9\\-]+\\.(com|edu|gov|int|mil|net|org|biz|arpa|info|name|pro|aero|coop|museum|[a-zA-Z]{2}))(\\:[0-9]+)*(/($|[a-zA-Z0-9\\.\\,\\?\\'\\\\\\+&amp;%\\$#\\=~_\\-]+))*$"
-            return matches(pattern: urlPattern)
+        if lowerSelf.contains("://") && !lowerSelf.hasSuffix("://") {
```
yuliaveres: i think a possible URL should contain some TLD. otherwise we can end up showing `OpenUrlButton` for `abc.` ... what should be opened in web view in this case ?

designatednerd: I agree to an extent, but we'd have to be constantly updating the list of valid TLDs in order to validate this (which is constantly expanding). As long as the URL actually validates to something that creates a URL, then SafariVC should be able to handle "this page does not exist" issues.

yuliaveres: then let SafariVC hadnle if the url is not existing, but I would prefer having smth typed after `.` without validation . wdyt ?

designatednerd: so you're saying it's got to have at least one character after the `.`? Like `dada.` will not start showing the URL but `dada.n` will?

yuliaveres: that is what I meant, yes

designatednerd: Great. Will work on that tomorrow morning.

designatednerd:
@yuliaveres requested changes pushed

yuliaveres:
Thank you.

PR #688, opened 2018-01-15
Allow synchronous access to web3.eth.accounts method in webview

faymontage:
Bootstrap the JS injection with the user's wallet address to allow synchronous access to the web3.eth.accounts property in web3.


yuliaveres:
@faymontage i assume it is tested ?

PR #689, opened 2018-01-15
iOS-447: Profile Updates, Pt. 1

designatednerd:
In this PR: 
- Redesign/refactor of the Profile View: 
simulator screen shot - iphone 5s - 2018-01-15 at 16 00 11

- Combines all profile display stuff in `ProfileViewController`, removing the separate `ProfileView` and `PersonalProfileViewController`. 
- Removes the old `ProfileActionView` and `ActionBarButton` classes as they are no longer used. 
- Favorites has now been moved over into the "more" menu. 
- There's now some minor visual distinctions between bots and regular users
- Added the basics for a Disappearing Background nav bar - basically, a nav bar whose background and title goes in and out as you scroll, but whose left and right buttons stay in place. Per Marek, this is going to be used in a bunch of places in new design. Right now the title/background part doesn't work but the buttons do. 
- A bunch of utility methods on UIStackView for easier composing of views. 

Still to come: 
~- Better handling of bots with about info~ Now in this PR
- Actual implementation of the appearing/disappearing of the Disappearing Background Nav Bar
- Redo of the Dapp view controller to follow this new pattern


Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -92,47 +90,24 @@ class ProfileViewController: UIViewController {
...
-            }
-        }
-
-        updateButton()
     }
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
+        
+        self.navigationController?.setNavigationBarHidden(false, animated: animated)
```
marijnschilling: When i press `Pay` from the profileController i see the navbar appearing just before it get's replaced with the `pay` navbar 🙈  Maybe we should do this in ViewDidDisappear and it will be solved (but maybe not)

Toshi/Extensions/UIStackView+Additions.swift
```diff
@@ -0,0 +1,140 @@
...
+extension UIStackView {
+    
+    /// Adds given view as an arranged subview standard-pins it based on the axis of the stack view.
+    /// A vertical stack view will cause a subview to be pinned to the left and right, since top and bottom are handled by the stack view
+    /// A horizontal stack view will cause a subview to be pinned to the top and bottom, since left and right are handled by the stack view.
+    ///
+    /// - Parameters:
+    ///   - view: The view to add and pin.
+    ///   - margin: Any additional margin to add. Defaults to zero. Will be the same on both sides it is applied to.
+    func addAndStandardPin(view: UIView, margin: CGFloat = 0) {
```
marijnschilling: noice! I just don't fully understand the title of the method with `pin`. Maybe I just have to get used to it, or maybe we can find a better word. I guess `anchor` is already taken by those stupid constraints 🙄  But might be good to find a synonym or use `pin` as a verb `pinning` 🤔📌

Toshi/Extensions/UIStackView+Additions.swift
```diff
@@ -0,0 +1,140 @@
...
+    /// - Parameter color: The color for the background.
+    func addBackground(with color: UIColor) {
+        let background = UIView()
+        background.backgroundColor = color
+        
+        self.addSubview(background)
+        background.edgesToSuperview()
+    }
+    
+    private static let spacerTag = 12345
```
marijnschilling: whuuud you can have static lets in extensions? I did not know that 🤔

Toshi/Extensions/UIStackView+Additions.swift
```diff
@@ -0,0 +1,140 @@
...
+            guard let indexOfArrangedSubview = self.arrangedSubviews.index(of: view) else {
+                assertionFailure("You can only do this with arranged subviews already in the view!")
+                
+                return
+            }
+            
+            let spacerViewIndex = indexOfArrangedSubview + 1
+            guard self.arrangedSubviews.count > spacerViewIndex else {
+                // There is no spacer to remove, just remove the main view.
+                removeArrangedSubview(view)
```
marijnschilling: Might be better to just do this is in an if statement since you will execute this code anyway.

Toshi/Views/DisappearingBackgroundNavBar.swift
```diff
@@ -0,0 +1,226 @@
...
+    ///
+    /// - Parameters:
+    ///   - image: The image to set on the left button as a template image.
+    ///   - accessibilityLabel: The accessibility label which should be read to voice over users describing the left button.
+    func setLeftButtonImage(_ image: UIImage, accessibilityLabel: String) {
+        let templateImage = image.withRenderingMode(.alwaysTemplate)
+        leftButton.setImage(templateImage, for: .normal)
+        leftButton.accessibilityLabel = accessibilityLabel
+        leftButton.isHidden = false
+
```
marijnschilling: soooo much space 🕊

Toshi/Views/DisappearingBackgroundNavBar.swift
```diff
@@ -0,0 +1,226 @@
...
+        let curve: UIViewAnimationOptions
+        if shouldShow {
+            targetAlpha = 1
+            curve = [.curveEaseOut]
+        } else {
+            targetAlpha = 0
+            curve = [.curveEaseIn]
+        }
+        
+        UIView.animate(withDuration: duration, delay: 0, options: curve, animations: {
```
marijnschilling: Can you just use the { } trailing closure annotation for the animation block here?

Toshi/Controllers/Profile/PersonalProfileViewController.swift
```diff
@@ -51,18 +56,18 @@ class PersonalProfileViewController: UIViewController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-
-        preferLargeTitleIfPossible(false)
+        self.navigationController?.setNavigationBarHidden(true, animated: animated)
```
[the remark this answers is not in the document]
designatednerd: 🤦‍♀️

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -13,52 +13,120 @@
...
+    
+    private let margin: CGFloat = 15
+    
+    private let tinyInterItemSpacing: CGFloat = 5
+    private let mediumInterItemSpacing: CGFloat = 10
+    private let largeInterItemSpacing: CGFloat = 20
+    private let giantInterItemSpacing: CGFloat = 40
+    private let belowTableViewStyleLabelSpacing: CGFloat = 8
+    
+    private let avatarSide: CGFloat = 60
```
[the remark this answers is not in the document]
designatednerd: I initially used `avatarSize` but then realized that would imply that the property is a `CGSize`. Maybe `avatarSideLength`? Or am I overthinking it?

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -13,52 +13,120 @@
...
-            usernameLabel.text = user.displayUsername
-        } else {
-            nameLabel.text = user.displayUsername
-            usernameLabel.text = nil
+    weak var navBarDelegate: DisappearingBackgroundNavBarDelegate?
+    
+    private let margin: CGFloat = 15
+    
+    private let tinyInterItemSpacing: CGFloat = 5
+    private let mediumInterItemSpacing: CGFloat = 10
```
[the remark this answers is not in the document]
designatednerd: Yeah - i think I thought there was one more than there actually is, I'll update `tiny` to `small`

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -13,52 +13,120 @@
...
+        view.delaysContentTouches = false
+        if #available(iOS 11, *) {
+            view.contentInsetAdjustmentBehavior = .never
         }
+        
+        return view
+    }()
+    
+    /// The view containing all content within the scroll view.
+    private lazy var containerView = UIView()
```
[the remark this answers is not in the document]
designatednerd: a) Good point - I'll see if I can make that change.
b) triple slashes indicates a docu-comment, which makes this show up when you option-click the property: 

screen shot 2018-01-15 at 9 34 13 pm

with just a double slash, it doesn't show up: 

screen shot 2018-01-15 at 9 35 31 pm

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -13,52 +13,120 @@
...
-    }
-
-    lazy var avatarImageView = AvatarImageView()
+    private lazy var avatarImageView = AvatarImageView()
 
     lazy var nameLabel: UILabel = {
         let view = UILabel()
         view.numberOfLines = 0
-        view.font = Theme.preferredTitle2()
+        view.font = Theme.displayName()
```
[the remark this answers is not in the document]
designatednerd: The minimum font size is different than `preferredTitle2`, that's why I created a new one. This is in Marek's new component library in Sketch, so it should be consistent throughout a number of screens. 

Updating the name to reflect dynamic type though 👍

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -92,119 +191,91 @@ class ProfileView: UIView {
...
-        view.setTitle("Rate this user", for: .normal)
+        view.setTitle(Localized("profile_rate_user"), for: .normal)
         view.setTitleColor(Theme.tintColor, for: .normal)
         view.setTitleColor(Theme.greyTextColor, for: .highlighted)
         view.titleLabel?.font = Theme.preferredRegular()
         view.titleLabel?.adjustsFontForContentSizeCategory = true
         view.clipsToBounds = true
 
-        view.addTarget(self, action: #selector(self.didTapRateUser), for: .touchUpInside)
+        view.addTarget(self, action: #selector(didTapRateUserButton), for: .touchUpInside)
```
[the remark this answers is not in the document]
designatednerd: To differentiate between a delegate method call and an internal method call - `didTapRateUser(in profileView: ProfileView)` is the delegate method, and `didTapRateUserButton` is the internal method which is called when the button itself is called.

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -92,119 +191,91 @@ class ProfileView: UIView {
...
+    private lazy var reputationBottomSeparatorView = BorderView()
+        
+    // MARK: - Setter
+    
+    func setProfile(_ user: TokenUser) {
+        if !user.name.isEmpty {
+            nameLabel.text = user.name
+        } else {
+            nameLabel.text = nil
+        }
```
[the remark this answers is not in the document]
designatednerd: yeahhhh probably there was more stuff in there before and also I hate ternaries but this is a reasonable one.

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -92,119 +191,91 @@ class ProfileView: UIView {
...
+        
+        usernameLabel.text = user.displayUsername
+        disappearingNavBar.setTitle(user.nameOrDisplayName)
+
+        if user.isApp {
+            if payButton.superview != nil {
+                profileDetailsStackView.removeArrangedSubviewAndSpacingAfter(arrangedSubview: payButton)
```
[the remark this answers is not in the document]
designatednerd: Hmm... that'd probably require passing the profile as a parameter to all the constructors, but that *might* be doable.

Toshi/Controllers/Profile/ProfileView.swift
```diff
@@ -92,119 +191,91 @@ class ProfileView: UIView {
...
+        aboutContentLabel.text = user.about
+        locationContentLabel.text = user.location
+        
+        if !aboutContentLabel.hasContent && !locationContentLabel.hasContent {
+            profileDetailsStackView.removeArrangedSubviewAndSpacingAfter(arrangedSubview: aboutContentLabel)
+            profileDetailsStackView.removeArrangedSubviewAndSpacingAfter(arrangedSubview: locationContentLabel)
+            profileDetailsStackView.removeArrangedSubviewAndSpacingAfter(arrangedSubview: aboutTopSeparatorView)
+        }
+        
+        AvatarManager.shared.avatar(for: user.avatarPath) { [weak self] image, _ in
```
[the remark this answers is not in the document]
designatednerd: I'm ok with it because the call to the API is an implementation detail - 9 times out of 10 this should come from the cache with no API call.

However this was also one of the things making me feel like I should refactor this all into a view controller, so ¯\_(ツ)_/¯

Toshi/Controllers/Profile/ProfileViewController.swift
designatednerd: This is where you're supposed to do it - `viewDidDisappear` doesn't have the `animated` parameter. I agree it looks a little rough on iOS 11 when you're coming from an expanded nav bar, but I'm not sure there's a better way to do it.

designatednerd: Ooh, maybe i need to do it before i call super...

Toshi/Extensions/UIStackView+Additions.swift
designatednerd: Maybe `addWithStandardConstraints`?

Toshi/Extensions/UIStackView+Additions.swift
designatednerd: yeah you can't have non-computed `var`s or instance-level `let`s, but `static let`s are no problemo.

Toshi/Extensions/UIStackView+Additions.swift
designatednerd: but guard statements are so much more fun! (kidding, on it)

Toshi/Views/DisappearingBackgroundNavBar.swift
designatednerd: Yep! Missed that because it had the `completion` parameter before that is now omitted.

designatednerd:
OK, some changes pushed but before I go too far down any road, you and I should chat via slack re: combining this down to a single view controller that is set up with a passed in `profile` as a parameter.

Toshi/Controllers/Profile/ProfileView.swift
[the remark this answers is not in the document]
marijnschilling: Ah yeah i get it... I always just use size 😬 You could use diameter if you want to be all correct right?

Toshi/Controllers/Profile/ProfileView.swift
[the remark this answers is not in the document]
marijnschilling: oooh is `profileDetailsStackView` a local var here? I think you could also solve it with readable bool names I mention in the comment below here

Toshi/Views/DisappearingBackgroundNavBar.swift
designatednerd: Tried this but had to revert it because it wouldn't compile - apparently you can only omit the `completion` parameter if you've still got the parens on there 😢

marijnschilling: ah well! You tried ✌️

Toshi/Theme/Theme.swift
```diff
@@ -18,9 +18,31 @@ import SweetUIKit
...
+    static let smallInterItemSpacing: CGFloat = 5
+    
+    /// A medium amount of inter-item spacing. Current value: 10
+    static let mediumInterItemSpacing: CGFloat = 10
+    
+    /// A large amount of inter-item spacing. Current value: 20
+    static let largeInterItemSpacing: CGFloat = 20
+    
+    /// A giant amount of inter-item spacing. Current value: 40
+    static let giantInterItemSpacing: CGFloat = 40
```
designatednerd: @marijnschilling heads up that I moved these here since they're used consistently throughout Marek's updated designs. This way we can avoid defining them for every view/vc.

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -49,130 +199,410 @@ class ProfileViewController: UIViewController {
...
+        if isProfileEditable {
+            disappearingNavBar.setTitle(Localized("profile_me_title"))
+        } else {
+            disappearingNavBar.setTitle(profile.nameOrDisplayName)
+        }
+        
+        //TODO: Remove!
+        if isBotProfile && profile.name == "Spambot" {
+            aboutContentLabel.text = "We eat ham and jam and spam a lot, and I like to push the pram a lot"
+        }
```
designatednerd: Leaving this in for a hot second so you can test @marijnschilling - but note to self to remove it 😛

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -49,130 +199,410 @@ class ProfileViewController: UIViewController {
...
+        
+        if aboutStackView.superview != nil {
```
marijnschilling: I don't fully understand why you check for this here

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -49,130 +199,410 @@ class ProfileViewController: UIViewController {
...
-        super.viewWillDisappear(animated)
-
-        preferLargeTitleIfPossible(true)
+    
+    // MARK: - Action Targets
+    
+    // MARK: Button targets
+    
+    @objc private func didTapMessageButton() {
+        // create thread if needed
```
marijnschilling: Is this comment really necessary since the method is called `getOrCreateThread`?

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -49,130 +199,410 @@ class ProfileViewController: UIViewController {
...
+    @objc private func didSelectMoreButton() {
+        presentMoreActionSheet()
+    }
+    
+    // MARK: Action sheet targets
+    
+    private func didSelectBlockedState(_ shouldBeBlocked: Bool) {
+        if shouldBeBlocked {
+            // Alert to confirm before actually blocking.
+            let alert = UIAlertController(title: Localized("block_alert_title"), message: Localized("block_alert_message"), preferredStyle: .alert)
```
marijnschilling: As you might have realised in my PR comments I'm not the biggest fan of comments since they can easily get outdated. I think here you could name the alert `blockConfirmationAlert` which would convey the same information, just without the change it will get outdated.

Toshi/Theme/Theme.swift
```diff
@@ -18,9 +18,31 @@ import SweetUIKit
...
         return 1 / UIScreen.main.scale
     }
+    
+    /// The default height of buttons. Current value: 44
+    static let defaultButtonHeight: CGFloat = 44
+    
+    /// The default margin from the sides of the view. Current value: 15
+    static let defaultMargin: CGFloat = 15
+
+    /// The default height of an avatar (implicitly also the width). Current value: 60
```
marijnschilling: I think the comment here about the width is necessary, but I fell the other comments are redundant because of their names. And I wonder why you mention what the current value is, any benefit in that that im not seeing? Right now i can only think of the downside that if it's not the same as the value itself you get very much confused.

Toshi/Theme/Theme.swift
marijnschilling: That is great!

Toshi/Theme/Theme.swift
```diff
@@ -201,6 +223,10 @@ extension Theme {
         return dynamicType(for: regular(size: 16), withStyle: .title3, inSizeRange: range)
     }
     
+    static func preferredDisplayName(range: ClosedRange<CGFloat> = 25...35) -> UIFont {
```
marijnschilling: 🎉

marijnschilling:
I think you did solve it quite nicely with the setup of the different arrangements of the views.

I do wonder if itwould be possible to just distinguish in the `viewDidLoad()` between the different cases: `isBotProfile`, `isForCurrentUserProfile`, `isEditable` and normal style.

So have 4 distinct setup methods for all of these... Hmm I'm not sure because it would make the code very redudant.. But I also can see it might be hard to follow which viewes are and which views arent placed on screen the way it is now, since it's a bit scattered over the code.

Toshi/Theme/Theme.swift
designatednerd: Good point on the redundancy of the other comments. The benefit I was thinking of was having that comment be visible at the call site if you were trying to figure out what size something was, but option-click and command-click are the same number of clicks, and the comment can easily get out of sync. 

Removing the rest.

Toshi/Controllers/Profile/ProfileViewController.swift
designatednerd: I'm usually a lot more into comments, but what you're suggesting also follows the pattern of the rest of the code here, so 👍

Toshi/Controllers/Profile/ProfileViewController.swift
designatednerd: basically, if the stack view is used, about and location are all in one stack view and should be considered together. if not, only the about view's visibility should be affected, and only by the about content.

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -19,27 +19,177 @@ import CoreImage
...
         let appDelegate = UIApplication.shared.delegate as? AppDelegate
-
         return appDelegate?.messageSender
     }
+    
+    private let belowTableViewStyleLabelSpacing: CGFloat = 8
+    
+    private lazy var disappearingNavBar: DisappearingBackgroundNavBar = {
+        let navBar = DisappearingBackgroundNavBar(delegate: self)
+        navBar.setupLeftAsBackButton()
```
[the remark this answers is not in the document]
designatednerd: Aiming for flexibility: There are other potential things which could be on the left, like a Cancel button. That's why I made it `leftButton` instead of `backButton`, and so that's why I want to specify that you're setting up the left button as the back button.

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -19,27 +19,177 @@ import CoreImage
...
 
 class ProfileViewController: UIViewController {
+    
+    var profile: TokenUser {
+        didSet {
+            configureForCurrentProfile()
+        }
+    }
+    
+    private let isReadOnlyMode: Bool
```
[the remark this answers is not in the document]
designatednerd: The user can view their profile from both the settings page and from the main browse section (if they are in most recent public users). 

`readOnlyMode` is set to false when opening from the main browse section to indicate that the profile can't be edited from the current flow, even if it is the current user's profile. 

`isProfileEditable` i'm gonna flip-flop the order of the checks on - first it will check whether the profile is in read only mode, then it'll do an additional check that this is the current user's profile, since we don't want someone editing another user's profile. 

Here are some screenshots that will hopefully help clear this up: 

**This is what another user's profile looks like:**
simulator screen shot - iphone 5s - 2018-01-16 at 19 31 15





**This is what your profile looks like in read-only mode:**
simulator screen shot - iphone 5s - 2018-01-16 at 19 29 44





**And this is what it looks like in editable mode:**
simulator screen shot - iphone 5s - 2018-01-16 at 19 29 47

Toshi/Controllers/Profile/ProfileViewController.swift
designatednerd: From old code, killing it with 🔥

designatednerd:
Yeah this code is definitely a balancing act between confusion and redundancy. 😛 

Personally i've gotten bitten enough times by redundancy that I try to centralize everything I can, though I try to name things as clearly as possible when doing so.

Toshi/Extensions/UILabel+Additions.swift
```diff
@@ -0,0 +1,26 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+extension UILabel {
+    
+    /// - Returns: true if there is text of length greater than zero, false if not
+    var hasContent: Bool {
+        guard let text = self.text else { return false }
```
yuliaveres: what if text is empty ? or we check for nil only ?

Toshi/Extensions/UILabel+Additions.swift
```diff
@@ -0,0 +1,26 @@
...
+        
+        return text.length > 0
```
yuliaveres: oh god. sorry

designatednerd: heh, fair point though, that should be `isEmpty` 😄

Toshi/Views/BorderView.swift
```diff
@@ -0,0 +1,33 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import TinyConstraints
+import UIKit
+
+class BorderView: UIView {
```
yuliaveres: do you ever subclass it ? if not - mind making it `final` ?

Toshi/Views/DisappearingBackgroundNavBar.swift
```diff
@@ -0,0 +1,232 @@
...
+import UIKit
+
+protocol DisappearingBackgroundNavBarDelegate: class {
+    
+    func didTapLeftButton(in navBar: DisappearingBackgroundNavBar)
+    func didTapRightButton(in navBar: DisappearingBackgroundNavBar)
+}
+
+/// A view to allow a fake nav bar that can appear and disappear as the user scrolls, but allowing its buttons to stay in place.
+class DisappearingBackgroundNavBar: UIView {
```
yuliaveres: could you make it `final` ?

designatednerd: i need a little electric shock every time i forget to make a class final

Toshi/Views/DisappearingBackgroundNavBar.swift
```diff
@@ -0,0 +1,232 @@
...
+    
+    private let animationSpeed = 0.25
+    private let interItemSpacing: CGFloat = 8
+    
+    weak var delegate: DisappearingBackgroundNavBarDelegate?
+    
+    static let defaultHeight: CGFloat = 64
+    
+    private lazy var leftButton: UIButton = {
+        let button = UIButton(withAutoLayout: true)
```
yuliaveres: hmm could you please double check which size will button have depending on content ? i guess it will resize. and if image is small, button will be tricky to tap.. as far as i remember it should be not less then `44x44` ..

designatednerd: Good point - I'll set that as min/width height on both buttons.

designatednerd:
@marijnschilling @yuliaveres anything further?

PR #690, opened 2018-01-17
[IOS-457]: Fix missing recipients fetch for incoming thread

yuliaveres:
Previously we did not process inserted thread as new indexPath was never correct. We should use all threads collection without any filtering while listening for incoming DB changes. So I added `allThreadsMappings`  and retrieve correct thread while processing new or updated one on `ThreadsDataSource`. (Also I noticed a warning of calling avatarManager's completion being invoked in background, fixed it here.)


Toshi/Avatars/AvatarManager.swift
```diff
@@ -124,7 +124,10 @@ final class AvatarManager: NSObject {
             }
 
             guard let url = URL(string: path), let teapot = self.teapot(for: url) else {
-                completion?(nil, path)
+                DispatchQueue.main.async {
+                    completion?(nil, path)
+                }
```
designatednerd: Good catch - looking at the surrounding code:
- The cached completion above this should either be dispatched to main, or never sent to the global queue in the first place. 
- The bailout on `self` being nil should probably also be fired on the main queue.

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -211,31 +214,45 @@ final class ThreadsDataSource: NSObject {
...
+        }
 
-                    guard let user = foundUser else { return }
+        return thread
+    }
 
-                    AvatarManager.shared.downloadAvatar(for: user.avatarPath)
+    func updateNewThreadRecepientsIfNeeded(_ thread: TSThread) {
+        if let contactId = thread.contactIdentifier() {
+            guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { fatalError("Can't access App delegate") }
```
designatednerd: is this guaranteed to be called on the main thread?

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -92,16 +92,22 @@ extension MessagesRequestsViewController: UITableViewDataSource {
 
         if thread.isGroupThread() {
             title = thread.name()
-        } else if let recipient = thread.recipient() {
+        } else if let recipient = ThreadCellConfigurator.recipient(for: thread) {
```
designatednerd: So reading through this, I'm still a little confused as to why `thread.recipient()` would be nil but this other method would be non-nil. Is this something where the `thread` is still being parsed, so the recipient or recipients could be nil?

yuliaveres: the `recipient()` is outdated and i dnt use it now. will delete 👍 
(it was looking for a user in a different collection. i need to look in contacts).

Toshi/Avatars/AvatarManager.swift
yuliaveres: it should be sent to the global thought, as we do want to access cache in background, very heavy operation and in my opinion we should use different cache instead of this one.
I will send completion above to the main thread as well 👍

yuliaveres: all those calls in background are because of the `awesome` cache 😢

Toshi/Controllers/Recent/ThreadsDataSource.swift
yuliaveres: nope

Toshi/Avatars/AvatarManager.swift
designatednerd: "awesome"

Toshi/Controllers/Recent/ThreadsDataSource.swift
designatednerd: I see you added a `dispatch` to main at the call site - would it make more sense to add that here since whatever's calling this would need to make sure it got sent to the main queue?

yuliaveres:
@designatednerd please see again

PR #691, opened 2018-01-17
iOS-447: Profile Updates, pt 2 + Disappearing Nav Bar

designatednerd:
In this PR: 
- Dapp design is updated to follow new style
- ~`DisappearingNavBarScrollable` protocol added to let anything have a disappearing nav bar and scroll view which it can set up and adjust.~
- `DisappearingNavBarViewController` added to handle setup and management of scrolling changing things on the view (I kept having to add the same delegate code to both view controllers and then realized this is almost certainly not going to ever not be the root of a ViewController, so a subclass is actually warranted here). 
- Nav bar disappearing adjusted from animations to changing opacity based on percentage of the background target view which has scrolled past
- Title disappearing adjusted to be separate from background and to also adjust placement of the label. 
- Dapp VC now ~conforms to `DisappearingNavBarScrollable`~ inherits from `DisappearingNavBarViewController`
- Profile VC now ~conforms to `DisappearingNavBarScrollable`~ inherits from `DisappearingNavBarViewController`

Here's what the final product looks like (with temporarily triple-length Dapp about text): 
disappearing_nav_bar_final


designatednerd:
~425 additions and ~150 deletions is a big PR?!

designatednerd:
Hold on reviewing this, i'm trying to fix some iPhone X stuff

Toshi/Controllers/Browse/DappViewController.swift
```diff
@@ -100,30 +83,62 @@ final class DappViewController: UIViewController {
...
     // MARK: - Configuration
     
     private func configure(for dapp: Dapp) {
-        title = dapp.name
-        mainLabel.text = dapp.name
+        navBar.setTitle(dapp.name)
+        titleLabel.text = dapp.name
         descriptionLabel.text = dapp.description
+        
+        //TODO: Remove temp for loop
```
designatednerd: note to self: do not merge without removing this

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -458,21 +419,42 @@ class ProfileViewController: UIViewController {
...
     private func configureForCurrentProfile() {
         // Set nils for empty strings to make the views collapse in the stack view
         nameLabel.text = profile.name.isEmpty ? nil : profile.name
         aboutContentLabel.text = profile.about.isEmpty ? nil : profile.about
         locationContentLabel.text = profile.location.isEmpty ? nil : profile.location
-        
+
+        if isBotProfile {
+            aboutContentLabel.text = "Some text about this bot which would look totally different in production."
+        }
```
designatednerd: ...also remove this before merging

designatednerd:
OK @marijnschilling this is now ready for review. Please note that I updated the description to clarify some of the changes I made.

Toshi/Controllers/DisappearingNavBarViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+        contentView.width(to: scrollView)
+
+        addScrollableContent(to: contentView)
+    }
+
+    /// Called when scrollable content should be programmatically added to the given container view.
+    /// The size of the views added to the container will determine the scrollable area of the scroll view.
+    ///
+    /// - Parameter contentView: The content view to add scrollable content to.
+    func addScrollableContent(to contentView: UIView) {
```
marijnschilling: I did not understand this method. Is this just where you need to setup the entire view? And the constraints of the views added on that view?

Not sure how it could be better either. But you can see my struggles in this branch: test/marijn-nav-bar
I tried on the QRCodeView which you can reach from `Me` > `My QR Code`

designatednerd: Yeah, you're correct - the basic idea is that the `contentView` should size itself based on the views and constraints added to it, then because it's pinned to the scrollview that will dictate the content size. 

Looking at what you're working on, I see a few things: 

- `loadView` is overridden and the view is set to a `UIScrollView`
- It doesn't seem like there's enough content for this to be able to scroll.
- in `addScrollableContent` you're adding constraints to the content view again, which isn't necessary (and also the height constraint is probably fucking with its ability to automatically determine its own height).  
- The subtitle label is being pinned to the VC's `view` not to its superview (which would be the content view), which is probably fucking with things.
- I'm not completely clear why the top and bottom layout guides are necessary instead of pinning 

I took a swipe at where you started from and it seems like it's working - let me know if that help explain why what you were doing wasn't working. 

I also clarified the documentation on this method based on your experience - let me know if that makes the technique clearer.

Toshi/Controllers/Browse/BrowseAllViewController.swift
```diff
@@ -86,8 +86,10 @@ class BrowseAllViewController: UITableViewController {
 
     override func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
         if let contact = searchResults.element(at: indexPath.row) as? TokenUser {
+            dismissSearch()
             Navigator.push(ProfileViewController(profile: contact))
         } else if let dapp = searchResults.element(at: indexPath.row) as? Dapp {
+            dismissSearch()
```
yuliaveres: can we dismiss once ?

Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -369,10 +370,20 @@ extension BrowseViewController: BrowseCollectionViewCellSelectionDelegate {
     func didSelectItem(at indexPath: IndexPath, collectionView: SectionedCollectionView) {
         if let section = items.element(at: collectionView.section) {
             if let user = section.element(at: indexPath.item) as? TokenUser {
+                dismissSearch()
```
yuliaveres: can we dismiss once ?

Toshi/Controllers/Browse/DappViewController.swift
yuliaveres: can we remove it now ?

Toshi/Controllers/Browse/BrowseAllViewController.swift
designatednerd: fair, though i'm going to add `else` cases here with assertion failures for unhandled types

PR #692, opened 2018-01-17
[IOS-436]: Add wallet qr code test

yuliaveres:
Test wallet QR code image.


Tests/CerealTests.swift
```diff
@@ -67,4 +67,13 @@ class CerealSignEthereumTransactionWithWalletTests: XCTestCase {
         XCTAssertNil(cereal.signEthereumTransactionWithWallet(hex: transactionSkeleton))
 
     }
+
+    func testWalletQRCodeImage() {
+        let expectedQRCodeImage = UIImage.imageQRCode(for: "\(QRCodeController.paymentWithAddressPath)\(cereal.paymentAddress)", resizeRate: 20.0)
+        if let expectedImageData = UIImagePNGRepresentation(expectedQRCodeImage), let resultQRCodeImageData = UIImagePNGRepresentation(cereal.walletAddressQRCodeImage(resizeRate: 20.0)) {
```
designatednerd: This is testing that the image data is identical, which is a good first step. 

The next step would be detecting whether the `resultQRCodeData` is a) Recognizable as a QR code and b) once recognized, contains the expected data. 

Obnoxiously Socratic Question time: How would you detect if an image contains a QR code? How would you get the data from that QR code once it's detected?

yuliaveres: if some expected image data which i know for 100% is QR code, is identical to the result image data - is that possible that the result image is not QR code ?

designatednerd: In theory, you're correct. In practice, by comparing image data to image data, you're testing that calling `UIImage.QRCode` with the same parameters works the same every time. 

That's a reasonable thing to double check, but I think more critical to the user is "When this image which is output by UIImage gets scanned by another QR code reader, is it going to have the correct information?"

yuliaveres: not sure how can I trigger "scanning"  🤔  @designatednerd

designatednerd: Let me put this a slightly more obvious way: If you use the built-in tools of iOS to see if a UIImage has a QR code in it and what that QR code says, are the results correct?

yuliaveres: we use `AVCaptureMetadataOutputObjectsDelegate` which triggers delegate callback with  already scanned result which is `String`.

designatednerd: Right, but that's for a live `AVCaptureSession`. How would you do that for `UIImage`? 

[let me know if this is getting annoying and you just want me to tell you]

designatednerd: There's a good description of it in the "Detecting QR codes" section here: https://www.shinobicontrols.com/blog/ios8-day-by-day-day-13-coreimage-detectors

Basically, use `CIDetector` with `CIDetectorTypeQRCode` on `image.CIImage`.

yuliaveres: oh thank you !

yuliaveres:
@designatednerd please take another look

PR #694, opened 2018-01-17
[IOS-448]: Payment router flow

marijnschilling:
Payment router will be responsible for payment flow. You can initialise a PaymentRouter from anywhere with any values and it will handle payment. If the value is not given it will show the 'PaymentValueController' where the user can put in the value. Is the recipientAddress not given? It will present the `PaymentAddressViewController` where the user can fill that in. Once all values are present to make a payment it will show the `PaymentConfirmationViewController` where the user sees an overview of the payment, the freshly fetched estimated network fees and where it can confirm the payment. The `PaymentConfirmationViewController` has a `PaymentManager` that deals with the payment and the fetching of the fees. When the payment is cancelled or succeeds the Router will get back to the Controller that initialy intialised the PaymentRouter to tell them all about it. ✌️


Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -0,0 +1,104 @@
...
+
+
+import Foundation
+
+protocol PaymentRouterDelegate: class {
+    func paymentRouterDidCancel(paymentRouter: PaymentRouter)
+    func paymentRouterDidSucceedPayment(paymentRouter: PaymentRouter)
+}
+
+class PaymentRouter {
```
yuliaveres: please go over all classes and consider making them `final`

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -0,0 +1,104 @@
...
+    weak var delegate: PaymentRouterDelegate?
+
+    private var paymentViewModel: PaymentViewModel
+
+    init(withAddress address: String? = nil, andValue value: NSDecimalNumber? = nil) {
+      self.paymentViewModel = PaymentViewModel(recipientAddress: address, value: value)
+    }
+
+    // I purposefully created this method so the caller is aware that this object will present a VC
+    func present() {
```
yuliaveres: hmm .. would not `proceed` fit more for Router ? ...

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -0,0 +1,104 @@
...
+        //here should be decided what controller should be presented first
+        guard let value = paymentViewModel.value else {
```
yuliaveres: lets check if the value is valid (> 0.0)

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -0,0 +1,104 @@
...
+            presentPaymentValueController()
+            return
+        }
+
+        guard let address =  paymentViewModel.recipientAddress else {
```
yuliaveres: lets check if the address is valid (`EthereumAddress.validate(address)`)

Toshi/Controllers/Payments/PaymentRouter/PaymentViewModel.swift
```diff
@@ -0,0 +1,28 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+struct PaymentViewModel {
+    var recipientAddress: String?
```
yuliaveres: lets make those `private(set)` ?

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    private func presentPaymentConfirmationController(withValue value: NSDecimalNumber, andRecipientAddress address: String) {
+        let paymentConfirmationController = PaymentConfirmationViewController(withValue: value, andRecipientAddress: address)
+        paymentConfirmationController.delegate = self
+
+        presentViewControllerOnNavigator(paymentConfirmationController)
+    }
+
+    private func presentViewControllerOnNavigator(_ controller: UIViewController) {
+
```
yuliaveres: we should check if it is a confirmation controller `if controller is PaymentConfirmationViewController`in the first place and if so present modally as well

yuliaveres:
@designatednerd  ready !

designatednerd:
@yuliaveres merge conflicts 😢

yuliaveres:
on it

Tests/EthereumConverterTests.swift
```diff
@@ -25,14 +25,6 @@ class EthereumConverterTests: QuickSpec {
...
                 }
 
-                it("returns the string representation of the ethereum network fees") {
-                    let ethereumNetworkFeeString = EthereumConverter.estimatedEthereumNetworkFeeString(for: wei, exchangeRate: exchangeRate)
-
-                    let dollarSting = String(format: "$100%@00 USD", TokenUser.current?.cachedCurrencyLocale?.decimalSeparator ?? ".")
-                    let feeString = String(format: Localized("payment_estimated_fee"), dollarSting, "1.0000 ETH")
-                    expect(ethereumNetworkFeeString).to(equal(feeString))
-                }
-
```
designatednerd: why is this test removed?

Toshi/APIClients/idAPIClient.swift
```diff
@@ -543,6 +542,37 @@ typealias DappCompletion = (_ dapps: [Dapp]?, _ error: ToshiError?) -> Void
         }
     }
 
+    func findUserWithPaymentAddress(_ paymentAddress: String, completion: @escaping ((TokenUser?) -> Void)) {
```
designatednerd: can we please add some tests on this method?

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -113,38 +113,14 @@ final class ChatInteractor: NSObject {
...
 
-    func sendPayment(in value: NSDecimalNumber?, completion: ((Bool) -> Void)? = nil) {
-        guard let value = value else {
-            return
-        }
-
-        guard let tokenId = self.thread.contactIdentifier() else {
-            return
-        }
+    // need to write a test for this
```
designatednerd: heh, oops. ^ This comment can go away as soon as the test is written

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -748,18 +736,16 @@ extension ChatViewController: MessagesPaymentCellDelegate {
             messageText = String(format: Localized("payment_request_confirmation_warning_message_fallback"), self.thread.name())
         }
 
-        if let parameters = transactionParameter(for: indexPath) {
-            showActivityIndicator()
+        guard let parameters = transactionParameter(for: indexPath), let value = parameters["value"] as? NSDecimalNumber else { return }
```
designatednerd: is there a way we could avoid hard-coding this string?

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -748,18 +736,16 @@ extension ChatViewController: MessagesPaymentCellDelegate {
...
 
-            PaymentConfirmation.shared.present(for: parameters, title: Localized("payment_request_confirmation_warning_title"), message: messageText, presentCompletionHandler: { [weak self] in
+        paymentRequestActiveCell = cell
 
-                self?.hideActivityIndicator()
-            }, approveHandler: { [weak self] transaction, error in
+        viewModel.interactor.retrieveRecipientAddress { [weak self] address in
+            self?.paymentRouter = PaymentRouter(withAddress: address, andValue: value)
```
designatednerd: Might be worth pulling out `strongSelf` here so `self` doesn't get dealloced in the middle of this setup. Either that or just putting all this stuff into a separate method so you only have to `self?.` once when calling that method.

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -900,32 +886,58 @@ extension ChatViewController: ChatFloatingHeaderViewDelegate {
...
-        let paymentController = PaymentController(withPaymentType: .send, continueOption: .send)
-        paymentController.delegate = self
-        
-        let navigationController = PaymentNavigationController(rootViewController: paymentController)
-        Navigator.presentModally(navigationController)
+        viewModel.interactor.retrieveRecipientAddress { [weak self] address in
+            self?.paymentRouter = PaymentRouter(withAddress: address)
+            self?.paymentRouter?.delegate = self
+            self?.paymentRouter?.userInfo = self?.thread.recipient()?.userInfo
+            self?.paymentRouter?.present()
```
designatednerd: oh yeah this is basically the same thing as above, this can definitely be pulled out as a separate method

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -92,7 +92,7 @@ extension MessagesRequestsViewController: UITableViewDataSource {
 
         if thread.isGroupThread() {
             title = thread.name()
-        } else if let recipient = ThreadCellConfigurator.recipient(for: thread) {
+        } else if let recipient = thread.recipient() {
```
designatednerd: I think this got put back by accident ^^

Tests/EthereumConverterTests.swift
yuliaveres: Marijn removed it as she made some changes. she is going to change the test and re-add it

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -0,0 +1,474 @@
+// Copyright (c) 2017 Token Browser, Inc
```
designatednerd: I probably should have caught this when I was working on it but: It's now 2018 😄

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -0,0 +1,474 @@
...
+    dapp(info: DappInfo)
+}
+
+class PaymentConfirmationViewController: UIViewController {
+
+    weak var delegate: PaymentConfirmationViewControllerDelegate?
+
+    let paymentManager: PaymentManager
+    private var recipientType: RecipientType
+    private var shouldSendSignedTransaction: Bool = true
```
designatednerd: Does this need to be a `var` with a default value or can it be a `let` since it's being set by the init method?

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -0,0 +1,474 @@
...
+            })
+
+            return
+        }
+
+        paymentManager.sendPayment { [weak self] error, transactionHash in
+            guard let weakSelf = self else { return }
+
+            guard error == nil else {
+                let alert = UIAlertController(title: "Error completing transaction", message: (error?.description ?? ToshiError.genericError.description), preferredStyle: .alert)
```
designatednerd: 🚨 unlocalized user-facing strings here and below 🚨

Toshi/Controllers/Payments/PaymentController.swift
```diff
@@ -36,7 +36,8 @@ enum PaymentControllerPaymentType {
 enum PaymentControllerContinueOption {
     case next
     case send
-    
+
+    // this should always be next now
```
designatednerd: if it should always be next do you still need the enum or this property?

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
+import UIKit
+
+typealias PaymentInfo = (fiatString: String, estimatedFeesString: String, totalFiatString: String, totalEthereumString: String, balanceString: String, sufficientBalance: Bool)
```
designatednerd: Why are we doing these as `typealias` instead of structs? I did it as well to follow the pattern but thinking on it, why not use a struct?

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    lazy var parameters: [String: Any] = {
+        return [
+            "from": Cereal.shared.paymentAddress,
+            "to": paymentAddress,
+            "value": value.toHexString
+        ]
+    }()
+
+    init(withValue value: NSDecimalNumber, andPaymentAddress address: String) {
```
designatednerd: naming nit: Generally the `with` pattern for a first parameter and the `and` pattern for a second parameter is used only in Objective-C, unless you're doing it as `init(with:and:)`. Personally I find `init(value: paymentAddress:)` to to be a solid combination of easy to read and Swifty.

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+                let gasPriceValue = NSDecimalNumber(hexadecimalString: gasPrice)
+                let gasValue = NSDecimalNumber(hexadecimalString: gas)
+
+                let fee = gasPriceValue.decimalValue * gasValue.decimalValue
+                let decimalNumberFee = NSDecimalNumber(decimal: fee)
+
+                let exchangeRate = ExchangeRateClient.exchangeRate
+
+
+                //WARNING: we need to test these values that the correspond with each other
```
designatednerd: a) 2 blank lines above this 
b) What needs to correspond with what? Is there a way to do that when this is so deep in here, or do you just mean use it in practice rather than automated-test it?

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+                let fiatString = EthereumConverter.fiatValueStringWithCode(forWei: weakSelf.value, exchangeRate: exchangeRate)
+                let estimatedFeesString = EthereumConverter.fiatValueStringWithCode(forWei: decimalNumberFee, exchangeRate: exchangeRate)
+
+                let totalWei = weakSelf.value.adding(decimalNumberFee)
+                let totalFiatString = EthereumConverter.fiatValueStringWithCode(forWei: totalWei, exchangeRate: exchangeRate)
+                let totalEthereumString = EthereumConverter.ethereumValueString(forWei: totalWei)
+
+                /// We don't care about the cached balance since we immediately want to know if the current balance is sufficient or not.
+                EthereumAPIClient.shared.getBalance(cachedBalanceCompletion: { _, _ in }, fetchedBalanceCompletion: { fetchedBalance, error in
```
designatednerd: why not pass `nil` to the `cachedBalanceCompletion`?

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    func fetchPaymentInfo(completion: @escaping ((_ paymentInfo: PaymentInfo) -> Void)) {
+        EthereumAPIClient.shared.transactionSkeleton(for: parameters) { [weak self] skeleton, error in
+            guard error == nil else {
+                // Handle error
+                return
+            }
+
+            if let gasPrice = skeleton.gasPrice, let gas = skeleton.gas, let transaction = skeleton.transaction {
```
designatednerd: I would recommend doing this as a `guard let` since there's not really anything that happens below if these unwraps fail

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    init(withValue value: NSDecimalNumber, andPaymentAddress address: String) {
+       self.value = value
+       self.paymentAddress = address
+    }
+
+    func fetchPaymentInfo(completion: @escaping ((_ paymentInfo: PaymentInfo) -> Void)) {
+        EthereumAPIClient.shared.transactionSkeleton(for: parameters) { [weak self] skeleton, error in
+            guard error == nil else {
+                // Handle error
```
designatednerd: should `paymentInfo` be optional so at least you can fire off the completion block without the retrieved info?

Toshi/Controllers/Payments/PaymentManager.swift
designatednerd: also I would recommend breaking this call out into a separate method - especially in terms of looking at crash reports, nested API calls can be a giant pain in the ass to debug

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+                //WARNING: should deal with error
+            }
+        }
+    }
+
+    func sendPayment(completion: @escaping ((_ error: ToshiError?, _ transactionHash: String?) -> Void)) {
+        guard let transaction = transaction else { return }
+        let signedTransaction = "0x\(Cereal.shared.signWithWallet(hex: transaction))"
+
+        print("--- SENDING PAYMENT ---")
```
designatednerd: `DLog` please

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        EthereumAPIClient.shared.sendSignedTransaction(originalTransaction: transaction, transactionSignature: signedTransaction) { success, transactionHash, error in
+
+            print("COMPLETED Payment: SUCCESS: \(success)")
```
designatednerd: `DLog`

Toshi/Controllers/Payments/PaymentManager.swift
```diff
@@ -0,0 +1,83 @@
...
+                })
+            } else {
+                //WARNING: should deal with error
+            }
+        }
+    }
+
+    func sendPayment(completion: @escaping ((_ error: ToshiError?, _ transactionHash: String?) -> Void)) {
+        guard let transaction = transaction else { return }
+        let signedTransaction = "0x\(Cereal.shared.signWithWallet(hex: transaction))"
```
designatednerd: Is this now the only place where we're actually pre-pending the `0x`? Seems like something we might want to have a wrapper for otherwise

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -0,0 +1,130 @@
...
+    private var shouldSendSignedTransaction = true
+
+    private var paymentViewModel: PaymentViewModel
+
+    init(withAddress address: String? = nil, andValue value: NSDecimalNumber? = nil, shouldSendSignedTransaction: Bool = true) {
+        self.paymentViewModel = PaymentViewModel(recipientAddress: address, value: value)
+        self.shouldSendSignedTransaction = shouldSendSignedTransaction
+    }
+
+    // I purposefully created this method so the caller is aware that this object will present a VC
```
designatednerd: Good info but it'd be nice to have a `///` docu-comment that'll show up at the call site.

Toshi/Controllers/Payments/View/PaymentAddressInputView.swift
```diff
@@ -95,7 +94,7 @@ class PaymentAddressInputView: UIView {
 extension PaymentAddressInputView: UITextFieldDelegate {
 
     func textFieldShouldReturn(_ textField: UITextField) -> Bool {
-        delegate?.didRequestSendPayment()
+//        delegate?.didRequestSendPayment()
```
designatednerd: can this be removed now?

Toshi/Controllers/Payments/View/ReceiptView.swift
```diff
@@ -0,0 +1,88 @@
...
+
+        addWithDefaultConstraints(view: fiatAmountLine)
+        addSpacing(.mediumInterItemSpacing, after: fiatAmountLine)
+
+        addWithDefaultConstraints(view: estimatedNetworkFeesLine)
+        addSpacing(.largeInterItemSpacing, after: estimatedNetworkFeesLine)
+
+        let separator = BorderView()
+        addWithDefaultConstraints(view: separator)
+        separator.addHeightConstraint()
```
designatednerd: this can now be updated to use the addDefaultBorder() method I added to `UIStackView`

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -509,11 +511,10 @@ class ProfileViewController: UIViewController {
     }
     
     @objc private func didTapPayButton() {
-        let paymentController = PaymentController(withPaymentType: .send, continueOption: .send)
-        paymentController.delegate = self
-        
-        let navigationController = UINavigationController(rootViewController: paymentController)
-        Navigator.presentModally(navigationController)
+        self.paymentRouter = PaymentRouter(withAddress: profile.paymentAddress)
```
designatednerd: maybe do this as a local let, do all the setup and presenting on that local let, and then store the payment router to the variable after that? avoids a lot of unnecessary optionals.

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -687,79 +688,10 @@ extension ProfileViewController: RateUserControllerDelegate {
...
-                }, failure: { error in
-                    DLog("\(error)")
-                    if error.localizedDescription == "ERROR_DESCRIPTION_UNREGISTERED_RECIPIENT" {
-                        CrashlyticsLogger.nonFatal("Could not send payment because recipient was unregistered", error: (error as NSError), attributes: nil)
-                    } else {
-                        CrashlyticsLogger.log("Can not send message", attributes: [.error: error.localizedDescription])
-                    }
-                })
-            }
-        }
```
designatednerd: a+++ code deletion would delete again

Toshi/Controllers/Recent/ThreadCellConfigurator.swift
```diff
@@ -47,7 +47,7 @@ final class ThreadCellConfigurator: CellConfigurator {
         if thread.isGroupThread() {
             avatar = (thread as? TSGroupThread)?.groupModel.groupImage ?? UIImage(named: "avatar-placeholder")
             title = thread.name()
-        } else if let recipient = ThreadCellConfigurator.recipient(for: thread) {
+        } else if let recipient = thread.recipient() {
```
designatednerd: ^^ this looks like it accidentally got reverted too

Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -3,13 +3,20 @@ import UIKit
 import TinyConstraints
 
 class BalanceController: UIViewController {
+    private let reuseIdentifier = "BalanceControllerCell"
```
designatednerd: this should probably be moved to a `static let` on the cell itself - I think there's some kind of protocol in SweetUIKit that might do this for you, too

Toshi/Controllers/Recent/ThreadCellConfigurator.swift
yuliaveres: no, now it is on thread extension

Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -108,12 +109,9 @@ extension BalanceController: UITableViewDelegate {
...
         if indexPath.row == 1 {
-            
-            let paymentController = PaymentController(withPaymentType: .send, continueOption: .next)
-            paymentController.delegate = self
-            
-            let navigationController = PaymentNavigationController(rootViewController: paymentController)
-            Navigator.presentModally(navigationController)
+            paymentRouter = PaymentRouter()
+            paymentRouter?.delegate = self
+            paymentRouter?.present()
```
designatednerd: same thing about pulling this to a local variable to avoid unnecessary optionality

Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -168,12 +166,9 @@ extension BalanceController: UITableViewDataSource {
...
-    
-    func paymentControllerFinished(with valueInWei: NSDecimalNumber?, for controller: PaymentController) {
-        guard let valueInWei = valueInWei else { return }
-        
-        let paymentAddressController = PaymentAddressController(with: valueInWei)
-        controller.navigationController?.pushViewController(paymentAddressController, animated: true)
+extension BalanceController: PaymentRouterDelegate {
+
+    func paymentRouterDidSucceedPayment(_ paymentRouter: PaymentRouter, parameters: [String: Any], transactionHash: String?, unsignedTransaction: String?, error: ToshiError?) {
+        print("payment succeeded!")
```
designatednerd: `DLog` plz

Toshi/Controllers/TabBarController.swift
```diff
@@ -281,26 +284,16 @@ extension TabBarController: ScannerViewControllerDelegate {
 
     private func proceedToPayment(userInfo: UserInfo, parameters: [String: Any], confirmationText: String) {
 
-        if parameters["value"] != nil, let scannerController = self.scannerController as? ScannerController {
+        self.payedUserInfo = userInfo
+
+        if let hexValue = parameters["value"] as? String, let destinationAddress = parameters["to"] as? String, let scannerController = self.scannerController as? ScannerController {
```
designatednerd: more hard coded instances of the same keys

Toshi/Controllers/TabBarController.swift
```diff
@@ -22,6 +22,9 @@ let TabBarItemTitleOffset: CGFloat = -3.0
 class TabBarController: UITabBarController, OfflineAlertDisplaying {
     let offlineAlertView = defaultOfflineAlertView()
 
+    var paymentRouter: PaymentRouter?
+    var payedUserInfo: UserInfo?
```
designatednerd: English nitpick: `paidUserInfo`

Toshi/Controllers/TabBarController.swift
```diff
@@ -327,6 +320,22 @@ extension TabBarController: ScannerViewControllerDelegate {
...
+    func paymentRouterDidSucceedPayment(_ paymentRouter: PaymentRouter, parameters: [String: Any], transactionHash: String?, unsignedTransaction: String?, error: ToshiError?) {
+        guard error == nil, let userInfo = payedUserInfo else {
+            scannerController.startScanning()
+            return
+        }
+
+        if let scannerController = self.scannerController as? ScannerController {
+            scannerController.approvePayment(with: parameters, userInfo: userInfo, transaction: transactionHash, error: error)
+        } else {
+            scannerController.startScanning()
```
designatednerd: why does this VC need to start scanning again if the payment succeed?

Toshi/Extensions/NSDecimalNumber+Additions.swift
```diff
@@ -65,4 +65,12 @@ extension NSDecimalNumber {
 
         self.init(string: decimal)
     }
+
+    //WARNING: this method needs muchos testing!
+    func isGreaterOrEqualThen(value: NSDecimalNumber) -> Bool {
```
designatednerd: a) Agreed on testing
b) English nit: `isEqualOrGreaterThan`

Toshi/Extensions/NSDecimalNumber+Additions.swift
```diff
@@ -65,4 +65,12 @@ extension NSDecimalNumber {
...
+        let result = compare(value)
+
+        return (result == .orderedAscending) ? false : true
```
designatednerd: it might be clearer to the reader to do this as a `switch` statement

Toshi/Extensions/TSThread+Addition.swift
```diff
@@ -17,6 +17,25 @@ import Foundation
...
+
+        var recipient: TokenUser?
+
+        let retrievedData = contactData(for: recipientAddress)
+
+        if let userData = retrievedData,
+            let deserialised = (try? JSONSerialization.jsonObject(with: userData, options: [])),
+            let json = deserialised as? [String: Any] {
+
+            recipient = TokenUser(json: json, shouldSave: false)
```
designatednerd: this should likely get indented one more time so it's not confused with the additional lets above

Toshi/Extensions/TSThread+Addition.swift
```diff
@@ -17,6 +17,25 @@ import Foundation
 
 extension TSThread {
 
+    func recipient() -> TokenUser? {
```
designatednerd: do we need to throw a note on here about needing to hit this from the main thread since it could wind up hitting the app delegate?

designatednerd: (or just wrap the whole mess in a dispatch to main?)

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -34,6 +34,8 @@ final class SOFAWebController: UIViewController {
     }
 
     private let rcpUrl = ToshiWebviewRPCURLPath
+    var paymentRouter: PaymentRouter?
+    var currentTransactionSignCallbackId: String?
```
designatednerd: can either of these be `private` or at least `private(set)`

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -324,22 +338,18 @@ extension SOFAWebController: WKScriptMessageHandler {
         }
     }
 
-    private func presentPaymentConfirmation(with messageText: String, parameters: [String: Any], userInfo: UserInfo, callbackId: String) {
+    private func presentPaymentConfirmation(with messageText: String, parameters: [String: Any], userInfo: UserInfo, dappInfo: DappInfo, callbackId: String) {
 
-        PaymentConfirmation.shared.present(for: parameters, title: Localized("payment_request_confirmation_warning_title"), message: messageText, approveHandler: { [weak self] transaction, error in
+        guard let destinationAddress = parameters["to"] as? String, let hexValue = parameters["value"] as? String else { return }
```
designatednerd: moar hard-coded string keys

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -324,22 +338,18 @@ extension SOFAWebController: WKScriptMessageHandler {
...
-
-            self?.approvePayment(with: parameters, userInfo: userInfo, transaction: transaction, callbackId: callbackId)
-            }, cancelHandler: { [weak self] in
-                let payload = "{\\\"error\\\": \\\"Transaction declined by user\\\", \\\"result\\\": null}"
-                self?.jsCallback(callbackId: callbackId, payload: payload)
-        })
+        paymentRouter?.userInfo = userInfo
+        paymentRouter?.dappInfo = dappInfo
+        
+        paymentRouter?.present()
```
designatednerd: same thing on avoiding unnecessary optionals by pulling things out into a local var

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -367,6 +377,47 @@ extension SOFAWebController: WKScriptMessageHandler {
...
 }
 
+extension SOFAWebController: PaymentRouterDelegate {
+
+    func paymentRouterDidSucceedPayment(_ paymentRouter: PaymentRouter, parameters: [String: Any], transactionHash: String?, unsignedTransaction: String?, error: ToshiError?) {
+
+        guard let callbackId = currentTransactionSignCallbackId else {
+            let message = "No current signed transcation callBack Id on SOFAWebVontroller when payment router finished"
+            assertionFailure(message)
+            CrashlyticsLogger.log(message, attributes: parameters)
```
designatednerd: Worth doing this as a non-fatal error so we get notified when it's happening? (same applies to the the one below)

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -367,6 +377,47 @@ extension SOFAWebController: WKScriptMessageHandler {
...
+            return
+        }
+
+        guard let userInfo = paymentRouter.userInfo else {
+            let message = "Not found any user info on SOFAWebVontroller payment router after it finished"
```
designatednerd: english nit: `No user info...` or `Did not find any user info...` instead of what's there now.

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -367,6 +377,47 @@ extension SOFAWebController: WKScriptMessageHandler {
...
+            assertionFailure(message)
+            CrashlyticsLogger.log(message, attributes: parameters)
+            return
+        }
+
+        guard error == nil else {
```
designatednerd: a+ use of guard statements here to keep control flow really obvious, btw

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -367,6 +377,47 @@ extension SOFAWebController: WKScriptMessageHandler {
...
+
+            return
+        }
+
+        approvePayment(with: parameters, userInfo: userInfo, transaction: unsignedTransaction, callbackId: callbackId)
+    }
+
+    func paymentRouterDidCancel(paymentRouter: PaymentRouter) {
+        guard let callbackId = currentTransactionSignCallbackId else {
+            let message = "No current signed transcation callBack Id on SOFAWebVontroller when payment router finished"
```
designatednerd: i would probably update this to `cancelled` instead of `finished` so we can distinguish between this log and the one for when the payment router finishes

Toshi/Views/ActionButton.swift
```diff
@@ -231,6 +239,29 @@ class ActionButton: UIControl {
...
+        addSubview(activityIndicator)
+        activityIndicator.centerX(to: self)
+        activityIndicator.centerY(to: self)
+
+        activityIndicator.startAnimating()
+
+        UIView.animate(withDuration: 0.2) {
+            self.titleLabel.alpha = 0
+            self.activityIndicator.alpha = 1
+
```
designatednerd: excess newline

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
yuliaveres: nope, this is new

Tests/IDAPIClientTests.swift
```diff
@@ -136,6 +136,33 @@ class IDAPIClientTests: QuickSpec {
...
 
+                it("finds a user by address") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: IDAPIClientTests.self), mockFilename: "fetchUsersFromPaymentAddresses")
+                    subject = IDAPIClient(teapot: mockTeapot)
+
+                    let paymentAddress = "0x1ad0bb2d14595fa6ad885e53eaaa6c82339f9b98"
+
+                    waitUntil { done in
+                        subject.findUserWithPaymentAddress(paymentAddress) { user, error in
+                            expect(user).toNot(beNil())
```
designatednerd: a) might be better to do a `guard let` on the user and then an `XCTFail` (or I think maybe just `fail()` in QuickSpec) rather than having to use a zillion `?`s below.
b) Can you assert anything about the error parameter here?

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
designatednerd: (yah sorry I hadn't seen your earlier comment yet)

Toshi/Controllers/Payments/PaymentController.swift
yuliaveres: 🤔  prolly Marijn's part of refactor 🙈

Toshi/Controllers/Payments/PaymentManager.swift
yuliaveres: as for me - while I do not need any functionality with that data, i would go for tuple

designatednerd: out of curiosity, why do you prefer a tuple? 

my instinct is to `Struct` since it can give you some options for easier mainipulation

Toshi/Controllers/TabBarController.swift
yuliaveres: it should get ready to scan any new item, as after sending a payment nothing else happens ..

Toshi/Controllers/TabBarController.swift
yuliaveres: all of those were addressed 👍

Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -689,9 +689,19 @@ extension ChatViewController: UITableViewDataSource {
...
     }
 
-    private func presentPaymentRouter(address: String?, value: NSDecimalNumber? = nil) {
+    private func presentPaymentRouter(address: String, value: NSDecimalNumber? = nil) {
 
-        self.paymentRouter = PaymentRouter(withAddress: address, andValue: value)
+        var parameters = ["to": address]
+
+        if let value = value {
+            parameters["value"] = value.toHexString
```
designatednerd: still hard-coded (also at line 694)

Toshi/Controllers/TabBarController.swift
```diff
@@ -263,7 +263,7 @@ extension TabBarController: ScannerViewControllerDelegate {
 
     private func proceedToPayment(address: String, weiValue: String?, confirmationText: String) {
         let userInfo = UserInfo(address: address, paymentAddress: address, avatarPath: nil, name: nil, username: address, isLocal: false)
-        var parameters = ["from": Cereal.shared.paymentAddress, "to": address]
+        var parameters = [PaymentParameters.from: Cereal.shared.paymentAddress, PaymentParameters.to: address]
         parameters["value"] = weiValue
```
designatednerd: still hard-coded

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -0,0 +1,478 @@
...
+
+            return
+        }
+
+        paymentManager.sendPayment { [weak self] error, transactionHash in
+            guard let weakSelf = self else { return }
+
+            guard error == nil else {
+                let alert = UIAlertController(title: Localized("transaction_error_message"), message: (error?.description ?? ToshiError.genericError.description), preferredStyle: .alert)
+                alert.addAction(UIAlertAction(title: "Ok", style: .default, handler: { _ in
```
designatednerd: 🚨 non-localized user-facing string 🚨

Toshi/Controllers/Chat/ChatInteractor.swift
```diff
@@ -113,38 +113,18 @@ final class ChatInteractor: NSObject {
...
-            let fiatValueString = EthereumConverter.fiatValueString(forWei: value, exchangeRate: ExchangeRateClient.exchangeRate)
-            let ethValueString = EthereumConverter.ethereumValueString(forWei: value)
-            let messageText = String(format: Localized("payment_confirmation_warning_message"), fiatValueString, ethValueString, user.name)
-
-            PaymentConfirmation.shared.present(for: parameters, title: Localized("payment_request_confirmation_warning_title"), message: messageText, approveHandler: { [weak self] transaction, _ in
-
-                self?.output?.didFinishRequest()
-
-                guard let transaction = transaction else { return }
+    // need to write a test for this
```
designatednerd: is this comment still valid since there's now an API test for this?

Tests/EthereumConverterTests.swift
marijnschilling: This test was to test for a valid string for the old paymentConfirmation. Since we don't need that string anymore I removed the method and the test. 

`The estimated Ethereum network fees are 0,06 EUR (o,000001 ETH)` (that was the string)
Now we just show this cool receipt ✌️

Toshi/Controllers/Payments/View/ReceiptView.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    private lazy var ethereumAmountLabel: UILabel = {
+        let view = UILabel()
+        view.numberOfLines = 0
+        view.font = Theme.preferredFootnote()
+        view.textColor = Theme.lightGreyTextColor
+        view.textAlignment = .right
+        view.adjustsFontForContentSizeCategory = true
+
+        view.text = "Just testing 193819283 ETH"
```
designatednerd: this should probably be removed before we merge

Toshi/Extensions/TSThread+Addition.swift
```diff
@@ -17,6 +17,26 @@ import Foundation
 
 extension TSThread {
 
+    // Needs to be called on main thread as it involves UIAppDelegate
```
designatednerd: I'd add a third slash to this so it shows up at the call site.

Toshi/Controllers/Recent/ThreadCellConfigurator.swift
designatednerd: [nevermind]

Toshi/Controllers/TabBarController.swift
```diff
@@ -327,6 +321,22 @@ extension TabBarController: ScannerViewControllerDelegate {
...
+extension TabBarController: PaymentRouterDelegate {
+
+    func paymentRouterDidSucceedPayment(_ paymentRouter: PaymentRouter, parameters: [String: Any], transactionHash: String?, unsignedTransaction: String?, error: ToshiError?) {
+        guard error == nil, let userInfo = paidUserInfo else {
+            scannerController.startScanning()
+            return
+        }
+
+        if let scannerController = self.scannerController as? ScannerController {
+            scannerController.approvePayment(with: parameters, userInfo: userInfo, transaction: transactionHash, error: error)
```
marijnschilling: What is this check for the scannerController being a ScannerController here? And why does it need to approve a payment after the PaymentRouter did just succeed the payment?

marijnschilling: I think it;s just old code, I removed it in my PR. But correct me if i'm wrong!

yuliaveres: i think so @marijnschilling , please double check

Toshi/Extensions/NSDecimalNumber+Additions.swift
marijnschilling: Great! I always do that wrong 🙈

Toshi/Controllers/Settings/Balance/BalanceController.swift
marijnschilling: That way the delegate isnt there when i want to call it :(

Toshi/Controllers/Payments/PaymentController.swift
```diff
@@ -197,8 +200,9 @@ extension PaymentController: UITextFieldDelegate {
         guard let newValue = (textField.text as NSString?)?.replacingCharacters(in: range, with: string) else {
             return true
         }
-        
-        navigationItem.rightBarButtonItem?.isEnabled = !newValue.isEmpty
+
+        let newIntValue = Int(newValue) ?? 0
```
marijnschilling: This does not work when the value get's entered with a `,` then the value will be `0,10` which wont be parsed as an int. 

We should make a method added to String to check wether it's bigger than 0, with added tests. @yuliaveres

marijnschilling: Working on a tiny PR for this now ✌️

marijnschilling:
I added alerts!
screenshot 2018-01-24 17 34 12
screenshot 2018-01-24 17 34 21
screenshot 2018-01-24 17 34 28

PR #695, opened 2018-01-18
Help out with updated Payments flow

yuliaveres:


yuliaveres:
Merging this into initial branch and making PR from there.

PR #696, opened 2018-01-19
Payment view for dapps

designatednerd:
In this PR: 
- Dropping a couple things down to pure stack views for simpler use
- Setup of Dapp style of payment confirmation view
- Some shuffling of deck chairs in the payment confirmation view to make setup of the dapp version easier
- Convenience method for adding a border view to a stack view


Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -67,11 +70,18 @@ class PaymentRouter {
     }
 
     private func presentPaymentConfirmationController(withValue value: NSDecimalNumber, andRecipientAddress address: String) {
-        let paymentConfirmationController = PaymentConfirmationViewController(withValue: value, andRecipientAddress: address, userInfo: userInfo)
-        paymentConfirmationController.delegate = self
+        if let dappInfo = dappInfo {
+            let paymentConfirmationController = PaymentConfirmationViewController(withValue: value, andRecipientAddress: address, recipientType: .dapp(info: dappInfo))
+            paymentConfirmationController.delegate = self
```
yuliaveres: i guess we have to adjust `modalPresentationStyle` ? so the top part is transparent ?

designatednerd: yeah right now this still needs some refining and direction - I could probably set it up so it's presented instantly with a screenshot background of whatever it's coming in over, and then slide up the payment thing. 

But that can most certainly wait until monday 😛

PR #697, opened 2018-01-22
[IOS-461]: Do not open thread when db and session is not setup completely

yuliaveres:


yuliaveres:
i believe this is the only place of accessing DB before it was setup :(  .

PR #698, opened 2018-01-22
Improve Payment Modal Animation

designatednerd:
In this PR: 
- Distinguish presentation style separately from content
- Now modal style looks more like the Apple pay share sheet: 

modal

- Fix an issue where the size of the receipt view would jump when the ethereum exchange rate loaded.
- Fix swiftLint warnings

### DIFF
--- a/Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
+++ b/Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
@@ -26,6 +26,12 @@ enum RecipientType {
     dapp(info: DappInfo)
 }
 
+enum PresentationMethod {
+    case
+    pushedInNav,
+    modalBottomSheet
+}
+
 class PaymentConfirmationViewController: UIViewController {
 
     weak var delegate: PaymentConfirmationViewControllerDelegate?
@@ -110,6 +116,36 @@ class PaymentConfirmationViewController: UIViewController {
     }()
 
     // MARK: Payment sheet style title
+
+    private var hasLaidOutBefore = false
+    private var bottomOfReceiptConstraint: NSLayoutConstraint?
+
+    var backgroundView: UIView? {
+        didSet {
+            guard let background = backgroundView else { return }
+
+            let alpha = UIView()
+            alpha.backgroundColor = .black
+            alpha.alpha = 0.5
+
+            background.addSubview(alpha)
+            alpha.edgesToSuperview()
+
+            view.insertSubview(background, at: 0)
+        }
+    }
+
+    var presentationMethod: PresentationMethod = .pushedInNav {
+        didSet {
+            switch presentationMethod {
+            case .pushedInNav:
+                // Not much to do here
+                break
+            case .modalBottomSheet:
+                self.modalTransitionStyle = .crossDissolve
+            }
+        }
+    }
 
     private lazy var paymentSheetTitleLabel: UILabel = {
         let view = UILabel()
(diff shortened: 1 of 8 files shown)


Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -111,6 +117,36 @@ class PaymentConfirmationViewController: UIViewController {
...
+    }
+
+    var presentationMethod: PresentationMethod = .fullScreen {
+        didSet {
+            switch presentationMethod {
+            case .fullScreen:
+                // Not much to do here
+                break
+            case .modalBottomSheet:
+                self.modalTransitionStyle = .crossDissolve
```
yuliaveres: crossDissolve ? 🤔  it does not look like . any reason why it is crossDissolve ?

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -228,6 +264,34 @@ class PaymentConfirmationViewController: UIViewController {
...
+
+        switch presentationMethod {
+        case .fullScreen:
+            // do nothing
+            return
+        case .modalBottomSheet:
+            guard !hasLaidOutBefore, bottomOfReceiptConstraint != nil else { return }
+
+            hasLaidOutBefore = true
+            setReceiptShowing(false, animated: false)
```
yuliaveres: could you please explain why we do in this way? if that is some workaround - maybe lets add a comment ?

yuliaveres: i mean `setReceiptShowing(false)` and `setReceiptShowing(true)` ?

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -367,15 +439,25 @@ class PaymentConfirmationViewController: UIViewController {
...
-        background.bottomToTop(of: viewToPinToTopOf)
-    }
+    private func setReceiptShowing(_ showing: Bool, animated: Bool = true, completion: (() -> Void)? = nil) {
+        guard let bottomConstraint = bottomOfReceiptConstraint else { /* nothing to adjust yet */ return }
+
+        let targetConstraintConstant: CGFloat = showing ? .largeInterItemSpacing : view.frame.height
+
+        guard bottomOfReceiptConstraint?.constant != targetConstraintConstant else { /* already where we want it to be */ return }
+
+        bottomConstraint.constant = targetConstraintConstant
```
yuliaveres: it is always recommended to call `layoutIfNeeded()` before actual animation block so any pending layout changes are not included in your animation.

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -441,6 +523,17 @@ class PaymentConfirmationViewController: UIViewController {
     // MARK: - Action Targets
 
     @objc func cancelItemTapped() {
+        switch presentationMethod {
+        case .fullScreen:
+            actuallyDismiss()
+        case .modalBottomSheet:
+            self.setReceiptShowing(false) {
+                self.actuallyDismiss()
```
yuliaveres: do we want to use `weakSelf` in the block ?

Toshi/Extensions/UIStackView+Additions.swift
```diff
@@ -70,13 +70,22 @@ extension UIStackView {
...
+    ///   - margin: [Optional] The margin to add around the view. Useful when there's a margin where the view is pinned which you want to make sure has the appropriate background color. When nil, edges will simply be pinned to the stack view itself.
+    func addBackground(with color: UIColor, margin: CGFloat? = nil) {
         let background = UIView()
         background.backgroundColor = color
         
         self.addSubview(background)
-        background.edgesToSuperview()
+        if let margin = margin {
+            background.topToSuperview(offset: -margin)
+            background.leftToSuperview(offset: -margin)
```
yuliaveres: do we want `-value` here ?

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
designatednerd: I'll throw in a comment. Basically, it shoves the receipt off the screen in a non-animated fashion so it can be animated in

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
designatednerd: So `crossDissolve` fades in the background and then the VC itself animates showing of the bottom modal. same deal on dismiss - when in modal mode, the completion block actually dismisses the view controller (it just looks the same because the background view is a screenshot of the window)

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
designatednerd: No, since this is basically an animation block.

Toshi/Extensions/UIStackView+Additions.swift
designatednerd: yes - this way the view will get pinned so that the background view is larger than the stack view itself, and the background will cover the margin between the edge of the screen and the edge of the stack view.

yuliaveres:
still good 👍

PR #700, opened 2018-01-23
[IOS-466]: Handle invalid certificate for secure http requests

yuliaveres:
The HTTPS request for some urls failed to load due to invalid certificate. We need to handle this and use url credential from server trust.


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -164,6 +164,20 @@ extension SOFAWebController: WKNavigationDelegate {
...
+        let protectionSpace = challenge.protectionSpace
+
+        guard protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust,
+            let trust = challenge.protectionSpace.serverTrust else {
+                completionHandler(URLSession.AuthChallengeDisposition.performDefaultHandling, nil)
+                return
+        }
+
+        let urlCredentials = URLCredential(trust: trust)
+        completionHandler(URLSession.AuthChallengeDisposition.useCredential, urlCredentials)
```
designatednerd: Can you throw some comments in here explaining what's happening for people who aren't familiar with what's happening here? 

For example, I'm not entirely clear on why we'd be getting an authentication challenge for https on something where we don't actually need to provide auth credentials for a user, or why sending it the server trust credential solves this problem. 

If there's some kind of document or webpage you can link to that explains that stuff, that'd also work.

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -164,6 +164,22 @@ extension SOFAWebController: WKNavigationDelegate {
         backBarButtonItem.isEnabled = webView.canGoBack
         forwardBarButtonItem.isEnabled = webView.canGoForward
     }
+
+    func webView(_ webView: WKWebView, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Swift.Void) {
+
+        let protectionSpace = challenge.protectionSpace
+
+        guard protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust,
+            let trust = challenge.protectionSpace.serverTrust else {
```
designatednerd: you can probably reduce this to just `protectionSpace` given the `let` you've pulled out above

Toshi/SOFAWebview/SOFAWebController.swift
designatednerd: [addressed]

PR #701, opened 2018-01-23
Finish payment router flow

marijnschilling:
- implemented feedback on main Payment Router flow PR
- Added test for 'greaterThanOrEqual` and 'greaterThan` `NSDecimalNumberOperators
- Renamed the classes related to Payment flow to be more consistent
- made all the things private and final
- Added the check for valid payment address in the address controller. (next button only becomes enabled with valid paymentaddress)
- Removed all the classes related to the old paymentConfirmation (alert thing)
- Show the spinner when the payment is in the making


Toshi/Controllers/Chat/ChatViewController.swift
```diff
@@ -879,7 +881,7 @@ extension ChatViewController: ChatFloatingHeaderViewDelegate {
 
     func messagesFloatingView(_: ChatFloatingHeaderView, didPressRequestButton _: UIButton) {
         
-        let paymentController = PaymentController(withPaymentType: .request, continueOption: .send)
+        let paymentController = PaymentValueViewController(withPaymentType: .request, continueOption: .send)
```
designatednerd: worth also updating the name of this var?

Toshi/Controllers/Payments/PaymentAddressViewController.swift
```diff
@@ -4,13 +4,19 @@ import TinyConstraints
...
 }
 
-class PaymentAddressController: UIViewController {
+class PaymentAddressViewController: UIViewController {
 
     private let valueInWei: NSDecimalNumber
 
+    private var paymentAddress: String? {
+        didSet {
+          validate(paymentAddress)
```
designatednerd: is there anything you want to do if the address doesn't validate?

Toshi/Controllers/Payments/PaymentAddressViewController.swift
```diff
@@ -45,18 +51,32 @@ class PaymentAddressController: UIViewController {
...
-    lazy var scannerController: ScannerViewController = {
+    private lazy var scannerController: ScannerViewController = {
         let controller = ScannerController(instructions: Localized("payment_qr_scanner_instructions"), types: [.qrCode])
         controller.delegate = self
 
         return controller
     }()
 
-    private lazy var sendBarButton = UIBarButtonItem(title: Localized("payment_send_button"), style: .plain, target: self, action: #selector(nextBarButtonTapped(_:)))
+    @discardableResult private func validate(_ address: String?) -> Bool {
```
designatednerd: do you mind putting `discardableResult` on the line above the function signature? I find it helps keep the primary functionality of the method clearer

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -378,7 +377,7 @@ class PaymentConfirmationViewController: UIViewController {
         profileDetailsStackView.addWithDefaultConstraints(view: userNameLabel)
     }
 
-    func addDappStackViewLayout(to parentView: UIView, above viewToPinToTopOf: UIView) {
+    private func addDappStackViewLayout(to parentView: UIView, above viewToPinToTopOf: UIView) {
```
designatednerd: oopsies. good catch 👍

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -444,6 +443,10 @@ class PaymentConfirmationViewController: UIViewController {
     }
 
     // MARK: - Animation
+    private func addBackgroundView(to parentView: UIView, above viewToPinToTopOf: UIView) {
```
designatednerd: is this used?

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -120,14 +120,18 @@ extension PaymentRouter: PaymentConfirmationViewControllerDelegate {
...
 
-        if Navigator.topViewController is PaymentNavigationController {
-            Navigator.rootViewController?.dismiss(animated: true)
-        }
+        guard let tabBarController = Navigator.tabbarController else { return }
+        guard let selectedNavigationController = tabBarController.selectedViewController as? UINavigationController else { return }
 
-        self.delegate?.paymentRouterDidSucceedPayment(self, parameters: parameters, transactionHash: transactionHash, unsignedTransaction: controller.originalUnsignedTransaction, error: error)
-    }
+        guard let firstPaymentPresentedController = selectedNavigationController.presentedViewController else { return }
```
designatednerd: worth combining these three `guard let`s using commas?

Toshi/Controllers/Payments/PaymentValueViewController.swift
```diff
@@ -16,7 +16,7 @@
 import UIKit
 
 protocol PaymentControllerDelegate: class {
```
designatednerd: worth updating the name of the payment type and this delegate to match the updated VC name?

Toshi/Controllers/Payments/View/PaymentAddressInputView.swift
```diff
@@ -60,6 +62,8 @@ class PaymentAddressInputView: UIView {
     override init(frame: CGRect) {
         super.init(frame: frame)
 
+        NotificationCenter.default.addObserver(self, selector: #selector(didChangeAddress), name: Notification.Name.UITextFieldTextDidChange, object: nil)
```
designatednerd: why do this instead of making this class a `UITextFieldDelegate`?

Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -3,9 +3,11 @@ import UIKit
 import TinyConstraints
 
 class BalanceController: UIViewController {
-    private let reuseIdentifier = "BalanceControllerCell"
-
     private var paymentRouter: PaymentRouter?
+    
+    enum BalanceItem: Int {
+        case balance, send, deposit
```
designatednerd: do you mind putting each of these on a new line for readability? You only need the one case statement though

Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -107,16 +109,19 @@ class BalanceController: UIViewController {
...
+        case .send:
+            let paymentRouter = PaymentRouter()
+            paymentRouter.present()
+
+            self.paymentRouter = paymentRouter
+        case .deposit:
             guard let current = TokenUser.current else { return }
             let controller = DepositMoneyController(for: current.displayUsername, name: current.name)
             self.navigationController?.pushViewController(controller, animated: true)
+        default: break
```
designatednerd: do you mind putting the `break` down one line so it's easier to parse visually that it's there?

Toshi/Controllers/Payments/PaymentAddressViewController.swift
marijnschilling: yeah, we disable the next button when it's not valid ✌️

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
marijnschilling: Hmm apparantly not! It's sneaked in with a merge conflict.

Toshi/Controllers/Payments/View/PaymentAddressInputView.swift
marijnschilling: Well what delegate method do you need to overwrite then? I tried that first but there's no delegate method that is called after the value changed, so i needed this notification instead.

marijnschilling: There's only did/should end/begin editing and `shouldChangeCharactersInRange` and `textFieldShouldClear` and return :( 

I just want textFieldDidChange 😭

yuliaveres: then `shouldChangeCharactersInRange ` is something you need

marijnschilling: But then I need to create the address string by doing difficult logic with the replacement string and the current textField.text right?

Toshi/Controllers/Payments/PaymentAddressViewController.swift
designatednerd: I missed that line in the method, and I wouldn't expect that behavior from the name or return type. Maybe `enableButtonIfValid(paymentAddress:)`?

Toshi/Controllers/Payments/View/PaymentAddressInputView.swift
```diff
@@ -94,7 +96,15 @@ class PaymentAddressInputView: UIView {
...
 
     func textFieldShouldReturn(_ textField: UITextField) -> Bool {
-//        delegate?.didRequestSendPayment()
+        delegate?.didRequestSendPayment()
         return false
     }
+
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        let textFieldText: NSString = (textField.text ?? "") as NSString
+        let textAfterUpdate = textFieldText.replacingCharacters(in: range, with: string)
```
designatednerd: Probably worth doing in a separate pull but [some stuff Igor just added to SweetFoundation](https://github.com/UseSweet/SweetFoundation/pull/31) should make this a lot simpler.

Toshi/Controllers/ScannerController.swift
```diff
@@ -54,29 +54,9 @@ class ScannerController: ScannerViewController {
...
-                self.presentPaymentError(withErrorMessage: error?.localizedDescription ?? ToshiError.genericError.description)
-                self.startScanning()
-
-                return
-        }
-
-        let signedTransaction = self.createSignedTransaction(from: transaction)
-
-        sendTransaction(originalTransaction: transaction, signedTransaction: signedTransaction)
+    func setStatusBarHidden(_: Bool) {
```
designatednerd: does this need a parameter if you're never using it?

Toshi/Controllers/Payments/PaymentAddressViewController.swift
marijnschilling: yup I just removed the button enabling from this function, it was a weird side effect

Toshi/Controllers/Payments/PaymentAddressViewController.swift
```diff
@@ -4,13 +4,20 @@ import TinyConstraints
...
 
-class PaymentAddressController: UIViewController {
+class PaymentAddressViewController: UIViewController {
 
     private let valueInWei: NSDecimalNumber
 
+    private var paymentAddress: String? {
+        didSet {
+          let isValid = validate(paymentAddress)
+            navigationItem.rightBarButtonItem?.isEnabled = isValid
```
designatednerd: alignment seems off here

Toshi/Controllers/Payments/PaymentAddressViewController.swift
```diff
@@ -148,15 +169,29 @@ extension PaymentAddressController: ScannerViewControllerDelegate {
...
-extension PaymentAddressController: PaymentAddressInputDelegate {
+extension PaymentAddressViewController: PaymentAddressInputDelegate {
 
     func didRequestScanner() {
         Navigator.presentModally(scannerController)
     }
+
+    func didRequestSendPayment() {
+        let isValid = validate(paymentAddress)
+        navigationItem.rightBarButtonItem?.isEnabled = isValid
```
designatednerd: Hm. I didn't have a problem with it *having* the side effect, since it centrlaized this "if not valid then disable button" logic - I had a problem with it not making the side-effect obvious. I'm fine with leaving this for now but it's definitely something we may want to take anotehr look at in terms of what the best approach is if we need to do more validation in the future.

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -375,6 +393,12 @@ extension SOFAWebController: WKScriptMessageHandler {
     }
 }
 
+extension SOFAWebController: ActivityIndicating {
+    var activityIndicator: UIActivityIndicatorView {
+        return activityView
```
designatednerd: worth just naming the var `activityIndicator` so you don't have to override this?

marijnschilling: It's a protocol, i think it's not possible.

designatednerd: If you just change the name of `activityView` to `activityIndicator` and remove this getter, I think that should satisfy conformance. You can't declare it here, you have to declare it in the main body of the class. But it should at least satisfy conformance.

designatednerd:
@marijnschilling looks like there's still something that needs to be updated in the tests to `Than`

marijnschilling:
yeah im in a terrible 🌀 of working on two branches at the same time that have to keep up to date with eachother 🙈  I think its merge conflictus

marijnschilling:
Thats why i want to merge this branch asap 🙏

Toshi/SOFAWebview/SOFAWebController.swift
marijnschilling: I don't think that works, it's implemented like this everywhere

yuliaveres: i think we should separate protocol improvements , if we need any, to separate PR, as it is same in a lot of places

yuliaveres: the thing also here is that activity view is supposed to be lazy, as it is `defaultActivityIndicator` on default protocol implementation which returns a new view each time

designatednerd: OK, fair enough

PR #702, opened 2018-01-23
Fix check for payment value

marijnschilling:
Improved the way we check wether the value form the payment controller is valid.


Toshi/Extensions/String+Additions.swift
```diff
@@ -64,4 +64,14 @@ extension String {
             return self
         }
     }
+
+    func isValidValue() -> Bool {
```
yuliaveres: it is not clear what kind of value we deal with here. payment value ? .. 🤔

marijnschilling: Yeah i was questioning what to call this! I thought payment value might be a bit to narrow since it could be used for others money values. But i will change the name!

designatednerd:
Change the method name and it's 👍  by me

PR #703, opened 2018-01-23
Noted need for Carthage and link to installation

wanderingstan:


README.md
```diff
@@ -2,6 +2,7 @@
 
 ## Running the project
 
+- Requires installation of [Carthage dependency manager](https://github.com/Carthage/Carthage#installing-carthage).
```
designatednerd: Hey Stan - thanks for the PR! It should only require Carthage if you're building the Distribution target. In your experience, did you need to use it for other targets?

wanderingstan:
I would have *sworn* I was on the `Development` target, but I just tried to repro and you're absolutely right: It only complains on `Distribution`. Sorry! Closing.

designatednerd:
No worries, glad you were able to figure it out. It might be helpful to document this for other developers interested in poking at the code - would you consider updating and reopening your PR?

PR #705, opened 2018-01-24
[IOS-329]: Rewrite AppDelegate into swift; remove all session managers into separate place

yuliaveres:
Got rid of some objc code;
now storing all session related managers in shared instance of SessionManager;
I still would like to think on thread safety of those, but this might be in separate pr.


Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+            assertionFailure("Can't access bundle identifier")
+            CrashlyticsLogger.log(message)
+            return nil
+        }
+
+        var path: String?
+        let groupName = "group.\(bundleIdentifier)"
+
+        if let groupUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: groupName) {
+            let documentsPath = "groupUrl.pat/Documents"
```
designatednerd: did you mean to include `groupUrl.pat` in the string directly?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+        didSet {
+            updateRemoteNotificationsCredentials()
+        }
+    }
+
+    var isFirstLaunch: Bool {
+        return !UserDefaultsWrapper.launchedBefore
+    }
+
+    lazy var documentsPath: String? = {
```
designatednerd: Could this be a lazy var on something else so that it doesn't need to be accessed on the main thread?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+        if shouldProceedToDBSetup {
+            if tryToOpenDB() {
+                configureForCurrentSession()
+            } else {
+                // There might be a case when filesystem state is weird and it doesn't return true results, saying file is not present even if it is.
+                // to determine this we might check keychain for database password being there
+                // in this case we want to wait a bit and try to open file again
+                // if it still fails - both password and database file, whatever is present in Yap, is deleted and splash is presented
+
+                DispatchQueue.main.asyncAfter(seconds: 3, execute: {
```
designatednerd: Any recollection in terms of why this was 3 seconds? I see it's the same in the objc code so I'm mostly curious

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+
+        UIApplication.shared.applicationIconBadgeNumber = 0
+
+        if !Yap.isUserDatabaseFileAccessible && !Yap.isUserDatabasePasswordAccessible {
+            configureAndPresentWindow()
+            return true
+        }
+
+        let shouldProceedToDBSetup = UIApplication.shared.applicationState != .background
+        if shouldProceedToDBSetup {
```
designatednerd: Might be worth doing this as a guard since there's nothing else to do here if this `if` is false.

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+        }
+
+        return true
+    }
+
+    func applicationWillResignActive(_ application: UIApplication) {
+
+        guard TSAccountManager.isRegistered() else { return }
+
+        DispatchQueue.main.async {
```
designatednerd: why dispatch this to main? This should theoretically only be called on the main thread.

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+    func application(_ app: UIApplication, open url: URL, options: [UIApplicationOpenURLOptionsKey: Any] = [:]) -> Bool {
+
+        guard url.scheme == "toshi" else { return false }
+
+        Navigator.tabbarController?.openDeepLinkURL(url)
+
+        return true
+    }
+
+    func createNewUser() {
```
designatednerd: worth pulling out the bits of this that actually create the new user into some kind of `UserManager` instead of tying it to the app delegate?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+        if storageManager.database == nil {
+            CrashlyticsLogger.log("Failed to create chat databse for the user")
+        }
+
+        SessionManager.shared.setupSecureEnvironment()
+        UserDefaultsWrapper.launchedBefore = true
+    }
+
+    private func setupSignalService() {
+        // Encryption/Descryption mutates session state and must be synchronized on a serial queue.
```
designatednerd: nit: `Descryption`-> `Decryption`

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+
+        UIView.animate(withDuration: 0.3, animations: {
+            self.screenProtectionWindow?.alpha = 0
+        }) { _ in
+            self.screenProtectionWindow?.isHidden = true
+        }
+    }
+}
+
+extension AppDelegate: UNUserNotificationCenterDelegate {
```
designatednerd: might also be worth pulling this and the other notification methods out into a notification manager, but that could be a separate pull

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,416 @@
...
+
+        completionHandler()
+    }
+
+    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
+        token = deviceToken.hexadecimalString
+    }
+
+    func application(_ application: UIApplication, didFailToRegisterForRemoteNotificationsWithError error: Error) {
+        OCDLog.alog("Failed to register for remote notifications. \(error)",
```
designatednerd: This can just use `ALog()` directly now that it's in Swift.

Toshi/Supporting Files/main.m
```diff
@@ -1,30 +0,0 @@
...
-//
-// You should have received a copy of the GNU General Public License
-// along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-
-#import <UIKit/UIKit.h>
-#import "AppDelegate.h"
-#import "TestAppDelegate.h"
-
-int main(int argc, char* argv[]) {
```
designatednerd: can you update to use a Swift main so we still get the appropriate use of a different app delegate in tests? There's an example of how to do this in my [open Testing pr](https://github.com/toshiapp/toshi-ios-client/pull/677/files#diff-7fb1acc41fddbe2263d9831ef6f293efR16), though it's using a couple convenience methods that haven't been merged yet as well.

designatednerd:
Also I forgot to say YAAAAAAAY THIS MAKES ME VERY HAPPY

Toshi/Supporting Files/AppDelegate.swift
yuliaveres: there was no particular reason for 3. it might be 2, might be 1. but i decided to give it 3 seconds if this kind of race condition happened.

yuliaveres:
@designatednerd ready with feedback

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,322 @@
...
+    private func setupSignalService() {
+        // Encryption/Decryption mutates session state and must be synchronized on a serial queue.
+
+        SessionCipher.setSessionCipherDispatchQueue(OWSDispatch.sessionStoreQueue())
+        CrashlyticsClient.setupForUser(with: Cereal.shared.address)
+
+        TSSocketManager.requestSocketOpen()
+        RTCInitializeSSL()
+
+        DispatchQueue.main.asyncAfter(seconds: 3) {
```
heumn: Minor nit: since this sounded like it stems from a race condition, maybe document that for "the next guy" so he understands why there is a delay here? 🤔

Toshi/Supporting Files/SessionManager.swift
```diff
@@ -0,0 +1,130 @@
...
+            UserDefaultsWrapper.requiresSignIn = false
+
+            TokenUser.createCurrentUser(with: user.dict)
+            idClient.migrateCurrentUserIfNeeded()
+
+            TokenUser.current?.updateVerificationState(true)
+
+            ChatAPIClient.shared.registerUser()
+
+            guard let delegate = UIApplication.shared.delegate as? AppDelegate else { return }
```
heumn: Nit: would it not be a bit weird to have a `return` when this has a completion block? 🤔  I somehow see the reason for it, but should a caller care if the delegate exist?

yuliaveres: i think i can avoid guard at all 🤦‍♂️

Toshi/Supporting Files/AppDelegate.swift
heumn: > so he understands

> so he or she understands

:)

Toshi/Extensions/UIApplication+Testing.swift
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2017 Token Browser, Inc
```
designatednerd: can you verify after you pull in master that everything's back to 2018?

Toshi/Extensions/UIApplication+Testing.swift
```diff
@@ -0,0 +1,34 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import UIKit
```
designatednerd: this is probably my fault since it came from my PR but can you please add a space between the license comment and the import?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,322 @@
...
+
+    func applicationWillResignActive(_ application: UIApplication) {
+
+        guard TSAccountManager.isRegistered() else { return }
+        self.activateScreenProtection()
+    }
+
+    func applicationDidBecomeActive(_ application: UIApplication) {
+
+        UIApplication.shared.applicationIconBadgeNumber = 0
```
designatednerd: is this necessary in both `didBecomeActive` and `didFinishLaunching`?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -0,0 +1,322 @@
...
+        setupSignalService()
+
+        NotificationCenter.default.post(name: .ChatDatabaseCreated, object: nil)
+    }
+
+    private func setupTSKitEnvironment() {
+        OCDLog.alog("Setting up Signal KIT environment",
+                    filePath: #file,
+                    function: #function,
+                    line: #line)
```
designatednerd: you should be able to just call `ALog("Setting up Signal KIT environment")` - the other parameters will be passed in by default.

Toshi/Supporting Files/AppDelegate.swift
designatednerd: Also since this is the second place we're using 3 as a quasi-arbitrary number to avoid race conditions, maybe worth centralizing it with a name?

yuliaveres: When I think about this delay, I can't think of any reason why it was needed actually.
The only case i suspect was wrong signal prepareation flow which caused storage to be not setup completely while  registering for PNs with a token (the authToken and userAddress is retrieved for request from storage). What is not a case now. So I am not sure if it should stay anymore. Also, to me, it always feels weird seeing the alert with such a delay. 🤔  wdyt @heumn  ?

notif_register

Toshi/Supporting Files/AppDelegate.swift
designatednerd: (should be a separate pull)

Toshi/Supporting Files/PrekeyHandler.swift
```diff
@@ -0,0 +1,67 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+import Foundation
```
designatednerd: blank line above the import plz

Toshi/Supporting Files/main.m
designatednerd: [addressed]

Toshi/Supporting Files/AppDelegate.swift
yuliaveres: 😮  leftover 🙈

Toshi/Supporting Files/AppDelegate.swift
heumn: I am not familiar enough with that part of the flow so would be difficult to evaluate if it would break anything related to the notifications. If not it sounds like you have found a reason to remove it :)

PR #706, opened 2018-01-24
In the end, make navigation buttons on web view work

yuliaveres:
Also, get rid of useless share button.


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -160,10 +155,6 @@ final class SOFAWebController: UIViewController {
 }
 
 extension SOFAWebController: WKNavigationDelegate {
-    func webView(_ webView: WKWebView, didFinish _: WKNavigation!) {
```
marijnschilling: Aah was these methods messing it up?

yuliaveres: it is delegate which is called only once - when webview is loaded. the buttons were set to`disabled` here and were never enabled anymore.

marijnschilling: clear ✌️

yuliaveres: actually there are never reset, as we do not receive any callback on each navigation. so it is better to have them enabled (which is default) and let webview decide if it can go back/forward or no. the same way as it decides while we swipe left / ride

PR #709, opened 2018-01-29
[IOS-380]: Fix sofa buttons being shown even if last message is outgoing

yuliaveres:
Do not show sofa buttons panel if the user just have sent a message. Show only when there is a response and it is the last message.


heumn:
A minor thing I came to think about when I saw this, could a message appear from the receiver (as in an incoming message) and thus the buttons would still be `nil`? 🤔

yuliaveres:
@marijnschilling  try messaging spambot with Network link conditioner 3G or worse. you will notice buttons are not shown if the last message is from you :-)
(it is tricky to notice if he responds quickly 😉  )

yuliaveres:
@heumn  if the message is from the receiver - it is a message of `TSIncomingMessage` type (that is how the `isOutgoing` value is determined.) in that case the buttons would be nil only if none of previous messages has some. does it answer what you were wondering about ? let me know if I misunderstood the question 😄

heumn:
Yeah got and and double checked the flow. Looks good!

PR #713, opened 2018-01-30
Cancel payment router only when its controllers stack is empty

yuliaveres:
F.e. - if there is payment value controller and payment confirmation for specific payment scenario, and after confirmation is cancelled , router fires cancel callback on its delegate (but payment value input is still there). So after payment value got cancelled -> the state on sofa web controller is broken, callback id is nil what is not expected for executing JS callback and of course an assertion failure is thrown. 
While if there is only payment confirmation on router -> the flow is correct.

SO: the fix is -> we need to fire router cancel delegate calls only if there is one last controller left on the stack.


Toshi/Controllers/Payments/PaymentAddressViewController.swift
```diff
@@ -115,6 +116,9 @@ class PaymentAddressViewController: UIViewController {
         super.viewWillDisappear(animated)
 
         addressInputView.addressTextField.resignFirstResponder()
+
+        guard isMovingFromParentViewController else { return }
+        delegate?.paymentAddressControllerDidCancel(self)
```
marijnschilling: Can't we connect this to the back-button being pressed?

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
```diff
@@ -54,6 +58,21 @@ final class PaymentRouter {
...
+    private func increaseControllersStack() {
+        controllersInStackCount += 1
+    }
+
+    private func decreaseControllersStack() {
+        controllersInStackCount -= 1
+    }
+
+    private func cancelIfNeeded() {
+        let controllersStackIsEmpty = (controllersInStackCount == 0)
```
marijnschilling: nice to make a separate bool var for this!

I think we might call it `hasDimissedRootController` or `hasDismissedPaymentRouterFlow`. Because that might make our intention clearer that we only cancel when the Payment flow is finito 👋

Toshi/Controllers/Payments/PaymentAddressViewController.swift
yuliaveres: not sure if making a custom back button to just handle this is better 🤔  as this does the job 👌

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
yuliaveres: do you have a variable with different purpose than `controllersStackIsEmpty ` in mind ? .

Toshi/Controllers/Payments/PaymentAddressViewController.swift
marijnschilling: yeah but here you do need this check wether `viewWillDisappear` because it's presenting another view or dismissing itself. While with the back button you are always dismissing. So i think it would be more clear since that corresponds 1:1 with what we want.

Toshi/Controllers/Payments/PaymentRouter/PaymentRouter.swift
marijnschilling: No same var, just one of those suggested names that focusses more on the "intention" of this check.

Toshi/Controllers/Payments/PaymentAddressViewController.swift
designatednerd: I would at least throw a comment in with the `return` statement indicating why this is necessary - poking at `UINavigationControllerDelegate` I don't see an obvious way that's better than this, but I agree with @marijnschilling that it's a bit unclear.

yuliaveres:
that looks better @marijnschilling  thx!

PR #715, opened 2018-01-30
Fix edge case for payment confirmation being not dismissed

yuliaveres:
So the edge case is if a Dapp was searched on Browse. then in the end, when we do dismiss as usually, it is not actually dismissed as `topController` on `Navigator` is `searchController`.
We can just dismiss payment confirmation directly as we do on cancel.


Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -532,29 +532,34 @@ final class PaymentConfirmationViewController: UIViewController {
...
     }
 
     private func actuallyDismiss() {
         self.dismiss(animated: true, completion: nil)
     }
 
     @objc func didTapPayButton() {
-        payButton.showSpinner()
         guard shouldSendSignedTransaction else {
+            dismiss()
```
marijnschilling: I know it's actually from previous PR, but i wouldnt use a guard here. As this is just handling to equally important cases i would just use if else. Guard to me means more an early exit for a function when some conditions arent met.

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -532,29 +532,34 @@ final class PaymentConfirmationViewController: UIViewController {
     // MARK: - Action Targets
 
     @objc func cancelItemTapped() {
+        dismiss()
+        delegate?.paymentConfirmationViewControllerDidCancel(self)
+    }
+
+    private func dismiss() {
```
marijnschilling: I think I owuld keep this in one method 'cancelItemTapped' and call the `actuallyDismiss` just `dismiss`

Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
yuliaveres: it is all gone now 😎

PR #716, opened 2018-01-30
[IOS-430] Separate data source for all users

marijnschilling:
You can now use global search from new chat, new group and update group controller.

There are some things that still need a fix:

- [ ] In update group controller you shouldnt be able to select users that are already part of the group with global search.
- [ ] When you select 'Add by username' in the favorites controller, you should open global search as well.

global search


Toshi/Controllers/Browse/Views/BrowseSearchResultView.swift
```diff
@@ -20,12 +20,19 @@ import UIKit
...
+extension SearchSelectionDelegate {
+    func isSearchResultSelected(user: TokenUser) -> Bool { return false }
 }
 
 class BrowseSearchResultView: UITableView {
+    var isMultipleSelectionMode = false
 
     var searchResults: [TokenUser] = [] {
         didSet {
+            print("\n\n=== FOUND PEOPLE")
```
designatednerd: `DLog` or delete please

Toshi/Controllers/Browse/Views/BrowseSearchResultView.swift
```diff
@@ -76,19 +87,27 @@ extension BrowseSearchResultView: UITableViewDataSource {
...
+            assertionFailure("Could not get profile at indexPath: \(indexPath)")
+            return UITableViewCell()
         }
 
+        var isCheckmarkShowing: Bool?
+        if isMultipleSelectionMode {
+            isCheckmarkShowing = true
+        }
+
+        let tableData = TableCellData(leftImagePath: profile.avatarPath, title: profile.name, subtitle: profile.isApp ? profile.descriptionForSearch : profile.username, isChecked: isCheckmarkShowing)
```
designatednerd: so even if the item is not selected here, the checkmark will be checked. I think you might want to look at whether `isCheckmarkShowing` is the correct name for this variable or if there's some other way to pass this info through.

Toshi/Controllers/Browse/Views/BrowseSearchResultView.swift
```diff
@@ -76,19 +87,27 @@ extension BrowseSearchResultView: UITableViewDataSource {
...
+        let cellConfigurator = CellConfigurator()
+
+        guard let cell = tableView.dequeueReusableCell(withIdentifier: cellConfigurator.cellIdentifier(for: tableData.components), for: indexPath) as? BasicTableViewCell else {
```
designatednerd: I'd recommend a) pulling this farther up so if it fails, it fails before other logic executes and b) throwing an assertion failure if this happens so you don't just get the blank cell in dev.

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
```diff
@@ -6,7 +6,7 @@ import SweetUIKit
 final class ProfilesHeaderView: UIView {
     
     private let type: ProfilesViewControllerType
-    private let searchBar: UISearchBar?
+    let searchBar: UISearchBar?
```
designatednerd: Does this need to be settable from outside this view, or does it only need to be accessed?

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -43,25 +43,72 @@ protocol ProfilesListCompletionOutput: class {
 
 // MARK: - Profiles View Controller
 
-final class ProfilesViewController: UITableViewController, Emptiable {
-    
+final class ProfilesViewController: UIViewController, Emptiable, KeyboardAdjustable {
```
designatednerd: can you move the bits which are for `Emptiable` and `KeyboardAdjustable` to extensions wherever possible please?

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -70,31 +117,36 @@ final class ProfilesViewController: UITableViewController, Emptiable {
         controller.searchBar.delegate = self
         controller.searchBar.barTintColor = Theme.viewBackgroundColor
         controller.searchBar.tintColor = Theme.tintColor
-        controller.searchBar.placeholder = "Search by username"
-        
+
+        if type == .favorites {
```
designatednerd: if you're checking the type I would recommend doing so with a `switch` statement so it's easier to add future cases.

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -104,38 +156,43 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
+
         updateHeaderWithSelections()
 
         displayContacts()
-        setupEmptyView()
+
+        view.addSubview(tableView)
+        tableView.edges(to: view)
+
+        if type == .favorites {
```
designatednerd: same deal re: `switch`

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -150,12 +207,20 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
-        super.viewDidAppear(animated)
-        
-        scrollViewBottomInset = tableView.contentInset.bottom
+
+        if let indexPathForSelectedRow = searchResultView.indexPathForSelectedRow {
+            searchResultView.deselectRow(at: indexPathForSelectedRow, animated: true)
+        }
+
+        if #available(iOS 11.0, *) {
+        } else {
```
designatednerd: please throw in at least a comment about why this isn't necessary on iOS 11

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -343,25 +382,98 @@ final class ProfilesViewController: UITableViewController, Emptiable {
...
+            tableView.reloadData()
         }
     }
 }
 
-// MARK: - Search Bar Delegate
+// MARK: - Table View Delegate
+extension ProfilesViewController: UITableViewDelegate {
+    public func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
+        guard let profile = dataSource.profile(at: indexPath) else { return}
```
designatednerd: needs space after `return`

Toshi/Views/CellConfigurator/CellConfigurator.swift
```diff
@@ -40,11 +40,21 @@ class CellConfigurator {
         }
 
         if cellData.components.contains(.leftImage) {
+            if cellData.leftImage != nil {
             cell.leftImageView.image = cellData.leftImage
```
designatednerd: this needs to get tabbed over one now

Toshi/Views/CellSectionData/TableViewData.swift
```diff
@@ -64,17 +66,36 @@ public final class TableCellData {
...
+        self.title = title
+        self.subtitle = subtitle
+        self.leftImagePath = leftImagePath
+        self.details = details
+        self.switchState = switchState
+        self.badgeText = badgeText
+        self.isChecked = isChecked
+
+        setupComponents()
+    }
```
designatednerd: why add a separate initializer instead of just more parameters to the existing one?

Toshi/Views/CellSectionData/TableViewData.swift
```diff
@@ -121,5 +142,9 @@ public final class TableCellData {
         if badgeText != nil {
             components.insert(.badge)
         }
+
+        if isChecked != nil {
```
designatednerd: Yeah I think `isChecked` is probably not the right name here since the question is less "is the checkbox checked or not" than "is there a checkbox at all or not"

Toshi/Controllers/Profiles/ProfilesHeaderView.swift
designatednerd: wait, nevermind, that's only a problem with a var.

designatednerd:
@yuliaveres Let's merge this for now but there's definitely functionality which still needs to be added before I would call Marijn's story complete

PR #720, opened 2018-02-05
iOS-483: Table Header View for Wallet View

designatednerd:
Design's probably changing but i'll deal with that when it does: 

simulator screen shot - iphone 5s - 2018-02-05 at 15 22 24

Also did some tidying up in the UIView catch-all extension


Toshi/Controllers/Wallet/Views/WalletTableHeaderView.swift
```diff
@@ -17,91 +17,154 @@ import UIKit
...
 
         return label
     }()
 
-    private lazy var contentView: UIView = {
-        let view = UIView(withAutoLayout: true)
-        view.layer.cornerRadius = 10.0
-        view.backgroundColor = Theme.viewBackgroundColor
-        view.height(80).priority = .required
+    private lazy var copyButton: UIButton = {
```
yuliaveres: can we use `ActionButton` ?

designatednerd: Tried that, it's all the wrong colors. Also honestly I'd like to do a refactor on that since it's not a UIButton subclass, which I *really* think it ought to be

yuliaveres:
Thx !

PR #723, opened 2018-02-06
[IOS-479, IOS-485]: Wallet controller

yuliaveres:
Scanner tab is replaced with wallet controller;
2 types of generic cells are added;
tokens and collectibles fetch is added (including tests).


Toshi/Controllers/Wallet/WalletDatasource.swift
```diff
@@ -0,0 +1,91 @@
...
+    private func loadCollectibles() {
+        setupFakeItems()
+    }
+
+    // temporary
+    func setupFakeItems() {
+
+        switch itemsType {
+        case .collectibles:
+            let walletItem1 = Collectible(json: ["name": "Collectible name", "value": "90", "contract_address": "", "icon": "http://freedesignfile.com/upload/2017/09/house-icon-vector.jpg", "tokens": []])
```
yuliaveres: test code. will remove before merging and when we have endpoint working

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        cell.accessoryType = .disclosureIndicator
+
+        return cell
+    }
+}
+
+extension WalletViewController: UITableViewDelegate {
+
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        // show details
```
yuliaveres: separate PR

designatednerd: worth making this show the little quick alert with some placeholder text?

Toshi/Controllers/Wallet/WalletItemCellConfigurator.swift
```diff
@@ -0,0 +1,26 @@
...
+//  WalletItemCellConfigurator.swift
+//  Toshi
+//
+//  Created by Yuliia Veresklia on 06/02/2018.
+//  Copyright © 2018 Bakken&Baeck. All rights reserved.
+//
+
+import Foundation
+
+final class WalletItemCellConfigurator: CellConfigurator {
```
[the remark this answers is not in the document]
yuliaveres: no, that is the idea of configurator - to have generic configuration or custom whenever needed.

Toshi/Models/Navigator.swift
```diff
@@ -67,6 +67,17 @@ class Navigator: NSObject {
         parentViewController?.present(viewController, animated: true, completion: nil)
     }
 
+    static func presentScanner() {
+        guard let scannerController = tabbarController?.scannerController else {
+            assertionFailure("Can't retrieve scanner controller from tab bar controller")
```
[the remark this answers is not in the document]
yuliaveres: i guess why not. the root one, which is tab bar or navigator. do not want to bind it with Navigator though.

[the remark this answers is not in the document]
designatednerd: Do we have a way in the navigator to just grab the root VC? I think having it tied to the tab bar seems a bit inflexible given that the navigator should just present it over whatever the root VC is

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -224,20 +224,26 @@ final class EthereumAPIClient {
...
+                } catch {
+                    DispatchQueue.main.async {
+                        completion([], .invalidResponseJSON)
+                    }
                     return
                 }
 
-                for item in items {
-                    let collectible = Collectible(json: item)
+                for collectible in collectiblesResults.collectibles {
```
designatednerd: IIRC there's a method on array that should let you add all the objects in another array so you don't have to do this as a loop

designatednerd: (or alternately you could just return `collectibleResults.collectibles` and not even bother with the `resultItems` array)

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -267,20 +273,26 @@ final class EthereumAPIClient {
...
+                } catch {
+                    DispatchQueue.main.async {
+                        completion([], .invalidResponseJSON)
+                    }
                     return
                 }
 
-                for item in items {
-                    let token = Token(json: item)
+                for token in tokenResults.tokens {
```
designatednerd: Same comments here re `resultItems` and the for loop

Toshi/Controllers/Wallet/Models/Collectible.swift
```diff
@@ -15,29 +15,33 @@
...
+        case
+        name,
+        value,
+        contractAddress = "contract_address",
+        icon,
+        tokens
+    }
+}
+
+// Convenience class for decoding an array of Collectibles with the key "collectibles"
```
designatednerd: nit: This should have a third slash so that the description of the class shows up as documentation

Toshi/Controllers/Wallet/Models/Token.swift
```diff
@@ -15,21 +15,58 @@
...
+    let icon: String
+
+    lazy var displayValueString: String = {
+        let decimalNumberValue = NSDecimalNumber(hexadecimalString: self.value)
+        var decimalValueString = decimalNumberValue.stringValue
+
+        let valueFormatter = NumberFormatter()
+        valueFormatter.numberStyle = .decimal
+
+        if self.decimals > 0 {
```
designatednerd: maybe `guard self.decimals > 0 else...` here?

Toshi/Models/Navigator.swift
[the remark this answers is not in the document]
yuliaveres: root is the tab bar, and it owns scanner now.

Tests/EthereumAPIClientTests.swift
```diff
@@ -132,15 +132,15 @@ class EthereumAPIClientTests: QuickSpec {
...
-                            expect(collectible?.icon).to(equal("https://www.cryptokitties.co/icons/logo.svg"))
+//                            expect(collectibles).toNot(beNil())
+//                            expect(error).to(beNil())
+//
+//                            let collectible = collectibles.first as? Collectible
+//                            expect(collectible).toNot(beNil())
+//                            expect(collectible?.name).to(equal("Cryptokitties"))
+//                            expect(collectible?.value).to(equal("0x10"))
+//                            expect(collectible?.contractAddress).to(equal("0x0123456789012345678901234567890123456789"))
+//                            expect(collectible?.icon).to(equal("https://www.cryptokitties.co/icons/logo.svg"))
```
designatednerd: should this still be commented out?

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -15,7 +15,10 @@
 
 import UIKit
 
-final class WalletViewController: UIViewController {
+final class WalletViewController: UIViewController, Emptiable {
+
+    let walletHeaderHeight: CGFloat = 180
+    let sectionHeaderHeight: CGFloat = 44
```
designatednerd: 👍  - can these be `private` though?

Tests/EthereumAPIClientTests.swift
yuliaveres: still yes as currently endpoint is not working and returns an error. i will not merge until it dies work and succeeds

Toshi/Controllers/Wallet/Models/Collectible.swift
```diff
@@ -22,7 +22,7 @@ final class Collectible: Codable {
     let value: String
     let contractAddress: String
     let icon: String
-    let tokens: [CollectibleToken]
+    var tokens: [CollectibleToken]?
```
designatednerd: why does this need to be a `var` in addition to being optional?

yuliaveres: this object is used for both collectibles and particular collectible endpoints. in first case there are no tokens, only when u request for a separate collectible.
(regarding var - will change to `let`. .. :/ )

designatednerd: I understood why it's optional, I just didn't understand why it was a variable 😄

yuliaveres: pushed 👍

Toshi/Controllers/TabBarController.swift
```diff
@@ -152,20 +147,15 @@ class TabBarController: UITabBarController, OfflineAlertDisplaying {
             selectedIndex = 0
         case .messaging:
             selectedIndex = 1
-        case .scanner:
-            presentScanner()
+        case .wallet:
+            selectedIndex = 2
```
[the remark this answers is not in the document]
designatednerd: ^^ this is still open

Toshi/Models/Navigator.swift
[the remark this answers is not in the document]
designatednerd: ok, we'll leave it for the moment but I do think we should revisit this architecture at some point, I'm not really sure why it needs to specifically be tied to the tab bar.

Toshi/Controllers/TabBarController.swift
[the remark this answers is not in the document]
yuliaveres: aha 🤦‍♂️  sorry

PR #724, opened 2018-02-06
Actual Copy Functionality, Design Updates, and Quick Alert View

designatednerd:
In this PR: 
- Updated design of the header: 
simulator screen shot - iphone 5s - 2018-02-06 at 14 43 35
- Added monospaced fonts and support for the one we're using right now per Marek's request
- Cleaned up how we're grabbing fonts in the theme a little
- Added functionality to copy the QR code to the clipboard 
- Added a little convenience class for showing quick alerts for visual feedback: 
simulator screen shot - iphone 5s - 2018-02-06 at 14 43 37
  and made sure it works even with longer text: 
simulator screen shot - iphone 5s - 2018-02-06 at 14 45 16
-


Toshi/Controllers/Wallet/Views/WalletTableHeaderView.swift
```diff
@@ -30,28 +30,38 @@ final class WalletTableHeaderView: UIView {
 
     private lazy var qrCodeImageView: UIImageView = {
         let imageView = UIImageView()
-        let qrCodeImageSize: CGFloat = 40
+        let qrCodeImageSize: CGFloat = 28
 
         imageView.width(qrCodeImageSize)
         imageView.height(qrCodeImageSize)
 
+        imageView.image = #imageLiteral(resourceName: "qr")
```
yuliaveres: should not the user payment address qr code be here ?

Toshi/Views/QuickAlertView.swift
```diff
@@ -0,0 +1,121 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import TinyConstraints
+import UIKit
+
+/// A little quick alert that shows at the bottom of a given view.
+/// Useful for things where you wnat to provide visual feedback, but an
+/// alert would be complete overkill.
+class QuickAlertView: UILabel {
```
yuliaveres: do we want it to be `final` ?

designatednerd: yeah probably

Toshi/Controllers/Wallet/Views/WalletTableHeaderView.swift
designatednerd: No, per design they want an icon to make it clearer that you need to tap on it to get your actual QR code

PR #725, opened 2018-02-07
iOS-484: Tap to open wallet header

designatednerd:
In this PR: 
- Adds functionality requested in ticket
- Adds basic wallet QR code controller in prep for final design
- Creates a centralize QR code generator and tests it
- Simplifies getting string from a QR code image for tests


Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+class WalletQRCodeViewController: UIViewController {
```
yuliaveres: `final` ?

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    private lazy var qrCodeImageView: UIImageView = {
+        let imageView = UIImageView()
```
yuliaveres: return `UIImageView()` ?

Tests/UIImage+ReadQRCode.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  UIImage+ReadQRCode.swift
+//  Tests
+//
+//  Created by Ellen Shapiro (Work) on 2/7/18.
```
yuliaveres: same comments maybe here ?

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("Use the other initializer!")
+    }
+
+    // MARK: - View Setup
+
+    private func setupCloseButton() {
+        view.addSubview(closeButton)
+
+        closeButton.topToSuperview(offset: .largeInterItemSpacing)
```
yuliaveres: maybe better to use `safeAreaLayoutGuide` when available ? (iPhone X .. )

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
```diff
@@ -0,0 +1,109 @@
...
+
+        // Offset is * 2 so the same margin is applied to each side
+        qrCodeImageView.widthToSuperview(offset: -(.largeInterItemSpacing * 2))
+
+        NSLayoutConstraint.activate([
+            qrCodeImageView.heightAnchor.constraint(equalTo: qrCodeImageView.widthAnchor)
+        ])
+    }
+
+    private func setupAddressLabel(below viewToPinToBottomOf: UIView) {
```
yuliaveres: just curious - why do we pass a view here and do not pin to QR code image view always ?

Toshi/Utilities/QRCodeGenerator.swift
```diff
@@ -0,0 +1,48 @@
...
+struct QRCodeGenerator {
+
+    enum QRCodeType {
+        case
+        /// - username: The Toshi username which you wish to display for scanning.
+        addUser(username: String),
+        /// - address: The ethereum address you wish to display for scanning.
+        ethereumAddress(address: String)
+
+        fileprivate var scannableAddress: String {
```
yuliaveres: can we use `private` always ?

Tests/UIImage+ReadQRCode.swift
designatednerd: 🤦‍♀️

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
designatednerd: Good call - I thought I'd have to do more with it but nope!

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
designatednerd: 👍 good catch

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
designatednerd: good question - whenever there's some kind of dependency on a view I've been trying to pass it in so that if we need to add another view between the two views, we can just do that easily.

Toshi/Utilities/QRCodeGenerator.swift
designatednerd: unfortunately not here - because the `enum` is a different type than the class, `private` scoped variables and functions are only available within the enum. This is one of the few places where `fileprivate` is still necessary.

yuliaveres: ok, thx for explanation

designatednerd: to be fair i tried to put `private` on it first and it wouldn't compile 😛

PR #727, opened 2018-02-07
iOS-475: Navigation bar can disappear from browse VC when switching tabs

designatednerd:
Good catch @yuliaveres


Toshi/Controllers/Browse/BrowseViewController.swift
```diff
@@ -147,6 +147,8 @@ class BrowseViewController: SearchableCollectionController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
+        navigationController?.setNavigationBarHidden(false, animated: animated)
```
yuliaveres: i am wondering - is not it good place to set it back to `false` in the place where we set it to `true` at some point ?
looks like it is in `DisappearingNavBarViewController` .. ?

designatednerd: it is but it looks like switching tabs isn't triggering that properly

yuliaveres: ah okay

PR #729, opened 2018-02-08
[IOS-490]: Collectible details

yuliaveres:
Show tokens list under a collectible


Tests/EthereumAPIClientTests.swift
```diff
@@ -146,6 +146,32 @@ class EthereumAPIClientTests: QuickSpec {
...
                     }
                 }
+
+                it("gets a single collectibles") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: EthereumAPIClientTests.self), mockFilename: "getACollectible")
+                    subject = EthereumAPIClient(mockTeapot: mockTeapot)
+
+                    waitUntil { done in
+                        subject.getCollectible(contractAddress: "0xb1690c08e213a35ed9bab7b318de14420fb57d8c") { collectible, error in
+                            expect(collectible).toNot(beNil())
```
designatednerd: you could do this with a `guard let` and a `fail()` if you'd prefer to not need to do optionals throughout the rest of the test

Tests/Mocks/EthereumAPIClient/getACollectible.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "Cryptokitties",
+    "value": "0x10",
+    "contract_address": "0x0123456789012345678901234567890123456789",
+    "icon": "https://www.cryptokitties.co/icons/logo.svg",
+    "tokens": [{
+               "token_id": "abcdef0123456",
+               "name": "Kitten 423423",
+               "image": "https://storage.googleapis.com/ck-kitty-image/0x06012c8cf97bead5deae237070f9587f8e7a266d/467583.svg",
+               "description": "A kitten"
```
designatednerd: 😻

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -209,6 +209,52 @@ final class EthereumAPIClient {
         }
     }
 
+    func getCollectible(address: String = Cereal.shared.paymentAddress, contractAddress: String, completion: @escaping ((Collectible?, ToshiError?) -> Void)) {
+        self.activeTeapot.get("/v1/collectibles/\(address)/\(contractAddress)") { (result: NetworkResult) in
```
designatednerd: the `result` doesn't need the type or the parens

Tests/EthereumAPIClientTests.swift
```diff
@@ -146,6 +146,32 @@ class EthereumAPIClientTests: QuickSpec {
                         }
                     }
                 }
+
+                it("gets a single collectibles") {
```
designatednerd: english nit: This should just be "collectible" not "collectibles"

Toshi/Controllers/Wallet/CollectibleViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+        super.viewDidLoad()
+
+        view.backgroundColor = Theme.viewBackgroundColor
+        addSubviewsAndConstraints()
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        UIView.animate(withDuration: 0.5) {
```
designatednerd: a) this should take into account whether the `animated` parameter is true or false (same for `will disappear`)
b) I think maybe we might want to move this to one of the UINavigationControllerDelegate methods on the actual nav controller, since it looks like only the root VC of this controller is ever going to be that green color

Toshi/Controllers/Wallet/CollectibleViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+
+        AvatarManager.shared.avatar(for: token.image, completion: { image, path in
+            if token.image == path {
+                cell.leftImageView.image = image
+            }
+        })
+
+        return cell
+    }
+}
```
designatednerd: So this is a more general question coming out of my previous uses of the `DataSource` object as a separate object: Why is all this stuff not handled directly in the data source object instead of in the view controller? Putting it in the view controller makes it more difficult to reuse this code without having to rewrite all of these bits, and provides a layer of indirection that I'm not really sure is necessary. 

Curious what you think.

Toshi/Controllers/Wallet/WalletNavigationController.swift
```diff
@@ -18,7 +18,9 @@ import Foundation
 class WalletNavigationController: UINavigationController {
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
-        return .lightContent
+        let lastController = viewControllers.last
+
+        return (lastController is WalletViewController) ? .lightContent : .default
```
designatednerd: as of right now it appears you can change this to be more generic and see if the last controller is the same instance as the root view controller so you don't have to depend on the type

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -64,7 +64,7 @@ final class WalletViewController: UIViewController, Emptiable {
     }
 
     override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
+        super.viewWillAppear(false)
```
designatednerd: why does this not need to pass the animated parameter along to `super`?

Toshi/Views/RectImageTitleSubtitleTableViewCell.swift
```diff
@@ -0,0 +1,99 @@
...
+        leftImageView.centerY(to: contentView)
+        leftImageView.left(to: contentView, offset: BasicTableViewCell.horizontalMargin)
+        leftImageView.top(to: contentView, offset: BasicTableViewCell.imageMargin, relation: .equalOrGreater, priority: .defaultLow)
+        leftImageView.bottom(to: contentView, offset: -BasicTableViewCell.imageMargin, relation: .equalOrGreater, priority: .defaultLow)
+    }
+
+    private func setupSubtitleLabel() {
+        subtitleLabel.topToBottom(of: titleLabel, offset: BasicTableViewCell.smallVerticalMargin)
+        subtitleLabel.leftToRight(of: leftImageView, offset: BasicTableViewCell.interItemMargin)
+        subtitleLabel.right(to: contentView, offset: -BasicTableViewCell.horizontalMargin)
```
designatednerd: Why not put the title and subtitle into a stack view?

Tests/EthereumAPIClientTests.swift
yuliaveres: it is a typo 🤦‍♀️

Toshi/Controllers/Wallet/WalletViewController.swift
yuliaveres: mistake

Toshi/Controllers/Wallet/CollectibleViewController.swift
yuliaveres: I always used to deal with UI tableview on UI controller level.
 
Trying this is something I have as a plan 😄  But, having this code reused is only possible for controllers displaying cells of similar UI. 
Let take Collectibles list and Wallet controller with tokens list as an example. both theoretically could reuse datasource as contain same kind of objects: `Wallet Item`. the problem in this case would be first of all difference in design which caused me to have different cells, rather than those we use throughout the app. thus, the cell is composed in totally different way.- with Basic Table View Cell we will compose table data and use configurator. while in Collectibles list case - i have separate cell class and configure it in a simple old way.

But: it would be definitely interesting to try for controllers which both display BasicTableViewCell. then dealing with data source and configuring a table view in the same place might be a good idea ! as we won't duplicate `tableview datasource` code.

Toshi/Controllers/Wallet/WalletNavigationController.swift
```diff
@@ -44,9 +44,25 @@ class WalletNavigationController: UINavigationController {
...
-        self.navigationBar.titleTextAttributes = [ NSAttributedStringKey.foregroundColor: Theme.lightTextColor ]
+    func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
+
+        let isGoingToShowRootViewController = viewController is WalletViewController
+
+        let barTintColor = isGoingToShowRootViewController ? Theme.tintColor : nil
+        let shadowImage = isGoingToShowRootViewController ? UIImage() : nil
+        let titleTextAttributes = isGoingToShowRootViewController ? [ NSAttributedStringKey.foregroundColor: Theme.lightTextColor ] : nil
+
+        UIView.animate(withDuration: 0.5) {
```
designatednerd: this duration should still take into account whether `animated` is true or false

yuliaveres: oh right!

yuliaveres: done

designatednerd:
👍

PR #730, opened 2018-02-08
iOS-491: Token + Ether detail VC

designatednerd:
in this PR: 
- Added token / ether detail view: 

simulator screen shot - iphone 5s - 2018-02-08 at 14 06 16

NOT in this PR: 
- Changing the color of the nav bar, as that should be covered through @yuliaveres's work on #729. 
- Ability to view ether balance (coming in next PR since it involves some pretty significant updates to the wallet VC)


Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+class TokenEtherDetailViewController: UIViewController {
```
yuliaveres: `final` ?

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        label.textColor = Theme.darkTextColor
+        label.textAlignment = .center
+
+        return label
+    }()
+
+    private lazy var fiatValueLabel: UILabel = {
+        let label = UILabel()
+        label.font = Theme.preferredTitle3(range: 15...25)
+        label.textColor = Theme.darkTextColor.withAlphaComponent(0.5)
```
yuliaveres: don't we have light text color on `Theme` ? 🤔

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -96,6 +96,15 @@ final class WalletViewController: UIViewController, Emptiable {
         emptyView.isHidden = !datasource.isEmpty
         emptyView.title = datasource.emptyStateTitle
     }
+
+    private func currentItemType() -> WalletItemType {
+        let selectedIndex = tableHeaderView.selectedIndex
```
yuliaveres: we can get current items type from datasource (`itemType`)

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -146,8 +155,19 @@ extension WalletViewController: UITableViewDataSource {
...
-        let alert = UIAlertController.dismissableAlert(title: "Collectionable list coming soon")
-        Navigator.presentModally(alert)
+        switch currentItemType() {
+        case .token:
+            guard let token = datasource.item(at: indexPath.row) as? Token else {
+                assertionFailure("Can't retrieve item at index: \(indexPath.row)")
+                return
+            }
+
+            let detailView = TokenEtherDetailViewController(token: token)
```
yuliaveres: maye lets call controllers *...Controller* rather than *...View* ..? 🤔

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        stackView.addWithDefaultConstraints(view: buttonStackView)
+
+        buttonStackView.addArrangedSubview(sendButton)
+        buttonStackView.addArrangedSubview(receiveButton)
+    }
+
+    // MARK: - Setup For Token
+
+    private func configure(for token: Token) {
+        self.title = token.name
```
yuliaveres: not sure if it is needed to have the parameter here as we deal with controller's `var token` ..

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        tokenValueLabel.text = "\(token.displayValueString) \(token.symbol)"
+
+        AvatarManager.shared.avatar(for: token.icon) { [weak self] image, _ in
+            self?.iconImageView.image = image
+        }
+    }
+
+    private func configure(forETHBalance etherBalance: NSDecimalNumber) {
```
yuliaveres: same here about the parameter

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        self.title = Localized("wallet_ether_name")
+
+        self.iconImageView.image = #imageLiteral(resourceName: "ether_logo")
```
yuliaveres: are we using explicit `self` s ? 😄

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
designatednerd: we do, but this isn't it - this is a half-transparent black. :

Toshi/Controllers/Wallet/WalletViewController.swift
designatednerd: Oh derp, mah bad

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
designatednerd: i like the way `configure(for: token)` reads at the call site better than `configure(token)`

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
designatednerd: same thing - also makes it clearer that you're configuring for something other than the default

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
designatednerd: *hangs head in shame*

PR #731, opened 2018-02-09
Wallet QR VC Actual Design; Protocols!

designatednerd:
This PR: 
- Adds final design for the wallet QR code VC:
simulator screen shot - iphone 5s - 2018-02-09 at 11 10 20
- Adds functionality to split a string to a certain number of lines (useful for splitting the eth adddress in half, but I made it more generic and tested it just in case)
- Moves copy/pasting things to/from the clipboard to a protocol with a default implementation
- Moves sharing things with the system share sheet to a protocol with a default implementation


Toshi.xcodeproj/project.pbxproj
```diff
@@ -293,6 +293,8 @@
 		33643DC91FD6DA6E00229BE2 /* ProfilesAddedToGroupHeader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33CF51401FD56EA000293FD1 /* ProfilesAddedToGroupHeader.swift */; };
 		33643DCA1FD6DA6E00229BE2 /* ProfilesAddedToGroupHeader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33CF51401FD56EA000293FD1 /* ProfilesAddedToGroupHeader.swift */; };
 		33674E3F1FFFA6CE00089BF7 /* fetchUsersFromIDs.json in Resources */ = {isa = PBXBuildFile; fileRef = 33674E3E1FFFA6CD00089BF7 /* fetchUsersFromIDs.json */; };
+		336CC5DA202E136E0052511C /* ClipboardCopying.swift in Sources */ = {isa = PBXBuildFile; fileRef = 336CC5D9202E136E0052511C /* ClipboardCopying.swift */; };
```
yuliaveres: to all targets those 😏

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -351,6 +349,10 @@ final class ProfilesViewController: UITableViewController, Emptiable {
     }
 }
 
+// MARK: - Mix-in extensions
+
+extension ProfilesViewController: SystemSharing { /* mix-in */ }
```
yuliaveres: what does `mix-in` mean ? ..

Toshi/Protocols/SystemSharing.swift
```diff
@@ -0,0 +1,37 @@
...
+
+extension SystemSharing {
+
+    func shareWithSystemSheet(items: [Any]) {
+        let shareSheet = UIActivityViewController(activityItems: items, applicationActivities: [])
+
+        Navigator.presentModally(shareSheet)
+    }
+
+    func shareWithSystemSheet(item: Any) {
```
yuliaveres: do we want to pass default activities here as well ?

Toshi.xcodeproj/project.pbxproj
designatednerd: every damned time

Toshi/Controllers/Profiles/ProfilesViewController.swift
designatednerd: It's basically a way of saying "I am adding this functionality which does not require additional code"  - I forget what programming language it's based on (maybe Python?) but it's how I generally talk about protocols with default implementations

Toshi/Protocols/SystemSharing.swift
designatednerd: At this point we're not passing them anywhere. Maybe if we decide to do that in the future we can add it, but so far, I don't think it's worth it.

yuliaveres: ok

PR #732, opened 2018-02-10
Display ether as a token

designatednerd:
Now if you have ether, it displays in the list of tokens:
simulator screen shot - iphone 5s - 2018-02-09 at 16 25 27

and you can click through to the token ether details screen: 
simulator screen shot - iphone 5s - 2018-02-09 at 16 26 34

There's a TODO in here around displaying the fiat value in the cell since it involves some things I think may be related to things @yuliaveres is working on in other areas.


Toshi/Controllers/Wallet/Models/Token.swift
```diff
@@ -56,6 +57,63 @@ final class Token: Codable {
...
+/// A class which uses token to view Ether balances
+final class EtherToken: Token {
+
+    let wei: NSDecimalNumber
+
+    init(valueInWei: NSDecimalNumber) {
+        wei = valueInWei
+
+        let valueString = EthereumConverter.ethereumValueString(forWei: wei)
+        let updated = String(describing: valueString.dropLast(4))
```
yuliaveres: i would comment that 4 last are " ETH", or use extract into separate method value formatting without ETH.

Toshi/Controllers/Wallet/Models/Token.swift
```diff
@@ -56,6 +57,63 @@ final class Token: Codable {
...
+            // do nothing - this is read-only since it's lazy, but the compiler doesn't think so since it's still a var.
+        }
+    }
+
+    required init(from decoder: Decoder) throws {
+        fatalError("init(from:) has not been implemented")
+    }
+
+    func convertToFiat() -> String? {
+        return EthereumConverter.fiatValueString(forWei: wei, exchangeRate: ExchangeRateClient.exchangeRate)
```
yuliaveres: 👍

Toshi/Controllers/Wallet/TokenEtherDetailViewController.swift
```diff
@@ -64,37 +66,17 @@ final class TokenEtherDetailViewController: UIViewController {
         return button
     }()
 
-    private var token: Token?
-    private var etherBalance: NSDecimalNumber?
-    private var showFiatValue: Bool = false
+    private var token: Token
```
yuliaveres: 👍  added this in my PR as well 🙈

Toshi/Controllers/Wallet/WalletDatasource.swift
```diff
@@ -78,9 +78,36 @@ final class WalletDatasource {
...
 
     private func loadTokens() {
-        EthereumAPIClient.shared.getTokens { [weak self] items, _ in
-            self?.items = items
+        EthereumAPIClient.shared.getBalance(fetchedBalanceCompletion: { [weak self] balance, _ in
+
+            if balance.floatValue > 0 {
+                let etherToken = EtherToken(valueInWei: balance)
+                self?.updateOrAdd(walletItem: etherToken, atFront: true)
+                self?.delegate?.walletDatasourceDidReload()
```
yuliaveres: do we need to call it twice ? 🤔

Toshi/Controllers/Wallet/WalletDatasource.swift
```diff
@@ -78,9 +78,36 @@ final class WalletDatasource {
...
+            } // else, don't show ether balance.
+
             self?.delegate?.walletDatasourceDidReload()
+            EthereumAPIClient.shared.getTokens { items, _ in
+                self?.updateOrAdd(items: items)
+                self?.delegate?.walletDatasourceDidReload()
```
yuliaveres: maybe just here. ?

Toshi/Controllers/Wallet/WalletDatasource.swift
```diff
@@ -78,9 +78,36 @@ final class WalletDatasource {
...
+            }
+        })
+    }
+
+    private func updateOrAdd(items: [WalletItem]) {
+        items.forEach { self.updateOrAdd(walletItem: $0) }
+    }
+
+    private func updateOrAdd(walletItem: WalletItem, atFront: Bool = false) {
+        if let existingIndex = self.items.index(where: { $0.subtitle == walletItem.subtitle }) {
```
yuliaveres: is it it the best property to compare ? what if use `contractAddress` ? (can extract it to wallet item protocol i think)

Toshi/Controllers/Wallet/WalletDatasource.swift
designatednerd: Yes - once is for when the ether call returns, once is for when the other call returns.

Toshi/Controllers/Wallet/WalletDatasource.swift
designatednerd: If you only do it in one place you could wind up in a weird state where the ether client has updated the balance but this call has not returned yet. I'd rather refresh each when it comes in.

Toshi/Controllers/Wallet/WalletDatasource.swift
designatednerd: Hm - At least here I'm trying to compare the token abbreviation, since those should be unique per item coming in here, but maybe I'm missing something on the other side of the equation

PR #733, opened 2018-02-14
[IOS-492, IOS-496]: Send token controller

yuliaveres:
UI part here only;

feb-14-2018 09-14-34


Toshi/Controllers/Wallet/Models/Token.swift
```diff
@@ -47,6 +47,10 @@ final class Token: Codable {
         return decimalValueString
     }()
 
+    var isEthToken: Bool {
```
designatednerd: To bring it in line with how I'm referring to it in #732 but can you please rename this `isEtherToken`?

yuliaveres: i think this can be removed after https://github.com/toshiapp/toshi-ios-client/pull/732/files git merged

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+enum TokenType {
+    case eth
+    case nonEth
```
designatednerd: it may be worth distinguishing rather than between `eth` and `nonEth`  something like `canDetermineFiatValue` vs `cantDetermineFiatValue` (or something that means the same thing but isn't so damned long)

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        let label = UILabel(withAutoLayout: true)
+        label.text = String(format: Localized("wallet_token_balance_format"), self.token.symbol, self.token.displayValueString)
+        label.textColor = Theme.lightGreyTextColor
+        label.font = Theme.preferredProTextRegular(range: 13...15)
+        label.tag = SendTokenViews.balanceLabel.rawValue
+
+        return label
+    }()
+
+    private lazy var fiatValueLabel: UILabel = {
```
designatednerd: so is this the fiat value for what you're sending or the fiat value for your balance or ???

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        let label = UILabel(withAutoLayout: true)
+        label.text = EthereumConverter.fiatValueString(forWei: NSDecimalNumber.zero, exchangeRate: ExchangeRateClient.exchangeRate)
+        label.font = Theme.preferredHeavy()
+        label.setContentHuggingPriority(.required, for: .horizontal)
+        label.tag = SendTokenViews.fiatValueLabel.rawValue
+
+        return label
+    }()
+
+    private lazy var valueTextField: UITextField = {
```
designatednerd: Again, what type of value is being created here?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+
+        return view
+    }()
+
+    private lazy var maxButton: UIButton = {
+        let button = UIButton.borderedTintButton()
+        button.setTitle(Localized("wallet_max_value_title"), for: .normal)
+        button.addTarget(self, action: #selector(maxButtonTapped), for: .touchUpInside)
+        button.setContentCompressionResistancePriority(.required, for: .horizontal)
+        button.height(40)
```
designatednerd: can you add some kind of constant for `40` like `inlineButtonHeight` or something? Same deal for the corner radius.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    private func adjustViewsVisibility() {
+        view?.layoutIfNeeded()
+
+        // hide views
+        if viewConfiguration.hiddenViews.contains(.maxButton) {
+            maxButtonWidth.isActive = false
+            maxButtonZeroWidth.isActive = true
+            maxButton.alpha = 0
```
designatednerd: you should be able to avoid using all these constraints and alpha adjustments and just use `maxButton.hidden = true`

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+            return .nonEthInactive
+        }
+    }
+}
+
+enum TokenTypeViewConfiguration {
+    case ethInactive
+    case ethActive
+    case nonEthActive
+    case nonEthInactive
```
designatednerd: Same thing here about "eth" vs "canDetermineFiatValue".

It also might be worth making `active` and `inactive` into their own enum, and then making a view configuration something you can set up with a `tokenType` plus the `activeState` or whatever.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        case .ethInactive:
+            return .fiatValueLabel
+        case .nonEthActive:
+            return [.balanceLabel, .maxButton]
+        case .nonEthInactive:
+            return []
+        }
+    }
+
+    var hiddenViews: SendTokenViews {
```
designatednerd: Given that the views contained in `hiddenViews` never overlap with those in `visibleViews`, this seems like an awful lot of work to maintain these two separate lists instead of a single list. 

Is there a way to combine these into a single list?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+
+        return button
+    }()
+
+    private lazy var scanButton: UIButton = {
+        let button = UIButton.borderedTintButton()
+        button.setImage(#imageLiteral(resourceName: "scan"), for: .normal)
+        button.setTitle(Localized("wallet_scan_QR_label"), for: .normal)
+        button.imageEdgeInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 5)
+        button.titleEdgeInsets = UIEdgeInsets(top: 0, left: 5, bottom: 0, right: 5)
```
designatednerd: can you use the `.smallInterItemSpacing` for these margins please?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        stackView.addSpacing(1, after: valueContainerView)
+        stackView.addArrangedSubview(setupAddressContainerView())
+
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(didTapMainView(_:)))
+        view.addGestureRecognizer(tapGesture)
+
+        view.addSubview(continueButton)
+        continueButton.leftToSuperview(offset: 20)
+        continueButton.right(to: view, offset: -20)
+        continueButton.bottom(to: layoutGuide, offset: -20)
```
designatednerd: again can you please use the `interItemSpacing` here

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+
+        view.addSubview(stackView)
+        stackView.top(to: view)
+        stackView.leftToSuperview()
+        stackView.rightToSuperview()
+
+        let valueContainerView = setupValueContainerView()
+        stackView.addArrangedSubview(valueContainerView)
+        stackView.addStandardBorder(margin: CGFloat.defaultMargin)
+        stackView.addSpacing(1, after: valueContainerView)
```
designatednerd: why does this need spacing after the value container view? If it does, why add the border before inserting the spacing?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+    }()
+
+    private lazy var addressErrorHeight: NSLayoutConstraint = {
+        return addressErrorLabel.height(15)
+    }()
+
+    let token: Token
+    init(token: Token, view: UIView) {
+        self.token = token
+        self.viewConfiguration = token.symbol == "ETH" ? TokenType.eth.initialViewConfiguration : TokenType.nonEth.initialViewConfiguration
```
designatednerd: you should be able to use `isEtherToken` here.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        }
+
+        continueButton.isEnabled = errors.isEmpty && !valueText.isEmpty
+    }
+
+    private func showAddressErrorLabel() {
+        addressErrorZeroHeight.isActive = false
+        addressErrorHeight.isActive = true
+
+        UIView.animate(withDuration: 0.1, delay: 0, options: .curveEaseInOut, animations: {
```
designatednerd: this is a super-quick animation - is it even visible?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+                break
+            }
+        }
+
+        continueButton.isEnabled = errors.isEmpty && !valueText.isEmpty
+    }
+
+    private func showAddressErrorLabel() {
+        addressErrorZeroHeight.isActive = false
+        addressErrorHeight.isActive = true
```
designatednerd: you should be able to do this by updating the `isHidden` property on the address label instead of swapping out the constraints

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+
+        UIView.animate(withDuration: 0.1, delay: 0, options: .curveEaseInOut, animations: {
+            self.view?.layoutIfNeeded()
+        }, completion: nil)
```
designatednerd: nil completion parameters can be omitted (though you still need to have the trailing `)` after the last `}` or swift gets cranky)

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    private func hideAddressErrorLabelIfNeeded() {
```
designatednerd: this and the above can probably be combined into a single method which takes a parameter about whether the error label should be hidden or showing.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+    private func hideBalanceLabelIfNeeded() {
+        guard balanceValueHeight.isActive else { return }
+
+        balanceValueHeight.isActive = false
+        balanceValueZeroHeight.isActive = true
+
+        UIView.animate(withDuration: 0.1, delay: 0, options: .curveEaseInOut, animations: {
+            self.view?.layoutIfNeeded()
+        }, completion: nil)
+    }
```
designatednerd: same concerns about using hidden vs swapping out constraints, animation length, and not needing two methods here as with the ones for the address error label. 

Also since this is quite similar to the above, maybe you can even create a single method that shows or hides a passed in label. That could save you whole bunch of hassle.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        valueStackView.axis = .vertical
+        valueStackView.alignment = .fill
+
+        let margin = CGFloat.defaultMargin
+
+        containerView.addSubview(valueStackView)
+        valueStackView.leftToSuperview(offset: margin)
+        valueStackView.right(to: containerView, offset: -margin)
+        valueStackView.topToSuperview(offset: margin)
+        valueStackView.bottomToSuperview(offset: -5)
```
designatednerd: `interitemSpacing` throughout this method please

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        let margin = CGFloat.defaultMargin
+
+        addressContainerView.addSubview(addressStackView)
+        addressStackView.leftToSuperview(offset: margin)
+        addressStackView.right(to: addressContainerView, offset: -margin)
+        addressStackView.topToSuperview(offset: margin)
+        addressStackView.bottomToSuperview(offset: -margin)
+
+        addressStackView.addArrangedSubview(toTitleLabel)
+        addressStackView.addSpacing(10, after: toTitleLabel)
```
designatednerd: again, `interItemSpacing` please

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    @objc private func maxButtonTapped() {
+        valueTextField.text = token.displayValueString
+        valueTextField.layoutIfNeeded()
+        configureTextFieldRightView()
+    }
+
+    @objc private func swapButtonTapped() {
+        // Coming soon
```
designatednerd: alert or `QuickAlertView` for user feedback?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        addressStackView.alignment = .leading
+        addressStackView.clipsToBounds = true
+
+        let margin = CGFloat.defaultMargin
+
+        addressContainerView.addSubview(addressStackView)
+        addressStackView.leftToSuperview(offset: margin)
+        addressStackView.right(to: addressContainerView, offset: -margin)
+        addressStackView.topToSuperview(offset: margin)
+        addressStackView.bottomToSuperview(offset: -margin)
```
designatednerd: I recall making a convenience method on `stackView` where you can add a background that goes out underneath a stack view with a consistent margin around it... you might be able to use that instead of making both the container view and the stack view

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    private func configureTextFieldRightView() {
+        guard let rightViewLabel = valueTextField.rightView as? UILabel else { return }
+        guard let text = valueTextField.text else { return }
+
+        rightViewLabel.textColor = text.isEmpty ? Theme.placeholderTextColor : Theme.darkTextColor
+
+        rightViewLabel.font = valueTextField.font
+        rightViewLabel.sizeToFit()
```
designatednerd: why use `sizeToFit` here instead of using autolayout?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,661 @@
...
+        label.textColor = Theme.greyTextColor
+        label.font = Theme.preferredProTextMedium(range: 13...15)
+
+        return label
+    }()
+
+    private lazy var addressTextView: UITextView = {
+        let view = UITextView()
+        view.font = Theme.mediumMonospaced(size: 17)
+        view.text = addressPlaceholder
```
designatednerd: Oog, I hate when you have to do a placeholder in a UITextView. 

Have you considered isolating that logic (including handling changes from placeholder text to regular text) in a subclass instead of in the view controller?

Toshi/Controllers/Wallet/SendToken/SendTokenViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+
+        // We need to process only ethereum address intent and ignore all the rest
+        if let intent = QRCodeIntent(result: result) {
+            switch intent {
+            case .addressInput(let address):
+                controller.dismiss(animated: true, completion: { [weak self] in
+                    self?.configurator.destinationAddress = address
+                })
+            default:
+                controller.startScanning()
```
designatednerd: Might be worth showing some kind of alert noting that the scanned QR code was not a valid ethereum address before kicking them back to scan again

Toshi/Extensions/UIButton+Additions.swift
```diff
@@ -0,0 +1,33 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+extension UIButton {
+
+    static func borderedTintButton() -> UIButton {
```
designatednerd: why is this a tint button if everything in it is grey?

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -286,9 +286,16 @@
 "wallet_empty_tokens_description" = "Share your wallet address to\nreceive money and tokens.";
 "wallet_empty_tokens_button_title" = "Share Wallet Address";
 
+"wallet_amount_label" = "AMOUNT";
```
designatednerd: is this all caps in both iOS and Android or just in Android?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: `isFiatValueConvertible` ?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: for what are you sending

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: value which you send, i guess i should rename

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: stack vie will animate with `hidden` ?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: it does not need spacing ... just border, will remove

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: make it bit longer then .. 🤔

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: okay i will try it ! 👍

Toshi/Resources/Base.lproj/Localizable.strings
yuliaveres: caps ?

Toshi/Extensions/UIButton+Additions.swift
yuliaveres: should rename, was adjusting to latest design and it is not tint anymore 👍

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: yah though that's more of a boolean. Maybe you could make a struct that has two booleans - `isFiatValueConvertible` and `isActive` instead of the enums? 

I may be overthinking this.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: Can you clarify that in the name please

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: Yep, if you call animate on `layoutWithNeeded`! See attached playground. 
[stack view animate.playground.zip](https://github.com/toshiapp/toshi-ios-client/files/1723341/stack.view.animate.playground.zip)

Toshi/Resources/Base.lproj/Localizable.strings
designatednerd: CAPITAL LETTERS are also known as "caps" - ex, the caps lock key.

Toshi/Extensions/UIButton+Additions.swift
designatednerd: Might be fun to make it possible to just pass in the color.

Toshi/Resources/Base.lproj/Localizable.strings
yuliaveres: so yes - on both android and ios

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: does not behave as needed if right view is manage with layout..

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: eh but i want to use `options` where completion is not optional param.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: i do not need a background instead of container. i need the address section container as a part of / arranged subview of the main stack view.

designatednerd:
@yuliaveres can you merge in from the `send-flow` branch? I'm seeing a bunch of Marijn's changes which I think will go away if you merge.

yuliaveres:
@designatednerd done

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: ahhh ok - that makes more sense then. thank you!

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,568 @@
...
+
+        // Pasting while text view is a placehlder, the string from pasteboard is with whitespace in front
+        // we should just replace text with pasted string
+        if textView.text == addressPlaceholder && !text.isEmpty {
+            textView.text = text.trimmingCharacters(in: .whitespacesAndNewlines)
+
+            return false
+        }
+
+        return !isClearingPlaceholder
```
designatednerd: I think this comment got lost in the other stuff I was suggesting but did you give any thought to pulling the text-view-with-a-placeholder stuff out into a reusable view?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,568 @@
...
+
+    private lazy var scanButton: UIButton = {
+        let button = UIButton.borderedButton(with: Theme.lightGreyTextColor)
+        button.setImage(#imageLiteral(resourceName: "scan"), for: .normal)
+        button.setTitle(Localized("wallet_scan_QR_label"), for: .normal)
+        button.imageEdgeInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: CGFloat.smallInterItemSpacing)
+        button.titleEdgeInsets = UIEdgeInsets(top: 0, left: CGFloat.smallInterItemSpacing, bottom: 0, right: CGFloat.smallInterItemSpacing)
+        button.addTarget(self, action: #selector(scanButtonTapped), for: .touchUpInside)
+        button.height(inlineButtonHeight)
+        button.width(130)
```
designatednerd: i would recommend against hard-coding the width since this may be different in different languages. Is there a way to get it to size based on the size of title + image?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,568 @@
...
+
+        super.init()
+        self.adjustViewsVisibility()
+
+        NotificationCenter.default.addObserver(self, selector: #selector(keyboardDidHide(_:)), name: .UIKeyboardDidHide, object: nil)
+    }
+
+    func configureView(_ view: UIView) {
+        guard let layoutGuide = layoutGuide else {
+            assertionFailure("No know layout guide on configurator")
```
designatednerd: `known`?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,568 @@
...
+        guard addressErrorLabel.isHidden != hidden else { return }
+
+        addressErrorLabel.isHidden = hidden
+
+        UIView.animate(withDuration: 0.3, delay: 0, options: .curveEaseInOut, animations: {
+            self.view?.layoutIfNeeded()
+        }, completion: nil)
+    }
+
+    private func adjustBalanceLabelHidden(_ hidden: Bool) {
```
designatednerd: maybe for these you could use `to` instead of `_` - then at the call site it reads like: `adjustAddressErrorLabelHidden(to: true)`

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,568 @@
...
+        valueStackView.axis = .vertical
+        valueStackView.alignment = .fill
+
+        let margin = CGFloat.defaultMargin
+
+        containerView.addSubview(valueStackView)
+        valueStackView.leftToSuperview(offset: margin)
+        valueStackView.right(to: containerView, offset: -margin)
+        valueStackView.topToSuperview(offset: margin)
+        valueStackView.bottomToSuperview(offset: -CGFloat.smallInterItemSpacing)
```
designatednerd: should this be to the `layoutGuide`?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -0,0 +1,568 @@
...
+
+extension SendTokenViewConfigurator: UIGestureRecognizerDelegate {
+
+    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer,
+                           shouldReceive touch: UITouch) -> Bool {
+        if touch.view is UIControl {
+            return false
+        }
+
+        return true
```
designatednerd: worth validating that the gesture recognizer is the one you expect it to be in this? or at least throwing in a comment as to why this is necessary?

Toshi/Controllers/Wallet/SendToken/SendTokenViewController.swift
```diff
@@ -0,0 +1,125 @@
...
+}
+
+extension SendTokenViewController: ScannerViewControllerDelegate {
+    func scannerViewControllerDidCancel(_ controller: ScannerViewController) {
+        controller.dismiss(animated: true, completion: nil)
+    }
+
+    func scannerViewController(_ controller: ScannerViewController, didScanResult result: String) {
+
+        // We need to process only ethereum address intent and ignore all the rest
```
designatednerd: this comment should probably be updated since you're now also handling the admin login here

Toshi/Controllers/Wallet/SendToken/SendTokenViewModel.swift
```diff
@@ -0,0 +1,214 @@
...
+import Foundation
+
+enum TokenType {
+    case fiatRepresentable
+    case nonFiatRepresentable
+}
+
+enum PrimaryValue {
+    case token
+    case fiat
```
designatednerd: 👍  much better

Toshi/Controllers/Wallet/SendToken/SendTokenViewModel.swift
```diff
@@ -0,0 +1,214 @@
...
+
+        switch (viewConfiguration.tokenType, viewConfiguration.primaryValue) {
+        case (.fiatRepresentable, .fiat):
+            let ether = etherNumberFromFiatText(valueText)
+            if ether.isANumber {
+                final = ether.multiplying(byPowerOf10: EthereumConverter.weisToEtherPowerOf10Constant)
+            }
+
+        case (.fiatRepresentable, .token):
+            final = NSDecimalNumber(string: valueText, locale: Locale.current).multiplying(by: EthereumConverter.weisToEtherConstant)
```
designatednerd: I've got [some code you can steal from my branch](https://github.com/toshiapp/toshi-ios-client/pull/734/files#diff-e8718a2787b74208258d0292ca530fc4R39) on this - in the future I would *strongly* recommend keeping this type of of thing confined to the converter class, since that's easier to test and reuse.

Toshi/Resources/Base.lproj/Localizable.strings
designatednerd: OK - usually that's more of an Android thing. Maybe ping Marek on it next week when he's back with us

Toshi/Controllers/Wallet/SendToken/SendTokenViewModel.swift
yuliaveres: i will update that after we have confirmation 🤞

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: i will add a comment. don't see a reason of checking for a particular gesture as we have only one, which the configurator is delegate of.

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: why ? it is top section, and it should stick to its container..

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: Main reason is defensively against possible future gestures, but that's probably premature optimization, so a comment is fine

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: Oh DERP, missed that entirely. I'm not jetlagged at all!

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: taking control over text being edited or no is possible from this delegate call afaik, and the delegate is configurator ...... inside that class i can receive notification only after the text is actually updated 🤷‍♂️

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: looks like no. are you aware of any ?

yuliaveres: ok - using intristic content size + added insets work as needed

Toshi.xcodeproj/project.pbxproj
```diff
@@ -342,12 +342,6 @@
...
 		338472F1202A5DD4000F5A8A /* SFMono-Regular.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472E9202A5DD3000F5A8A /* SFMono-Regular.otf */; };
-		338472F2202A5DD4000F5A8A /* SFMono-Bold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472EA202A5DD3000F5A8A /* SFMono-Bold.otf */; };
-		338472F3202A5DD4000F5A8A /* SFMono-Bold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472EA202A5DD3000F5A8A /* SFMono-Bold.otf */; };
-		338472F4202A5DD4000F5A8A /* SFMono-Bold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472EA202A5DD3000F5A8A /* SFMono-Bold.otf */; };
-		338472F5202A5DD4000F5A8A /* SFMono-Semibold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472EB202A5DD4000F5A8A /* SFMono-Semibold.otf */; };
-		338472F6202A5DD4000F5A8A /* SFMono-Semibold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472EB202A5DD4000F5A8A /* SFMono-Semibold.otf */; };
-		338472F7202A5DD4000F5A8A /* SFMono-Semibold.otf in Resources */ = {isa = PBXBuildFile; fileRef = 338472EB202A5DD4000F5A8A /* SFMono-Semibold.otf */; };
```
designatednerd: Can we hold off on ripping out fonts until we're about to merge down to `master`? I know at least mono semibold is used on my branch

yuliaveres:
@designatednerd  any other thought in this ?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: Let's leave it for now - I think in the past what I've done is basically set up a secondary delegate but that's probably too big a hammer for this nail at this point

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
designatednerd: [addressed]

PR #735, opened 2018-02-16
Send token confirmation finish and max ether handling

yuliaveres:
Finalized confirmation , added alert for max ether value


Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -310,11 +310,38 @@ final class SendTokenViewConfigurator: NSObject {
                                      PaymentParameters.to: destinationAddress,
                                      PaymentParameters.value: finalValueInWeiHex]
 
-        if !token.canShowFiatValue {
-            params[PaymentParameters.tokenAddress] = token.contractAddress
+        if !self.token.canShowFiatValue {
+            params[PaymentParameters.tokenAddress] = self.token.contractAddress
```
designatednerd: these might be my fault but: is `self` necessary here?

designatednerd: Also is `canShowFiatValue` the best way to determine if this should be added, or would `isEtherToken` or checking that the contract address is not empty a better way?

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -310,11 +310,38 @@ final class SendTokenViewConfigurator: NSObject {
...
+        }
+
+        // If ether - check if maximum value and show alert
+        let weiValueHex = EthereumConverter.etherToWei(NSDecimalNumber(string: token.value)).toHexString
+        guard weiValueHex == finalValueInWeiHex else {
+            proceed(with: params)
+            return
+        }
+
+        let alert = UIAlertController(title: "Final transaction amount and network fees will be shown on the next screen", message: nil, preferredStyle: .alert)
```
designatednerd: 🚨 non-localized user-facing strings 🚨

Toshi/Controllers/Wallet/SendToken/SendTokenViewModel.swift
```diff
@@ -109,7 +109,6 @@ final class SendTokenViewModel {
             final = NSDecimalNumber(string: valueText, locale: Locale.current)
             let tokenToWeiPower = Int16(token.decimals)
             final = final.multiplying(byPowerOf10: tokenToWeiPower)
-            print(final.toHexString)
```
designatednerd: This can probably be updated to use the `EthereumConverter` methods I added for wei to eth and eth to wei now

Toshi/Controllers/Wallet/TokenSendConfirmationViewController.swift
```diff
@@ -93,6 +112,60 @@ class TokenSendConfirmationViewController: UIViewController {
...
+
+                let estimatedFeesFiatString = EthereumConverter.fiatValueStringWithCode(forWei: rawPaymentInfo.estimatedFees, exchangeRate: exchangeRate)
+                let estimatedFeesEtherString = EthereumConverter.ethereumValueString(forWei: rawPaymentInfo.estimatedFees, fractionDigits: fractionDigits)
+                strongSelf.networkFeesSection.setupWith(primaryCurrencyString: estimatedFeesFiatString, secondaryCurrencyString: estimatedFeesEtherString)
+
+                strongSelf.confirmButton.isEnabled = rawPaymentInfo.sufficientBalance
+
+                if !rawPaymentInfo.sufficientBalance {
+                    strongSelf.errorLabel.isHidden = false
+                    strongSelf.errorLabel.text = "Insuffisient balance: \(rawPaymentInfo.balanceString)"
```
designatednerd: a) 🚨 user-facing non-localized string 🚨 
b) Thank you, this is WAY clearer than what I was trying to do.

Toshi/Controllers/Wallet/Views/SendConfirmationSection.swift
```diff
@@ -19,15 +20,15 @@ class SendConfirmationSection: UIStackView {
...
+    private(set) lazy var primaryCurrencyLabel: UILabel = {
         let label = UILabel()
         label.textAlignment = .right
         label.textColor = Theme.darkTextColor
 
         return label
     }()
 
-    private let secondaryCurrencyLabel: UILabel = {
+    private(set) lazy var secondaryCurrencyLabel: UILabel = {
```
designatednerd: a) whoopsies on making this a let instead of a lazy var
b) Do you need to be able to access this outside the class now that you have the `setupWith` method? (same question on all three labels)

Toshi/Controllers/Wallet/SendToken/SendTokenViewModel.swift
yuliaveres: right !

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
yuliaveres: in general, for now - all three might work ! BUT, ideally, if we check for ETHER in particular - we should use `isEtherToken`. Who knows if later we might have other token with `canShowFiatValue = true` or empty `contractAddress` ..

Toshi/Controllers/Wallet/SendToken/SendTokenViewModel.swift
yuliaveres: opps , nope!  this is `nonFiatRepresentable`, and decimals vary per token. decimals determine the power here ☺️

Toshi/Controllers/Wallet/Views/SendConfirmationSection.swift
yuliaveres: a) don't need to access UI guys from outside of the class, will put `private` back 👍 
b) so why i use `setupWith`:put needed values from outside this time, can't instantiate section with values ready straight away; section should be build up in `init`, but values are put after we have those - after we fetch transaction skeleton 👍

yuliaveres:
Merging into feature branch

PR #737, opened 2018-02-17
iOS-495: Confirmation HUD

designatednerd:
Adds the ability to show a full-screen, blocking heads-up display for loading and confirmation: 

huds

One thing to be aware of is that you need to make sure to pass in the outermost view of the current view controller (the tab bar controller in the example gif above) or it won't cover the nav bar or the tab bar.


Toshi/Extensions/UIStackView+Additions.swift
```diff
@@ -138,4 +138,12 @@ extension UIStackView {
             spacerView.bottom(to: self)
         }
     }
+
+    /// Empties the Stack view of its arranged subviews.
+    func removeAllArrangedSubviews() {
+        for view in arrangedSubviews {
+            removeArrangedSubview(view)
+            view.removeFromSuperview()
```
yuliaveres: is not it enough to call `removeFromSuperview ` ?

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -311,6 +311,7 @@
 "wallet_send_confirmation_amount_title" = "Amount";
 "wallet_send_confirmation_fees_title" = "Est. Network Fees";
 "wallet_send_confirmation_total_title" = "Total";
+"wallet_send_confirmation_success_message" = "Transaction\nSumbitted";
```
yuliaveres: Submitted ?

Toshi/Views/LoadingHUD.swift
```diff
@@ -0,0 +1,176 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import TinyConstraints
+import UIKit
+
+/// A Heads-Up display for loading and success.
+class LoadingHUD: UIView {
```
yuliaveres: final

Toshi/Views/LoadingHUD.swift
```diff
@@ -0,0 +1,176 @@
...
+            stackView.addArrangedSubview(successImageView)
+            successImageView.image = image.withRenderingMode(.alwaysTemplate)
+            if let text = optionalText {
+                stackView.addArrangedSubview(textLabel)
+                textLabel.text = text
+            }
+        }
+
+        UIView.animate(withDuration: animated ? 0.1 : 0.0, animations: {
+            self.stackView.layoutIfNeeded()
```
yuliaveres: remember to call `layoutIfNeeded` before the animation block

Toshi/Extensions/UIStackView+Additions.swift
designatednerd: no it isn't - otherwise the stack view still sees it as an arranged subview.

Toshi/Resources/Base.lproj/Localizable.strings
designatednerd: *facepalm*

PR #738, opened 2018-02-17
[IOS-498]: Wallet Design updates

designatednerd:
Slightly different than what's in Sketch after verbal discussion with Marek, but here's where we landed: 

simulator screen shot - iphone 5s - 2018-02-17 at 10 49 12


Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -291,8 +291,10 @@ final class SendTokenViewConfigurator: NSObject {
         primaryValueTextField.resignFirstResponder()
 
         viewConfiguration.isActive = false
-        addressTextView.text = UIPasteboard.general.string
-        validateState()
+
+        if let pasteboardString = UIPasteboard.general.string, !pasteboardString.isEmpty {
+            destinationAddress = pasteboardString
+        }
```
designatednerd: you should be able to use the `clipboardCopying` protocol method here instead of accessing pasteboard directly

Toshi/Controllers/Wallet/WalletNavigationController.swift
```diff
@@ -85,13 +87,34 @@ class WalletNavigationController: UINavigationController {
...
+        let duration = animated ? 0.5 : 0.0
+        UIView.animate(withDuration: duration) {
+            self.navigationBar.barTintColor = barTintColor
+            self.navigationBar.shadowImage = shadowImage
+            self.navigationBar.titleTextAttributes = titleTextAttributes
+
+            self.navigationBar.layoutIfNeeded()
+        }
+    }
+}
```
designatednerd: this should probably go away - nav bar coloring is now handled in push and pop overrides

PR #739, opened 2018-02-20
[IOS-499]: Wallet datasource improvements 

yuliaveres:
Cache wallet items
add activity indicator if online and loading
add refresh control


Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -83,6 +97,16 @@ final class WalletViewController: UIViewController, Emptiable {
         emptyView.edges(to: layoutGuide(), insets: UIEdgeInsets(top: walletHeaderHeight + sectionHeaderHeight, left: 0, bottom: 0, right: 0))
     }
 
+    @objc private func refresh(_ refreshControl: UIRefreshControl) {
```
designatednerd: Does `valueChanged` only get called when it goes from not refreshing -> refreshing, or does it also get called again when it goes from refreshing to not refreshing? Might be worth adding something defensive to check the current value of `isRefreshing` on `refreshControl` here and bail out if it's not refreshing

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -83,6 +97,16 @@ final class WalletViewController: UIViewController, Emptiable {
...
+        guard Navigator.tabbarController?.reachabilityManager.currentReachabilityStatus != .notReachable else {
```
designatednerd: This seems like something that should be encapsulated in a convenience method - either on `Navigator` or as a stand alone `InternetConnection` class. For the sake of speed, maybe just do it on `Navigator` for now.

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -59,12 +69,16 @@ final class WalletViewController: UIViewController, Emptiable {
...
+
+        setupActivityIndicator()
     }
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
+        showActivityIndicatorIfOnline()
         datasource.loadItems()
+        tableView.reloadData()
```
designatednerd: why reload the data immediately, to get the cached items to show up? It might make more sense to fire this only if the load from cache actually produces any results.

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -83,6 +97,16 @@ final class WalletViewController: UIViewController, Emptiable {
...
     }
 
+    @objc private func refresh(_ refreshControl: UIRefreshControl) {
+        guard Navigator.tabbarController?.reachabilityManager.currentReachabilityStatus != .notReachable else {
+            refreshControl.endRefreshing()
+            return
+        }
+
+        datasource.loadItems()
+        tableView.reloadData()
```
designatednerd: Same as above re reloading data here instead of in the data source when the cache produces results

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -184,13 +213,18 @@ extension WalletViewController: SegmentedHeaderDelegate {
             return
         }
 
+        showActivityIndicatorIfOnline()
         datasource.itemsType = itemType
+        tableView.reloadData()
```
designatednerd: (once again with the reloading)

Toshi/Controllers/Wallet/WalletViewController.swift
yuliaveres: "Does valueChanged only get called when it goes from not refreshing -> refreshing" - **yes**

designatednerd: Interesting. That's not actually what I'd expect from that control, but I'll take your word for it if you've tested it

Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -184,13 +213,18 @@ extension WalletViewController: SegmentedHeaderDelegate {
...
 }
 
 extension WalletViewController: WalletDatasourceDelegate {
 
     func walletDatasourceDidReload() {
+        hideActivityIndicator()
+        refreshControl.endRefreshing()
+
         adjustEmptyStateView()
         tableView.reloadData()
```
yuliaveres: reload only if datasource selected item type is same as current one on the controller

designatednerd: This one made more sense which is why i didn't mark it

Toshi/Controllers/Wallet/WalletViewController.swift
designatednerd: ^is this reload still necessary since the data source calls `walletDataSourceDidReload` if there's anything in the cache and then when new data comes back?

yuliaveres: nope ! ...

PR #740, opened 2018-02-20
[IOS-500]: Show checksum encoded wallet address

yuliaveres:
Implements checksum encoding for addresses
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md

img_0086


Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
```diff
@@ -186,11 +187,11 @@ final class WalletQRCodeViewController: UIViewController {
     }
 
     @objc private func shareButtonTapped() {
-        shareWithSystemSheet(item: address)
+        shareWithSystemSheet(item: self.address)
     }
 
     @objc private func copyButtonTapped() {
-        copyToClipboardWithGenericAlert(address)
+        copyToClipboardWithGenericAlert(self.address)
```
designatednerd: why does this need `self` now?

Toshi/Extensions/String+Additions.swift
```diff
@@ -100,4 +100,33 @@ extension String {
         
         return lines.joined(separator: "\n")
     }
+
+    func toChecksumEncodedAddress() -> String? {
```
designatednerd: Is there someplace you can link to as the source for this algorithm - like is it documented as part of the Ethereum protocol or is this sample code from someplace?

Toshi/Extensions/String+Additions.swift
```diff
@@ -100,4 +100,33 @@ extension String {
...
+        guard EthereumAddress.validate(self) else { return nil }
+
+        let addressWithout0x = self.lowercased().dropFirst(2)
+        guard let data = addressWithout0x.data(using: .utf8) else { return nil }
+
+        let hash = (data as NSData).sha3(256).hexadecimalString
+
+        var output = "0x"
+        for (i, c) in addressWithout0x.enumerated() {
```
designatednerd: Please don't use single letters for variable names

Toshi/Extensions/String+Additions.swift
```diff
@@ -100,4 +100,33 @@ extension String {
...
+
+            let idx = index(startIndex, offsetBy: i)
```
designatednerd: I would also say this could be a more descriptive variable name - what is this the index of? The current letter?

Toshi/Extensions/String+Additions.swift
```diff
@@ -100,4 +100,33 @@ extension String {
...
+            let hashChar = hash[idx]
+
+            if let integer = Int(String(hashChar), radix: 16) {
```
designatednerd: might be able to do this part as a `guard let`

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
yuliaveres: cause self.address is checksum encoded

Toshi/Extensions/String+Additions.swift
tristan: I updated the PR description with a link to the EIP for it (which has the reference implementations in python and javascript)

tristan:
I think we should make a habit of using this form anywhere the address is displayed to the user.
so perhaps add it to the main wallet screen too? (the 0x123d....ABc3 view)

i think the only other place is the old deposit screen, which i believe is scheduled to be removed anyways from the looks.

otherwise QA 👍

Toshi/Extensions/String+Additions.swift
designatednerd: @tristan or @yuliaveres mind throwing a comment in the code itself with that link? It's good that it's in the PR description but it'll be a lot easier to find for future reference if it's in the code.

Toshi/Controllers/Settings/DepositMoney/DepositMoneyController.swift
```diff
@@ -32,14 +32,18 @@ class DepositMoneyController: UIViewController {
...
         self.init(nibName: nil, bundle: nil)
 
         title = Localized("deposit_money_title")
 
         items = [
             .header(Localized("deposit_money_header_text")),
-            .bulletPoint(Localized("deposit_money_1_title"), String(format: Localized("deposit_money_1_text"), Cereal.shared.paymentAddress)),
+            .bulletPoint(Localized("deposit_money_1_title"), String(format: Localized("deposit_money_1_text"), walletAddress)),
             .copyToClipBoard(Localized("copy_to_clipboard_action"), Localized("copy_to_clipboard_feedback"), #selector(copyToClipBoard(_:))),
             .QRCode(Cereal.shared.walletAddressQRCodeImage(resizeRate: 20.0)),
```
tristan: since you made changes here, can we replace
> .QRCode(Cereal.shared.walletAddressQRCodeImage(resizeRate: 20.0))

with the new `QRCodeGenerator` ? this is the only inconsistency left (scanned QR code from this is all lowercase).

(again i don't think it's necessary if it's a big job since this view is going to be removed soon)

Toshi/Controllers/Wallet/WalletQRCodeViewController.swift
```diff
@@ -148,8 +149,8 @@ final class WalletQRCodeViewController: UIViewController {
 
         setupCloseButton(in: container)
 
-        addressLabel.text = address.toLines(count: 2)
-        qrCodeImageView.image = QRCodeGenerator.qrCodeImage(for: .ethereumAddress(address: address))
+        addressLabel.text = self.address.toLines(count: 2)
+        qrCodeImageView.image = QRCodeGenerator.qrCodeImage(for: .ethereumAddress(address: self.address))
```
designatednerd: these probably don't need `self`

designatednerd: oh wait i see why that changed NEVERMIND

Toshi/Controllers/Settings/DepositMoney/DepositMoneyController.swift
designatednerd: I think I set up `walletAddressQRCodeImage` to use that under the hood but that's def worth double checking

PR #743, opened 2018-02-21
[IOS-502]: Handle longer than value decimals on wallet item

yuliaveres:
https://fabric.io/bakken--bck2/ios/apps/org.toshi.distribution/issues/5a8c404d8cb3c2fa63737ce8?time=1519084800000%3A1519171199000


designatednerd:
day late and a dollar short but can you please write a test for this when you have a second? :trollface:

PR #752, opened 2018-02-26
[IOS-520]: Swap values improvements

yuliaveres:
Allow swapping initial values;
do not hide max,swap buttons while presenting final value alert


Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -336,7 +343,12 @@ final class SendTokenViewConfigurator: NSObject {
 
     @objc private func swapButtonTapped() {
 
-        guard let primaryValueText = primaryValueTextField.text, !primaryValueText.isEmpty else { return }
+        guard let primaryValueText = primaryValueTextField.text, !primaryValueText.isEmpty else {
```
designatednerd: I feel like I've seen this combo of logic in a couple places - worth making an extension on `UITextField` for something like `hasText`? Or maybe `SweetUIKit` already has something like that?

PR #753, opened 2018-02-26
[IOS-521]: Reload wallet datasourcemore often

yuliaveres:
We need to add timer temporarily to reload wallet items.


Toshi/Controllers/Wallet/WalletViewController.swift
```diff
@@ -83,6 +87,12 @@ final class WalletViewController: UIViewController {
 
         showActivityIndicatorIfOnline()
         datasource.loadItems()
+
+        tokenDetailsViewController = nil
+
+        timer = Timer.scheduledTimer(withTimeInterval: 30.0, repeats: true) { [weak self] _ in
```
designatednerd: Out of curiosity, why is this timer necessary?

tristan: We want the wallet to auto reload while it's open. The original plan was to use PNs, but this is currently impossible as there seems to be absolutely no foreground handling of apple push notifications at the moment (we tried every possible thing we could find to make it work, to no avail).

We decided to do this in the short term just to get the build out as building in our own server push with websockets will be much more work (but desirable, we'll add it to our technical debt 😭).

designatednerd: Talked to Yulia via slack and it looks like pushes are coming through now (why they weren't yesterday, `¯\_(ツ)_/¯`)

And yeah we should probably start collecting bitcoins to pay off our tech debt 😛

PR #755, opened 2018-02-27
[IOS-523]: Improve send view fields activating areas

yuliaveres:
image
Reference ☝️  , apart from navigation bar which is not supposed to handle this.


Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -266,10 +290,13 @@ final class SendTokenViewConfigurator: NSObject {
...
 
         let valueContainerView = setupValueContainerView()
+        valueContainerView.addGestureRecognizer(valueEnablingTapGesture)
         stackView.addArrangedSubview(valueContainerView)
         stackView.addSpacing(CGFloat.mediumInterItemSpacing, after: valueContainerView)
         stackView.addStandardBorder(margin: CGFloat.defaultMargin)
-        stackView.addArrangedSubview(setupAddressContainerView())
+        let addressContainetView = setupAddressContainerView()
+        addressContainetView.addGestureRecognizer(addressEnablingTapGesture)
+        stackView.addArrangedSubview(addressContainetView)
```
designatednerd: nit: `addressContainerView`

Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -623,6 +650,10 @@ extension SendTokenViewConfigurator: UITextViewDelegate {
     func textView(_ textView: UITextView, shouldChangeTextIn range: NSRange, replacementText text: String) -> Bool {
         adjustAddressErrorLabelHidden(to: true)
 
+        if range.length == 0 && text == "\n" {
+            textView.resignFirstResponder()
```
designatednerd: I think you should be able to do this returning `false` in `textfieldshouldReturn` after calling this method

yuliaveres: it is `textView` 😞

designatednerd: aw dammit. 🤦‍♀️

PR #757, opened 2018-03-06
Dapp header animation

marijnschilling:
I think the code can use some improvements! Will go over it tomorrow, but any feedback would also be welcome ✌️ 

Will still need to add the `Cancel` button, but have to check how to animate that one.

header-animation


Toshi/Resources/Development.plist
```diff
@@ -91,5 +91,7 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>UIViewControllerBasedStatusBarAppearance</key>
```
yuliaveres: only in Development ?

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
```diff
@@ -73,25 +110,92 @@ final class DappsTableHeaderView: UIView {
...
+
+        searchFieldHeightConstraint?.constant = resize(1 - percentage, in: 36 ... 56)
+        searchFieldWidthConstraint?.constant = resize(percentage, in: minTextFieldWidth ... maxTextFieldWidth)
+        searchFieldBottomConstraint?.constant = percentage.map(from: 0 ... 1, to: -40 ... -15)
+
+        let shadowOpacity = Float((1 - percentage).map(from: 0 ... 1, to: 0 ... 0.3))
+        searchTextField.addShadow(xOffset: 1, yOffset: 1, radius: 3, opacity: shadowOpacity)
+
+        searchTextField.backgroundColor = UIColor.white.withAlphaComponent(fadeOut(percentage, in: 0.89 ... 1))
+        searchTextFieldBackgroundView.alpha = fadeIn(percentage, in: 0.89 ... 1)
```
yuliaveres: do not we need to call `layoutIfNeeded` after all these constraints changes ?

marijnschilling: no, they just animate like that :)

PR #760, opened 2018-03-12
Fix a mess of warnings

designatednerd:
Wanted to do this as a separate PR from some of the other fixes I'm planning on.


Toshi/Extensions/String+Additions.swift
```diff
@@ -81,7 +81,7 @@ extension String {
     /// - Parameter count: The number of lines to break the thing up into
     /// - Returns: The string split into newlines
     func toLines(count: Int) -> String {
-        let section = Int(round(Double(length) / Double(count)))
+        let section = Int(round(Double(self.count) / Double(count)))
```
marijnschilling: It should't need `self` right?

designatednerd: here it does since we have to distinguish between the passed-in `count` variable and the `count` which is calculated on the string itself.

PR #764, opened 2018-03-12
iOS-544: Cancel Button Resizing Fixes...ish

designatednerd:
cancel_sillies

This PR: 
- Moves the cancel button into a stack view and uses the show/hide on that rather than having it be separate
- Adds a text field with a custom inset rather than messing around with layer transforms
- Fixes a couple things causing autolayout to scream like its hair was on fire

Known issue: 
- If you delete all the way to the end of what you were inputting, the header re-expands instantly without hiding the cancel button. There's some weird shit going on there I haven't quite figured out, but the workaround to re-hide the cancel button is to tap outside the stack view (this is visible in the gif)


designatednerd:
@marijnschilling @yuliaveres let me know if I've missed something in terms of the requirements here - I'm going to try to figure out wtf is going on with the deletion tomorrow.

designatednerd:
@marijnschilling yes this is mentioned as a Known Issue that I plan to fix on another PR - basically in certain circumstances the code is causing the header to be reset. This PR is just to fix the fact that the cancel button was getting squished during text entry

Toshi/Views/InsetTextField.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+/*
+ A text field to allow you to easily inset the input as much as you want.
+ */
+class InsetTextField: UITextField {
```
yuliaveres: final

PR #766, opened 2018-03-13
[IOS-548]: See all and more frontpage improvements

yuliaveres:
compressed_search
screenshot at mar 13 12-36-16
screenshot at mar 13 12-36-24


Toshi/APIClients/DirectoryAPIClient.swift
```diff
@@ -135,7 +135,7 @@ final class DirectoryAPIClient: CacheExpiryDefault {
     ///                 - dappsFrontpageResults: A list of categories ncluding dapps, and categories info, or nil
     ///                 - toshiError: A toshiError if any error was encountered, or nil
     func getQueriedDapps(queryData: DappsQueryData, completion: @escaping QueriedDappsCompletion) {
-        var path = "/v1/dapps?offset=\(queryData.offset)"
+        var path = "/v1/dapps?offset=\(queryData.offset)&limit=\(queryData.limit)"
```
designatednerd: i think the docs for this method need to be updated to indicate that a `queryData` param is passed in instead of just the query string

Toshi/Controllers/Dapps/DappsDataSource.swift
```diff
@@ -167,16 +168,23 @@ final class DappsDataSource {
...
                 // Fetched categories
-                let fetchedDappsSections = frontPage.categories.flatMap({ category -> DappsDataSourceSection? in
 
-                    guard !strongOperation.isCancelled else { return nil }
+                var fetchedDappsSections: [DappsDataSourceSection] = []
 
-                    let items = category.dapps.map({ dapp -> DappsDataSourceItem in
+                for (index, category) in frontPage.categories.enumerated() {
+
+                    guard !strongOperation.isCancelled else { return }
```
designatednerd: Would it make more sense to put this guard statement outside the `for` loop?

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -296,6 +331,8 @@ extension DappsViewController: UITableViewDelegate {
             sofaController.load(url: url)
 
             Navigator.presentModally(sofaController)
+        case .seeAll:
+            break
```
designatednerd: can you throw in a comment here about why nothing happens when you select "see all"

Toshi/Controllers/Dapps/DappsDataSource.swift
yuliaveres: it is not actually needed ..

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -296,6 +331,9 @@ extension DappsViewController: UITableViewDelegate {
             sofaController.load(url: url)
 
             Navigator.presentModally(sofaController)
+        case .seeAll:
+            // We ignore selection as the cell containt action button which touch event do we process
```
designatednerd: `containt`->`contains`

Toshi/Controllers/Dapps/DappsDataSource.swift
```diff
@@ -162,21 +163,26 @@ final class DappsDataSource {
...
                 // Fetched categories
-                let fetchedDappsSections = frontPage.categories.flatMap({ category -> DappsDataSourceSection? in
 
-                    guard !strongOperation.isCancelled else { return nil }
+                var fetchedDappsSections: [DappsDataSourceSection] = []
+
+                for (index, category) in frontPage.categories.enumerated() {
 
-                    let items = category.dapps.map({ dapp -> DappsDataSourceItem in
+                    var items = category.dapps.map({ dapp -> DappsDataSourceItem in
```
designatednerd: I think this can be shortened to a one liner: 

```
category.dapps.map { DappsDataSourceItem(type: .dappFront, dapp: $0) }
```

PR #769, opened 2018-03-14
[IOS-549] Dapps api tests

marijnschilling:
Added tests for the api call to get the dapps front page and the test query result.

Wondering if I should test specific error cases, or are they generic?


designatednerd:
@marijnschilling can you please break the carthage changes into a separate PR?

marijnschilling:
Oh yeah pointing to wrong branch I guess 🙈

designatednerd:
I'd want to add some more stuff later to validate that the data is being parsed correctly but for now, 👍

PR #770, opened 2018-03-14
Wallet Header Design Feedback 

designatednerd:
simulator screen shot - iphone 5s - 2018-03-14 at 15 04 17

- Adds a color overlay to darken the wallet header 
- Removes an unnecessary border on the wallet tableview
- Adds and tests ability to middle-truncate strings in a custom fashion
- Uses custom middle-truncated string for wallet address in header.


Toshi/Controllers/Wallet/Views/WalletTableHeaderView.swift
```diff
@@ -99,9 +99,15 @@ final class WalletTableHeaderView: UIView {
 
         backgroundColor = Theme.tintColor
 
+        // + 10% black overlay
+        let darkeningView = UIView()
+        darkeningView.backgroundColor = UIColor.black.withAlphaComponent(0.1)
+        addSubview(darkeningView)
+        darkeningView.edgesToSuperview()
```
yuliaveres: better to add this view no top than change color ?

designatednerd: Basically what this will do is if the `tintColor` ever changes, this will make sure this view automatically changes with it. This would also prevent us from adding a color only used in one place.

designatednerd: Alternately, I could look into some way to programatically blend two colors, but this seemed a lot easier 😛

designatednerd: Actually blending colors programmatically isn't THAT bad so i'll do that instead 😄

designatednerd:
Gonna do the color mixing in another pr, so merging this

PR #771, opened 2018-03-14
[IOS-551]: Use separate tableView for search results on DappsViewController

yuliaveres:


Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -123,6 +122,28 @@ final class DappsViewController: UIViewController {
...
+    private lazy var searchResultsTableView: UITableView = {
+        let view = UITableView(frame: CGRect.zero, style: .grouped)
+        view.translatesAutoresizingMaskIntoConstraints = false
+
+        view.backgroundColor = Theme.viewBackgroundColor
+        BasicTableViewCell.register(in: view)
+        view.delegate = self
+        view.dataSource = self
+        view.layer.borderWidth = .lineHeight
+        view.layer.borderColor = Theme.borderColor.cgColor
```
designatednerd: why does this need a border?

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -151,6 +172,14 @@ final class DappsViewController: UIViewController {
         tableView.right(to: view)
         tableView.bottom(to: view)
 
+        view.addSubview(searchResultsTableView)
+        searchResultsTableView.topToBottom(of: headerView)
+        searchResultsTableView.left(to: view)
+        searchResultsTableView.right(to: view)
+        searchResultsTableView.bottom(to: view)
```
designatednerd: you could do this as `edgesToSuperview(excluding: .top)`

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -224,13 +255,19 @@ extension DappsViewController: UITableViewDataSource {
         switch item.type {
         case .goToUrl:
             let cell = tableView.dequeue(UITableViewCell.self, for: indexPath)
+
+            // Since UITableView class is used for some cells, we need to remove custom added subviews after cell is reused
+            cell.contentView.subviews.forEach({ $0.removeFromSuperview() })
```
designatednerd: would it work to centralize this before the switch: 

```swift
if let tableViewCell = cell as? UITableViewCell {
    //the line above
}
```

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -363,6 +400,8 @@ extension DappsViewController: UITableViewDelegate {
 extension DappsViewController: UIScrollViewDelegate {
 
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        guard scrollView != searchResultsTableView && searchResultsTableView.alpha < 1 else { return }
```
designatednerd: can you throw in a comment before returning about why we're not doing anything here

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -412,20 +451,25 @@ extension DappsViewController: DappsSearchHeaderViewDelegate {
...
 
-        if headerView.frame.height > -headerView.sizeRange.upperBound {
-            // perform after delay to not have the contentOffset animation interfere with showing the keyboard
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.2, execute: {
-                self.tableView.setContentOffset(CGPoint(x: 0, y: self.headerView.sizeRange.upperBound), animated: true)
-            })
-        }
+        statusBarStyle = UIStatusBarStyle.default
+
+        UIView.transition(with: searchResultsTableView, duration: 0.5, options: .curveEaseIn, animations: {
```
designatednerd: why this instead of `UIView.animate...`?

Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -412,20 +451,25 @@ extension DappsViewController: DappsSearchHeaderViewDelegate {
...
+            self.searchResultsTableView.alpha = 1
+        }, completion: nil)
```
designatednerd: I *think* if you've got a nil completion on most of these methods you can just omit that parameter

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
```diff
@@ -205,13 +209,27 @@ final class DappsTableHeaderView: UIView {
...
+        } else {
+            layoutIfNeeded()
         }
     }
 
+    func reveal() {
+        didScroll(to: revealedStateScrollPercentage)
+    }
+
+    func collapse() {
```
designatednerd: reveal and collapse what?

Toshi/Controllers/Dapps/DappsViewController.swift
yuliaveres: actually in my case it is not need as afair it also animated adding / removing of a view .. 🤔  i can replace it with animate, maybe implementation is simpler behind scenes

Toshi/Controllers/Dapps/DappsViewController.swift
yuliaveres: not in case of transition .. 😬

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
yuliaveres: if it is called on a header instance, that would mean revel/collapse header ?

Toshi/Controllers/Dapps/DappsViewController.swift
yuliaveres: i guess so ...

Toshi/Controllers/Dapps/DappsViewController.swift
yuliaveres: it does not :)

Toshi/Controllers/Dapps/DappsViewController.swift
yuliaveres: yay

Toshi/Controllers/Dapps/DappsViewController.swift
designatednerd: Yeah looking at [the docs for `transition`](https://developer.apple.com/documentation/uikit/uiview/1622574-transition) it looks like it mostly exists to help apply various transition sub-options of UIViewAnimationOptions so that you can have individual views which flip around or otherwise change the way that view controllers can as well

Toshi/Controllers/Dapps/DappsViewController.swift
designatednerd: :sad-trombone:

Toshi/Controllers/Dapps/DappsViewController.swift
yuliaveres: still not sure, i am dequeuing cell and returning it in each case block ... probably not. i would leave as it is

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
designatednerd: Hm...true, but you're calling them internally rather than externally here. 

Maybe `expand` instead of `reveal` since it's the opposite of `collapse` might make it read a little better when being called internally

Toshi/Controllers/Dapps/DappsViewController.swift
designatednerd: ohhh then that makes more sense to leave it

designatednerd: alternately: If you use a different cell identifier for each thing you dequeue, is this necessary?

yuliaveres: hmm sound like an idea

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
yuliaveres: nice thanks :)

PR #772, opened 2018-03-14
Fix minor things

marijnschilling:
Fixes: Position of header title
https://trello.com/c/UrAYTTQE/11-explore-the-decentralized-web-title-position

Fixes: padding of 20px on textfield in expanded state.
https://trello.com/c/DxNKpjUz/16-incorrect-left-padding-in-the-search-input

Looks like this: 
screenshot 2018-03-14 16 33 51


Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
```diff
@@ -169,19 +169,19 @@ final class DappsTableHeaderView: UIView {
         addSubview(iconImageView)
         addSubview(titleLabel)
 
-        searchFieldStackView.top(to: titleLabel, offset: .giantInterItemSpacing)
-
         iconImageView.size(CGSize(width: 48, height: 48))
         iconImageView.centerX(to: self)
-        iconImageView.bottomToTop(of: titleLabel, offset: -(.giantInterItemSpacing))
+        iconImageView.bottomToTop(of: titleLabel, offset: -10)
```
designatednerd: fwiw this is `.mediumInterItemSpacing`

Toshi/Views/InsetTextField.swift
```diff
@@ -20,7 +20,7 @@ import UIKit
  */
 final class InsetTextField: UITextField {
 
-    private let xInset: CGFloat
+    var xInset: CGFloat
```
designatednerd: if this is going to be an accessible var than the y inset probably should be as well for the sake of consistency

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
```diff
@@ -169,19 +169,19 @@ final class DappsTableHeaderView: UIView {
...
 
         titleLabel.centerX(to: self)
-
-        searchStackBottomConstraint = searchFieldStackView.bottom(to: self, offset: -40)
+        titleLabel.width(295, relation: .equalOrGreater)
+        titleLabel.height(20, relation: .equalOrGreater)
```
yuliaveres: do we need to set this / and make it `equalOrGreater ` if the font is fixed ? 🤔

Toshi/Theme/Theme.swift
```diff
@@ -43,6 +43,7 @@ extension CGFloat {
     static let smallInterItemSpacing: CGFloat = spacingx1
     static let mediumInterItemSpacing: CGFloat = spacingx2
     static let largeInterItemSpacing: CGFloat = spacingx4
+    /// 40
```
yuliaveres: why do we need this comment ?

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
designatednerd: is the font fixed or is it dynamic?

Toshi/Theme/Theme.swift
designatednerd: I'm going to be looking at this more deeply on another PR - trying to find a better way to make the value of these more obvious at the call site

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
marijnschilling: I tried making it dynamic, but it doesn't really work (only responds when you restart the app) So I decided it would be good to keep it static, it's not super important information anyway. 

So i will remove the equalOrGreater! ✌️

Toshi/Theme/Theme.swift
marijnschilling: Oh yes, this was just for testing purposes, will remove it in this branch. 🛀

Toshi/Views/InsetTextField.swift
marijnschilling: I <3 consistency.

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
marijnschilling: 💡 Great! Will adjust. And if you just always review my PR's we don't need the comments with these values ;)

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
```diff
@@ -171,11 +171,11 @@ final class DappsTableHeaderView: UIView {
 
         iconImageView.size(CGSize(width: 48, height: 48))
         iconImageView.centerX(to: self)
-        iconImageView.bottomToTop(of: titleLabel, offset: -10)
+        iconImageView.bottomToTop(of: titleLabel, offset: -(.mediumInterItemSpacing))
 
         titleLabel.centerX(to: self)
-        titleLabel.width(295, relation: .equalOrGreater)
-        titleLabel.height(20, relation: .equalOrGreater)
+        titleLabel.width(295)
```
designatednerd: Why a fixed width rather than pinning to the sides? How would this be affected if copy changed to something longer?

Toshi/Controllers/Dapps/Views/DappsTableHeaderView.swift
```diff
@@ -171,11 +171,11 @@ final class DappsTableHeaderView: UIView {
...
+        titleLabel.height(20)
```
designatednerd: Why not let the label set its own height based on its intrinsic content size?

PR #773, opened 2018-03-14
Add + use programmatic color mixing

designatednerd:
This seemed like it would be harder than it was. Yay!


designatednerd:
screen shot 2018-03-14 at 4 31 12 pm

^^ This is in the playground I was using to test the various things I found on stack overflow and then turn them into something readable 😛

designatednerd:
[ColorPallete.playground.zip](https://github.com/toshiapp/toshi-ios-client/files/1811846/ColorPallete.playground.zip)

(aforementioned playground)

Toshi/Extensions/UIColor+Additions.swift
```diff
@@ -15,6 +15,58 @@
...
 import UIKit
 
+extension UIColor {
+
+    /// Mixes one color with the caller at a given alpha.
+    ///
+    /// For instance, if a designer has added a black layer at 10% opacity to a view, you can mix colors
+    /// by calling this method as `color.mixedWith(.black, atAlpha: 0.1)`.
+    ///
+    /// Note that it looks screwy if you apply the alpha to the color rather than passing it into this
```
marijnschilling: You mean when you use colors with alpha values for mixing?
If that is the  case it would be clearer to say `... looks screwy if you pass in colors with an alpha component.`

Toshi/Extensions/UIColor+Additions.swift
```diff
@@ -15,6 +15,58 @@
...
+    /// method with an alpha of `1.0`, so please don't do that.
+    ///
+    /// - Parameters:
+    ///   - otherColor: The color to blend the caller with.
+    ///   - alpha: A transparency value between 0 and 1. An assertion failure will fire if this is not passed in correctly.
+    /// - Returns: The blended colors
+    func mixedWith(_ otherColor: UIColor, atAlpha alpha: CGFloat) -> UIColor {
+        guard alpha >= 0.0 && alpha <= 1.0 else {
+            assertionFailure("Please use an alpha value between 0 and 1")
```
marijnschilling: Will an `assertionFailure` crash the program? We could also just `clamp` the input value to 0...1. But I guess its more unclear if we do this under the hood...

Toshi/Extensions/UIColor+Additions.swift
```diff
@@ -15,6 +15,58 @@
...
+            return self
+        }
+
+        let beta = 1.0 - alpha
```
marijnschilling: why is it called beta?

Toshi/Extensions/UIColor+Additions.swift
designatednerd: Only during debug builds - that's the benefit of `assertionFailure` over `fatalError`

Toshi/Extensions/UIColor+Additions.swift
designatednerd: heh, from one of the SO things I was looking at - I should probably rename that

PR #774, opened 2018-03-14
Add SwiftLint to Danger

designatednerd:
This PR makes Danger run `SwiftLint` and then add inline comments where `SwiftLint` warnings occur, because I'm tired of chasing these down myself. 😄


designatednerd:
^^ In the above automated comments you can see what it looks like when a SwiftLint issue is found, and then when it is resolved. 

Nitpicking robots ftw

PR #775, opened 2018-03-15
Dapp browser

yuliaveres:
New dapps browser feature


Toshi/Controllers/Dapps/DappsListViewController.swift
```diff
@@ -0,0 +1,125 @@
...
+}
+
+extension DappsListViewController: DappsDataSourceDelegate {
+
+    func dappsDataSourcedidReload(_ dataSource: DappsDataSource) {
+        tableView.reloadData()
+    }
+
+    func dappsDataSourceDidEncounterError(_ dataSource: DappsDataSource, _ error: ToshiError) {
+
```
designatednerd: probably worth a comment as to why this is implemented with no content

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -10,12 +10,18 @@
...
-/* Browse */
-"browse-navigation-title" = "Search";
-"browse-search-placeholder" = "Search or type the url";
+/* Dapps */
+"dapps-search-placeholder" = "Search or enter Đapp URL";
+"dapps-search-header-title" = "Explore the decentralized web.";
 
 "browse-top-rated-apps" = "Top Rated";
 "browse-featured-dapps" = "Featured";
+"dapps-section-title" = "Dapps";
```
designatednerd: Should this and the couple things below use the Ɖ symbol instead of the letter `D` for Dapps?

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -80,7 +81,8 @@ final class SOFAWebController: UIViewController {
 
     private lazy var backButton: UIButton = {
         let view = UIButton(type: .custom)
-        view.bounds.size = CGSize(width: 44, height: 44)
+        view.size(CGSize(width: 44, height: 44))
```
designatednerd: for `44` there exists a `defaultButtonHeight` extension var on CGFloat (see also line 94 and 115)

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -127,15 +204,15 @@ final class SOFAWebController: UIViewController {
         view.addSubview(webView)
         view.addSubview(toolbar)
 
-        toolbar.height(44)
-        toolbar.bottom(to: layoutGuide())
+        toolbar.top(to: layoutGuide())
         toolbar.left(to: view)
         toolbar.right(to: view)
+        toolbar.height(44)
```
designatednerd: There is also a `defaultBarHeight` on `CGFloat`

Toshi/Resources/Base.lproj/Localizable.strings
yuliaveres: good catch 👍

PR #776, opened 2018-03-15
[IOS-552]: Dapp selection navigation

yuliaveres:
dapps-navigation

https://trello.com/c/kaGmeK6b/22-tapping-the-dapp-in-the-search-results-should-take-you-straight-to-the-browser-view-not-to-the-dapp-profile
https://trello.com/c/ZMzRxFIV/27-closing-the-browser-should-kick-you-straight-to-the-dapps-page-not-back-to-the-search


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -263,6 +269,8 @@ final class SOFAWebController: UIViewController {
     }
 
     @objc private func closeButtonTapped() {
+        // we do not need to wait till completion as DappsController is cancelling search and reloading, is better to dismiss to already reloaded non-search mode
```
designatednerd: I actually don't think this comment is necessary: 
- It's `willFinish` instead of `didFinish`, which implies it's before completion fires
- This happens a lot more places than the dapps controller, so if you want to add that comment, it might be better added in the dapps controller itself.

PR #778, opened 2018-03-15
[IOS-553] :Improve dapps search header layout animations

yuliaveres:
animation


Toshi/Controllers/Dapps/DappsDataSource.swift
```diff
@@ -145,8 +145,11 @@ final class DappsDataSource {
     }
 
     func cancelSearch() {
-        queryData.searchText = ""
-        queryData.isSearching = false
+        var presentQueryData = queryData
+        presentQueryData.searchText = ""
+        presentQueryData.isSearching = false
+
+        queryData = presentQueryData
```
designatednerd: why pull this out and create another instance, to trigger the setter? might be easier to just call `fetchItems` after adjusting the contents.

yuliaveres: not sure what do you mean by "after adjusting the contents.".
The reason for pulling that out - is to trigger the setter once, not twice (by searchText and isSearching change), which causes glitch.

designatednerd: ahhhhhh ok, that makes more sense. Might be worth throwing in a comment for the next person who wants to "clean it up" 😛

PR #779, opened 2018-03-15
[IOS-554]: Add pull to refresh on Browser

yuliaveres:
According to design we need to have a way to reload current page on Browser controller. As it was agreed - we are going to add Pull to refresh on webView's scrollView


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -585,6 +597,10 @@ extension SOFAWebController: WKUIDelegate {
 
         return nil
     }
+
+    public func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        refreshControl.endRefreshing()
```
designatednerd: does `didFinish` also get called when there's an error?

yuliaveres: not sure, but should we end refresh in case it fails ?

yuliaveres: i will implement didFinish as well

designatednerd: I would not recommend auto-refreshing if it fails - that'll infinite loop if there's a server error, or the user is not connected to the internet. 

and i think you need to implement the error one, this method is didFinish.

yuliaveres: @designatednerd done

PR #781, opened 2018-03-16
[IOS-556]: Get rid of Awesome cache

yuliaveres:
Awesome cache caused us lots of inconvenience due to all disk operations being executed on main thread. (Already replaced on AvatarManager)


Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -203,7 +200,7 @@ final class EthereumAPIClient {
             }
 
             DispatchQueue.main.async {
-                self?.cache.setObject(balance, forKey: EthereumAPIClient.CachedBalanceKey)
+                self?.cache.set(value: balance.stringValue, key: EthereumAPIClient.CachedBalanceKey)
```
designatednerd: should this still be on the main thread?

Toshi/APIClients/idAPIClient.swift
```diff
@@ -41,22 +40,6 @@ final class IDAPIClient: CacheExpiryDefault {
...
-    }()
-
-    private lazy var contactCache: Cache<TokenUser> = {
-        do {
-            return try Cache<TokenUser>(name: "tokenContactCache")
-        } catch {
-            fatalError("Couldn't instantiate the contact cache")
-        }
-    }()
-
```
designatednerd: why remove caching entirely from this  API?

Toshi/Models/Currency.swift
```diff
@@ -38,153 +36,148 @@ final class CurrenciesCacheData: NSObject, NSCoding {
     }
 }
 
-final class Currency: NSObject, NSCoding {
+final class Currency: Codable {
```
designatednerd: 🎉 `Codable` 🎉

Toshi/APIClients/EthereumAPIClient.swift
yuliaveres: not anymore

Toshi/APIClients/idAPIClient.swift
yuliaveres: it is not currently used at all. (maybe was for searching previously ? no usage of any of these right now)

PR #783, opened 2018-03-19
Fix sofa web controller not being deallocated properly

yuliaveres:


designatednerd:
Is there any way besides the close button to get out of this VC, like swiping or something?

yuliaveres:
nope. no other way

PR #787, opened 2018-03-20
Add auto-generated `Localized` with Marathon and Stencil

designatednerd:
Apologies for the length and boringness of this PR, since it adds and uses a bunch of machine-generated code. 

In this PR: 
- Use the [Marathon](https://github.com/JohnSundell/Marathon) Swift scripting helper and the [Stencil](https://github.com/kylef/Stencil) Swift templating engine to suck in the contents of `Localizable.strings` and generate a Swift file with method names tied to each of the keys, and which will show the english value of them when you use the documentation viewer: 
screen shot 2018-03-19 at 4 39 02 pm

- Set up a bash script to run Marathon appropriately, in a Bash folder so we can keep other bash scripts in there if need be. 
- Set up a build phase on the Debug target to automatically run the Marathon script before compiling
- Add some documentation about the usage of Marathon and Stencil. 

In future PRs: 
- Add support for Localized Plurals
- Use the fact that enums are set up for all the keys to add tests checking that there are values for all keys in the enums
- Add support for code gen for more stringly-typed things such as UIImages from Asset Catalogs.


Podfile.lock
```diff
@@ -174,4 +174,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: b99e22ed0fe9588c0bde91e0fd169f009adf41ac
 
-COCOAPODS: 1.3.1
+COCOAPODS: 1.4.0
```
designatednerd: apologies for the CP noise here, I wound up installing and uninstalling some stuff via pods before settling on Marathon. But we should be using 1.4.0 anyway so ¯\_(ツ)_/¯

marijnschilling:
😯 ooh these look like cool helpers! 

I'm just wondering, how would i go about adding a localized string now? Do i have to add it in the `LocaliedStrings.swift` file, or is that what get's autogenerated?

And what about the "old" helper `Localized("string_key") can we remove that now?

designatednerd:
Yes, to add a new localized string: 

1) Add it to `Localizable.strings` 
2) Build the debug target, assuming you have marathon and the appropriate dependencies installed
3) ✨ MAGIC ✨ 
4) Use the new method, which will be based on the new key. 

And I already got rid of the old `Localizable`, that's why I went through and made the eleventy frillion changes here.

Toshi/MessagesUI/Objects/Extensions.swift
```diff
@@ -126,16 +126,12 @@ extension UIBarButtonItem {
...
+        item.title = Localized.nav_bar_back
 
         return item
     }
 }
 
-func Localized(_ key: String, tableName: String? = nil, bundle: Bundle = Bundle.main, value: String? = nil, comment: String? = nil) -> String {
-    return NSLocalizedString(key, tableName: tableName, bundle: bundle, value: value ?? "", comment: comment ?? "")
-}
-
```
designatednerd: Here is where the old `Localized` is removed

README.md
```diff
@@ -9,3 +9,25 @@
 We use [SwiftLint](https://github.com/realm/SwiftLint) and [Danger](https://github.com/danger/danger) to keep our code consistent and our PR's mergeable.
 
 Check our SwiftLint [installation guide](https://github.com/toshiapp/toshi-ios-client/blob/master/installation-guide.md) to setup SwiftLint.
+
+### Code Generation
+
+We're using [Marathon](https://github.com/JohnSundell/Marathon) and [Stencil](https://github.com/kylef/Stencil). 
+
+You should install Marathon using [the instructions on how to install from the Swift Package manager](https://github.com/JohnSundell/Marathon#on-macos) - the version on Homebrew isn't working correctly.
```
marijnschilling: install Marathon in Toshi project folder? or? Where do I need to install this?

README.md
```diff
@@ -9,3 +9,25 @@
...
+
+### Code Generation
+
+We're using [Marathon](https://github.com/JohnSundell/Marathon) and [Stencil](https://github.com/kylef/Stencil). 
+
+You should install Marathon using [the instructions on how to install from the Swift Package manager](https://github.com/JohnSundell/Marathon#on-macos) - the version on Homebrew isn't working correctly. 
+
+There's some [weirdness going on with Stencil's test dependencies](https://github.com/kylef/Spectre/pull/34), so using a `Marathonfile` results in weird installation issues. In order to actually get Marathon to work for this project, please follow the following steps, IN THIS ORDER: 
+
+1. Remove the `~/.marathon` folder - this will remove all caches.
```
marijnschilling: I can't find this folder., or where is it supposed to be?

screenshot 2018-03-20 15 48 33

README.md
designatednerd: good point, that's not specified in the instructions from Homebrew - it actually doesn't matter where it's installed since the binary gets copied into `/usr/local/bin` after it's built.

README.md
designatednerd: The `~/` is a prefix that indicates your user folder. So mine is `/Users/ellenshapiro/.marathon/`

PR #789, opened 2018-03-21
[IOS-560]: Improve reload on search dapps

yuliaveres:
UITAbleViewCell reload animation behaviour is weird and causes some bouncy effect even using animation option .none. I was experimenting with reload by sections and reload by rows. The end result of both was tested and according to the feedback, bouncy effect caused by tableView reload inside beginUpdates/endUpdates is quite annoying and avoiding any reload animation feels better by a user.

Here we can see reload by rows.
mar-21-2018 13-17-48
Here we see slowed down reload:
rows

It was decided to stick to plain reload without animations, adjusting datasource a bit by not removing all content before search response arrives and getting rid of flicker which is present in the current release:
current

 So the result of changes is:
anim


Elland:
👏🏽

Toshi/Controllers/Dapps/DappsDataSource.swift
```diff
@@ -280,11 +282,22 @@ final class DappsDataSource {
...
+
         return DappsDataSourceSection(items: items)
     }
 
     func adjustToSearchText(_ searchText: String) {
-        content = [self.resultsForSearchText(searchText)]
+        guard let searchSection = resultsForSearchText(searchText) else { return }
+
+        // We should not remove previously shown results until search response comes, here we should change only first section if it is Search with Google and Url
+        if var existingSeardchSection = content.first, existingSeardchSection.categoryId == nil, existingSeardchSection.name == nil {
```
designatednerd: nit: `Seardch` needs the `d` removed

designatednerd:
still ✅ to me!

PR #790, opened 2018-03-21
Add Localized Plurals + Tests

designatednerd:
In this PR: 

- Support added for generating `LocalizedPlural` in the same method as `Localized`
- Replaced old stringly typed code with new hotness
- Added tests to ensure for every language we support, values exist for the regular and plural keys generated by Marathon, so that if someone without marathon installed deletes a key, our unit tests fail. This will also eventually allow us to ensure that there are, in fact, translations for all keys in all our supported languages. 
- Reorganized the Stencil script to try and make it clearer, started adding documentation to stuff I'm fairly confident isn't going to change. 

Question for the peanut gallery: 
- Should I break the plurals stuff out into a separate file? Might make it easier to combine some code around generating files if there have been changes in [place x], but I feel like all this crap should probably be in the same file.


Toshi/Generated/Scripts/StencilScript.swift
```diff
@@ -4,82 +4,158 @@ import PathKit
...
-try file.write(string: fileContents)
+let fileSystemLoader = FileSystemLoader(paths: [ Path(templatesFolder.path) ])
+let environment = Environment(loader: fileSystemLoader)
 
-print("Rendered \(file.name)")
+let localizableStringsChanged = try hasFileChanged(named: localizableFileName)
+let localizablePluralsChanged = try hasFileChanged(named: localizablePluralsFileName)
+
+if localizableStringsChanged || localizablePluralsChanged {
+    try writeLocalizableFile(withLocalized: try localizableStrings(from: baseLanguageFolder, fileName: localizableFileName),
```
marijnschilling: So you write both the plurals and the normal strings if one of them has changed?

marijnschilling: Ow wait it's the same file.

Tests/GeneratedCodeTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+class GeneratedCodeTests: XCTestCase {
+
+    /// An enum representing all currently supported languages.
+    /// Each key should be the name of the .lproj file so localized strings can be loaded
+    /// regardless of the language of the device or sim running the tests.
+    enum SupportedLanguages: String, StringCaseListable {
+        case
+        Base,
+        en // This is a workaround for a bug with StringCaseListable freaking out with single-case enums, which will (hopefully) go away in Swift 5
```
marijnschilling: He but `StringCaseListable` is an extension in Toshi right? So can't we fix it there? And otherwise i think the comment should be a line up. SInce `Base` is the weird one right?

Tests/GeneratedCodeTests.swift
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+    }
+
+    func testValuesExistForLocalizedPluralKeys() {
+        LocalizedPluralKey.allCases.forEach { key in
+            SupportedLanguages.allCases.forEach { language in
+                let format = language.localizedString(for: key.rawValue)
+                let zeroValue = String.localizedStringWithFormat(format, 0)
+                XCTAssertNotEqual(zeroValue, key.rawValue, "\(key.rawValue) does not have a value in the \(language.rawValue) Localizable.stringsdict for 0!")
```
marijnschilling: same question for this test

Tests/GeneratedCodeTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+                let oneValue = String.localizedStringWithFormat(format, 1)
+                XCTAssertNotEqual(oneValue, key.rawValue, "\(key.rawValue) does not have a value in the \(language.rawValue) Localizable.stringsdict for 1!")
```
marijnschilling: same question for this test

Tests/GeneratedCodeTests.swift
```diff
@@ -0,0 +1,81 @@
...
+        func localizedString(for key: String) -> String {
+            return bundle.localizedString(forKey: key, value: nil, table: nil)
+        }
+    }
+
+    func testValuesExistForAllLocalizedStringKeys() {
+        LocalizedKey.allCases.forEach { key in
+            SupportedLanguages.allCases.forEach { language in
+                let value = language.localizedString(for: key.rawValue)
+                XCTAssertNotEqual(value, key.rawValue, "\(key.rawValue) does not have a value in the \(language.rawValue) Localizable.strings!")
```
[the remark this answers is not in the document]
designatednerd: By default, if `NSLocalizedString` doesn't find a value for the provided key, it returns the key instead. This is why `snake_case_key_names` are pretty standard - because they make it REALLY easy to spot where a non-localized string has come from. 

Unless we have `snake_case_user_facing_text`, the key name and the text should never be the same, so this is why they should never be equal.

Toshi/Generated/Scripts/StencilScript.swift
```diff
@@ -4,82 +4,158 @@ import PathKit
...
+    let fileContents = try environment.renderTemplate(name: "LocalizedStrings.swift.stencil", context: context)
 
-let sortedKeys = localizableContents.keys.sorted()
-let localizedStrings: [LocalizedString] = sortedKeys.map { key in
-    let value = localizableContents[key]!
-    let valueWithoutNewlineCharacters = value.replacingOccurrences(of: "\n", with: "\\n")
-    return LocalizedString(key: key, value: valueWithoutNewlineCharacters)
+    let file = try codeFolder.createFileIfNeeded(withName: "LocalizedStrings.swift")
+    try file.write(string: fileContents)
+    print("Rendered \(file.name)")
```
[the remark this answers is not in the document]
designatednerd: I think so, it's helpful in terms of figuring out what's going on to know when the file actually changed or not.

Toshi/Generated/Scripts/StencilScript.swift
designatednerd: Yah, this is why I was considering breaking the plurals out into a separate file in my question in the PR description.

What do you think?

Tests/GeneratedCodeTests.swift
designatednerd: It is! However, I use this extension on several other projects and haven't figured out a good workaround to this quite yet.

Tests/GeneratedCodeTests.swift
designatednerd: Same answer 😄

Tests/GeneratedCodeTests.swift
designatednerd: Same answer 😄 😄

Toshi/Generated/Scripts/StencilScript.swift
marijnschilling: I think separate is good :) Localized strings is such a mess anyway 🙈

Tests/GeneratedCodeTests.swift
marijnschilling: Hmmm, oke! And what about putting the comment on 26?

Im curious to what the problem is with the base case :)

designatednerd: it's not the base case per se, it's that `StringCaseListable` gets caught in an infinite loop if there's only one case, since the next case is always....the current case.

designatednerd:
[hold off on merging, I'm gonna move localized plurals to a different file]

designatednerd:
@marijnschilling better?

designatednerd:
got the 👍  from @marijnschilling via slack, merging

PR #792, opened 2018-03-22
[IOS-563] Polish Dapp Profile

marijnschilling:
img_019914a57f9e-1

- [x] When there's no header image we show a smaller header
- [x] Set the right font and spacing for the Dapp title
- [x] Set line spacing on description
- [x] I also stuck the bottom insets to the tab bar.

https://trello.com/c/mzq4eObx/36-dapp-profile-polishing
https://toshiapp.atlassian.net/browse/IOS-562


Toshi/Controllers/Dapps/Views/DappInfoView.swift
```diff
@@ -138,7 +139,10 @@ final class DappInfoView: UIView {
         mainStackView.left(to: self, offset: .defaultMargin)
         mainStackView.right(to: self, offset: -.defaultMargin)
 
+        titleLabel.height(24)
```
yuliaveres: why do you constraint height for the label with Dynamic font ?

Toshi/Controllers/Dapps/DappViewController.swift
```diff
@@ -109,7 +110,11 @@ final class DappViewController: DisappearingNavBarViewController {
         guard let path = dapp.coverUrlString else { return }
```
yuliaveres: setup no cover image here as well please

Toshi/Controllers/Dapps/DappViewController.swift
```diff
@@ -18,7 +18,8 @@ import UIKit
 import SweetUIKit
 
 final class DappViewController: DisappearingNavBarViewController {
-    let dappConverHeaderHeight: CGFloat = 200
+    let dappCoverHeaderHeight: CGFloat = 200
+    let dappNoCoverHeaderHeight: CGFloat = 44
```
yuliaveres: this height is not enough, what about status bar ?

yuliaveres: img_0389

marijnschilling: add the safearea to it

PR #795, opened 2018-03-26
Add auto-generated Image Assets`

designatednerd:
In this PR: 
- Uses Marathon + Stencil to automatically generate `AssetCatalog.swift`, which contains keys based on the asset catalog item names and `ImageAsset` convenience methods which return a `UIImage` based on the key provided. 
- Updated everything in code using either `UIImage(named:)` or `#imageLiteral` to use `ImageAsset` instead. 
- Added test to validate that images exist for all keys to prevent someone who doesn't have marathon installed from deleting an image we still need. 

NOT in this PR but coming soon to a theater near you: 
- Going through all these asset names and figuring out which ones we still use so we can try to get rid of ones we're done with


designatednerd:
NOTE: Don't merge this until it gets repointed at `master`, this is presently dependent on #790, so I've pointed it at that branch.

designatednerd:
And now pointed at `master` 😄

Toshi/MessagesUI/Views/MessagesCornerView.swift
```diff
@@ -74,3 +70,10 @@ class MessagesCornerView: UIImageView {
         }
     }
 }
+
+private extension UIImage {
```
marijnschilling: I think we can better put this in its own file in the Extensions folder right? Or do we really never need this anywhere else?

Tests/GeneratedCodeTests.swift
```diff
@@ -78,4 +78,10 @@ class GeneratedCodeTests: XCTestCase {
             }
         }
     }
+
+    func testImagesExistForAssetCatalogItems() {
+        AssetCatalogItem.allCases.forEach { item in
+            XCTAssertNotNil(UIImage(named: item.rawValue), "No image for asset catalog item \(item.rawValue)")
```
marijnschilling: maybe we can make this messages clearer, since this test only fails when the user doesnt have marathon setup right? Would be cool if we could make clear how this failing test should be solved. Same for the other assertion descriptions for the localizable strings. I can imagine really waisting time on figuring this out :)

designatednerd: Good point - I'll update these failures to make what to do if they happen clearer.

Toshi/MessagesUI/Views/MessagesCornerView.swift
designatednerd: This was previously just a method within the view controller - it's really only used here, and I think it's heavily tied to the actual asset for the message text bubbles.

marijnschilling: Ah oke, I think we can leave it here :)

designatednerd:
what-you-did-there-i-see-it thumbnail

PR #796, opened 2018-03-26
[IOS-565] polish search dapps

marijnschilling:
Solves these issues, except "Semibold" title in search cell. We'll need to look at this in the bigger picture of making cells more flexible 🧘‍♂️
https://trello.com/c/AQxRmC6H/38-search-polishing
https://trello.com/c/q2m6PrQt/39-search-polishing
screenshot 2018-03-26 13 03 43

screenshot 2018-03-26 13 03 53


Toshi/Controllers/Dapps/DappsViewController.swift
```diff
@@ -262,24 +262,22 @@ extension DappsViewController: UITableViewDataSource {
...
             cell.textLabel?.textColor = Theme.tintColor
-            cell.separatorInset = .zero
             return cell
         case .searchWithGoogle:
             let cell = tableView.dequeueReusableCell(withIdentifier: genericCellReuseIdentifier, for: indexPath)
 
             cell.selectionStyle = .none
 
-            let googleText = " - \(Localized.dapps_search_with_google_section_title)"
+            let googleText = " – \(Localized.dapps_search_with_google_section_title)"
```
designatednerd: what...actually changed here?

Toshi/Views/Cells/TitleSubtitleSwitchCell.swift
```diff
@@ -58,7 +58,7 @@ final class TitleSubtitleSwitchCell: BasicTableViewCell {
     override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
 
-        titleTextField.font = Theme.preferredRegular()
-        subtitleLabel.font = Theme.preferredRegularSmall()
+        titleTextField.font = BasicTableViewCell.titleFont
+        subtitleLabel.font = BasicTableViewCell.subtitleFont
```
designatednerd: It looks like you're overriding and changing `traitCollectionDidChange` in multiple different places - would it be better to do this once in the `BasicTableViewCell`?

Toshi/Views/TintColorChangingButton.swift
```diff
@@ -15,7 +15,7 @@
 
 import UIKit
 
-/// A button which automatically updates its tint color based on the button's current state. 
+/// A button which automatically updates its tint color based on the button's current state.
```
designatednerd: sorry sorry sorry sorry

Toshi/Controllers/Dapps/DappsViewController.swift
marijnschilling: it's a slightly different character, a bit longer stripe :) I was actually not sure myself wether it changed or not 😂  But git thinks so, and Marek will probably be pleased as well.

Toshi/Views/Cells/AvatarTitleSubtitleDetailsCell.swift
```diff
@@ -69,8 +69,8 @@ final class AvatarTitleSubtitleDetailsCell: BasicTableViewCell {
     override open func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
 
-        titleTextField.font = Theme.preferredRegular()
-        subtitleLabel.font = Theme.preferredRegularSmall()
-        detailsLabel.font = detailsFont
+        titleTextField.font = BasicTableViewCell.titleFont
```
yuliaveres: in each case here - would it actually be different ? since those values are `let` constants on `BasicTableViewCell` 🤔

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -50,16 +50,18 @@ class BasicTableViewCell: UITableViewCell {
     static let largeVerticalMargin: CGFloat = 22.0
     static let badgeViewSize: CGFloat = 24.0
 
-    weak var actionDelegate: BasicCellActionDelegate?
+    static let titleFont = Theme.preferredSemibold()
```
yuliaveres: dynamic text size will work if we change this to var probably ? ..

marijnschilling: But why would we? I think we should keep this a static var, in this structure we have with all the different cell components. if we make this a var it's just an open invitation to make this in to a big mess. 

the whole point with these cells is to keep things easily consistent over the app right?

Toshi/Views/Cells/AvatarTitleSubtitleDetailsCell.swift
marijnschilling: You have to re-set it when the trait collection changes, to support dynamic type. I did it all once in the basicCell now ✌️

Toshi/Views/Cells/TitleSubtitleSwitchCell.swift
marijnschilling: I think the downside of this is that since it are lazy vars, even if you don't have a subtitleLabel it will get initialised because of this function.

But i think we can live with that right?

Toshi/Views/Cells/BasicTableViewCell.swift
designatednerd: I don't think whether it's a `var` or a `let` makes the difference, since theoretically the type being returned is dynamic - @marijnschilling  you have a way better understanding of this, am I correct on that? 

If I *am* correct, do we actually need to do anything in `traitCollectionDidChange`, though?

marijnschilling: Hmm interstingly not overriding the traitCollection does work! I remember having to add this to support dynamic type. But not anymore ¯\_(ツ)_/¯

I will remove it and the static font types.

designatednerd:
My only thought on the trait collection thing - maybe it was necessary for iOS 10? we should see where we're at on being able to drop that 🤔

marijnschilling:
I tried iOS 10 works like a charm 🧞‍♀️

Toshi/Views/Cells/BasicTableViewCell.swift
yuliaveres: it does make difference. as soon as text size is change - the app should react immediately and adjust text size. with let it must be changed on relaunch only

yuliaveres: @marijnschilling static as it is, but **var**

designatednerd: The `let` used a font which was set up using dynamic type though - it should still adjust.

marijnschilling: I already removed this since we do not need this anymore. It's how it was before now ✌️

marijnschilling: it does! It just happens automatically ✌️

yuliaveres: how ?

yuliaveres: that is how it used to change:
img_0397

how it is on the branch:
img_0399

marijnschilling: by setting `adjustsFontForContentSizeCategory` on the label, and by using a metric font. The need to override traitCollectionDidChange was only to fix a temporary bug I guess

designatednerd: So @yuliaveres @marijnschilling it looks like you're both getting different behavior with the same code. 

Might be worth each of you trying on different sims/devices and/or doing a clean to see if there's some issue in your local causing this

yuliaveres: all places should be tested 🤷‍♂️

marijnschilling: Apparantly i tested with cells that weren't the ToshiBasicTableViewCell. But are just UITbaleViewCells, thats where it only works automatically.

marijnschilling: Should be fixed btw, doesnt make sense to have a TableViewCell system set up and not use it. That is very confusing.

PR #797, opened 2018-03-26
Remove Unused Assets + Localized Strings

designatednerd:
Well this is a lot easier with the generated code. 

This PR: 
- Removes a whole mess of assets and localized strings we're not using anymore
- Regenerates the appropriate files with marathon
- Catches a couple things that had previously gotten away from me.


designatednerd:
went through and validated that the various keys were not used anywhere in the application, then double-checked that the other methods we were using before were really, really, really gone (that's where I found a couple stragglers). 

In *theory* we should be able to do this based on whether something is called or not using the `Callers` section of the `Related Items` drop down, but in practice my indexing is hopelessly broken so there was a lot of searching for the various keys manually 😛

PR #798, opened 2018-03-26
[IOS-566] Polish dapp browser

marijnschilling:
Adjusted sizing in the toolbar:
screenshot 2018-03-26 16 00 31

Cannot reproduce the back-button issue.

https://trello.com/c/ts88FZlU/40-browser-polishing


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -186,7 +186,7 @@ final class SOFAWebController: UIViewController {
         stackView.distribution = .fillProportionally
 
         view.addSubview(stackView)
-        stackView.edgesToSuperview()
+        stackView.edges(to: view, insets: UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 0))
```
designatednerd: In theory both of these things should do the same thing, since `view` is `stackView`'s superview, and by default `edgesToSuperview` should use zero insets. What's the practical difference here?

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -224,7 +224,7 @@ final class SOFAWebController: UIViewController {
         toolbar.top(to: layoutGuide())
         toolbar.left(to: view)
         toolbar.right(to: view)
-        toolbar.height(.defaultBarHeight)
+        toolbar.height(56)
```
designatednerd: Won't this make the bottom bit super-huge if you're not on a x?

Toshi/SOFAWebview/SOFAWebController.swift
marijnschilling: what did i not push? Letmecheck :) Doing 8 things at the same time atm

designatednerd:
@marijnschilling I fixed the back button issue in #794.

Toshi/SOFAWebview/SOFAWebController.swift
marijnschilling: No because the stack view centers it :)

designatednerd: "it" being the contents of the toolbar here?

designatednerd: @marijnschilling  ^^

marijnschilling: yes thats it!

yuliaveres:
the back button issue was fixed few days ago by @designatednerd  👍

PR #799, opened 2018-03-26
[IOS-567, IOS-568] Fix Ethereum address validation and input field resizing

yuliaveres:
Address field ***not resized*** on Send token view + address ***validation*** does not work properly as not the whole string is taken into consideration:
img_0123

***Fixed:***
img_0124
img_0125


Toshi/Models/EthereumAddress.swift
```diff
@@ -40,7 +40,7 @@ struct EthereumAddress {
 
     static func validate(_ address: String) -> Bool {
         do {
-            let regex = try NSRegularExpression(pattern: "0x[a-fA-F0-9]{40}")
+            let regex = try NSRegularExpression(pattern: "^0x[a-fA-F0-9]{40}$")
```
designatednerd: Out of curiosity: What do these changes to the Regex do to make this work better?

Tests/EthereumAddressTests.swift
```diff
@@ -35,4 +35,17 @@ class EthereumAddressTests: XCTestCase {
...
+        XCTAssertTrue(EthereumAddress.validate(canonicalAddress))
+
+        let tooLongAddress = canonicalAddress + "j89"
+        XCTAssertFalse(EthereumAddress.validate(tooLongAddress))
+
+        let tooShortAddress = String(canonicalAddress.prefix(10))
+        XCTAssertFalse(EthereumAddress.validate(tooShortAddress))
+
+        let nonHexAddress = String(canonicalAddress.suffix(canonicalAddress.count - 3))
+        XCTAssertFalse(EthereumAddress.validate(nonHexAddress))
```
designatednerd: are there examples of valid addresses which were previously failing which we should be testing here?

Toshi/Models/EthereumAddress.swift
yuliaveres: so the regex checks for the whole string, not just finds valid substring and all good

Tests/EthereumAddressTests.swift
yuliaveres: no, invalid addresses were succeeding

PR #801, opened 2018-03-29
Robot test sign in screen 🤖

marijnschilling:
First stab at robot testing the sign in screen!

- [ ] The `howDoesItWork` test does not succeed. I don't understand why. But it fails on `.validateOffHowDoesItWorkScreen() ` which checks wether the view with the title 'how does it work' is still on screen (which it shouldn't be after we pressed the bak button) Weirdddd! (hence my comment) I got a bit stuck with that and wonder if you have any leads on how to make it work.
- [ ] I'm very open to suggestions on other test scenario's that need to be covered. I kinda tested what I thought was necessary but there are many many ways to go from here!

note: I personally don't prefer putting function signatures on multiple lines. One big reason for that is just that im not used to it. But another is maybe that while coding i make a lot of use of the commands that can remove a single line, or move a single line up or down. And another reason is that Swift lint is also complaining about it, thats why i changed it in some spots. but we should reach consensus about it, i f we do want to do this we should mute swift line (and Marijn) 🤐


Tests_UI/Robots/BasicRobot.swift
```diff
@@ -45,27 +45,39 @@ protocol BasicRobot {
     func confirmViewVisibleWith(accessibilityIdentifier: AccessibilityIdentifier,
                                 file: StaticString,
                                 line: UInt)
+
+    /// Runs an action where the implementation's test framework validates that a view is presently visible based on its placeHolder.
```
designatednerd: "validates that a view" - I think this should probably be updated to "a UITextField" since it's looking specifically for that.

Tests_UI/Robots/BasicRobot.swift
```diff
@@ -45,27 +45,39 @@ protocol BasicRobot {
...
+    ///
+    /// - Parameters:
+    ///   - placeHolder: The placeHolder of the view to look for.
```
designatednerd: same here re view -> UITextField

Tests_UI/Robots/BasicRobot.swift
```diff
@@ -95,6 +107,41 @@ protocol BasicRobot {
     func tapCellWith(accessibilityLabel: String,
                      file: StaticString,
                      line: UInt)
+
+    /// Runs an action where the implementation's test framework types the input text on the view based on its accessibility label.
```
designatednerd: accessibility label or accessibility identifier? docs say one thing, param says something else

Tests_UI/Robots/BasicRobot.swift
```diff
@@ -95,6 +107,41 @@ protocol BasicRobot {
...
+                  onViewWith accessibilityIdentifier: String,
+                  file: StaticString, line: UInt)
+
+    /// Runs an action where the implementation's test framework clears the input text on the text field based on its placeholder.
+    ///
+    /// - Parameters:
+    ///   - placeholder: The placeholder of the text field to clear the text of.
+    ///   - file: The file from which this method is being called.
+    ///   - line: The line from which this method is being called.
+    func clearText(onViewWith placeholder: String, file: StaticString, line: UInt)
```
designatednerd: `View` -> `TextField` for same reasons as above, i think

Tests_UI/Robots/BasicRobot.swift
```diff
@@ -95,6 +107,41 @@ protocol BasicRobot {
...
+    ///   - line: The line from which this method is being called.
+    func confirmButtonEnabled(accessibilityLabel: String, file: StaticString, line: UInt)
+
+    /// Runs an action where the implementation's test framework validates that a button is disabled based on its accessibility identifier.
+    ///
+    /// - Parameters:
+    ///   - accessibilityLabel: The Accessibility Label of the button. This would be read out loud to VoiceOver users.
+    ///   - file: The file from which this method is being called.
+    ///   - line: The line from which this method is being called.
+    func confirmButtonDisabled(accessibilityLabel: String, file: StaticString, line: UInt)
```
designatednerd: would it make sense to combine these into a single `confirmEnabled(_ shouldBeEnabled: Bool, ...` method?

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
```diff
@@ -0,0 +1,72 @@
...
+                      file: file,
+                      line: line)
+        
+        return self
+    }
+    
+    // MARK: - Validators
+    
+    @discardableResult
+    func validateOnHowDoesItWorkScreen(file: StaticString = #file, line: UInt = #line) -> HowDoesItWorkScreenRobot {
```
designatednerd: Couple of thoughts here: 
- Since there's only one button, I think the enum is overkill - just do "tap back" as the name, since that's really the only thing you can do here. 
- Since this screen IS so simple (and probably will remain very simple), while conceptually you usually want one robot per screen, I'd be OK with dropping this back down to the `SignInRobot` so you don't have to do a whole other robot.  

Also, I spotted an X bug on the layout when I was investigating this: 
simulator screen shot - iphone x - 2018-03-29 at 11 49 54

Tests_UI/Robots/SignInRobot.swift
```diff
@@ -43,12 +43,25 @@ enum SignInScreenButton {
     }
 }
 
+enum SingInScreenLabel {
```
designatednerd: `Sing` -> `Sign`. Although I've seen this typo enough I think building a "Sing in screen" might be a fun idea for Stupid Hackathon

Tests_UI/Robots/SignInRobot.swift
```diff
@@ -80,4 +93,76 @@ extension SignInRobot {
...
+
+        return self
+    }
+
+    // MARK: - Typing
+
+    @discardableResult
+    func enterValidPassPhraseWords(amount: Int, file: StaticString = #file, line: UInt = #line) -> SignInRobot {
+
+        typeText(String(repeating: "ask ", count: amount), onViewWith: "login_textField",
```
designatednerd: Here and in the other methods, the string in `onViewWith` should probably be something in the `AccessibilityIdentifier` rather than a hard-coded string.

Toshi/Controllers/Sign In/SignInView.swift
```diff
@@ -50,6 +50,7 @@ final class SignInView: UIView {
         view.textColor = Theme.darkTextColor
         view.returnKeyType = .next
         view.alpha = 0
+        view.accessibilityIdentifier = "login_textField"
```
designatednerd: here's the other place you'd use the `AccessibilityIdentifier` instead of a hard-coded string

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
```diff
@@ -0,0 +1,72 @@
...
+        confirmViewVisibleWith(accessibilityLabel: Localized.passphrase_sign_in_explanation_title,
+                               file: file,
+                               line: line)
+        return self
+    }
+    
+    @discardableResult
+    func validateOffHowDoesItWorkScreen(file: StaticString = #file, line: UInt = #line) -> HowDoesItWorkScreenRobot {
+        confirmViewGoneWith(accessibilityLabel: Localized.passphrase_sign_in_explanation_title,
```
designatednerd: so i think the reason this is failing is that the title of the button you're clicking is identical to this title - they're both "How does it work?" - So there's still a view with that accessibility label on screen, even though it's not the same view. 

Maybe look for the actual explanation text instead

designatednerd:
Looking awesome btw 💪

Tests_UI/Robots/BasicRobot.swift
marijnschilling: *Must remember to update code comments* 😬

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
marijnschilling: Funny thing, when i change the `validateOffHowDoesItWorkScreen` to check for the text it works, but for the `validateOnHowDoesItWorkScreen` it fails now :( 

While i do that check after selecting the `How does it work` button and validating to be off the sing in screen. ✌️

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
marijnschilling: Funny thing, when i change the validateOffHowDoesItWorkScreen to check for the text it works, but for the validateOnHowDoesItWorkScreen it fails now :(

While i do that check after selecting the How does it work button and validating to be off the sing in screen. ✌️

So problem solved, but other problem introduced.

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
marijnschilling: but... consistency! I would just make a different robot for it -just because- it would make it clear where you can find things, since all the other screens will be their own Robot :shrug:

marijnschilling: So you mean... Here i shouldn't use the enum but a hardcoded string?

designatednerd: fair enough re: consistency. 

And I meant use `Localized.back_action_title` directly instead of `Enum.caseName.accessibilityLabel`

and spoiler alert, my answer will almost never be "use a hardcoded string" 😄

marijnschilling: haha i thought so! Thats why i asked :D

marijnschilling: I don't get though what you mean, so no enum at all? Just  make the whole method `selectBackButton`?

designatednerd:
@marijnschilling yes since that’s the only thing you can do here (and probably will be for the foreseeable future) 

`func selectBackButton()`, but with the same file and line stuff as other things. 

The enum helpers with the options for what buttons you can select wind up just being noise when there’s only one thing to do which can be encapsulated in the function name, IMO.

marijnschilling:
Nothing runs anymore, but deleting the file doesn't solve it.

marijnschilling:
I think i have to revert a few commits and start all over. Strange things are happening.

marijnschilling:
I found the problem 😬 its because the textLabel on 'How does it work' is so long, the accesibilityLabel is only a piece of the text 🙇 I will just add an identifier to it 😞 But maybe in the future it could be interesting to write some earl grey stuff that checks wether the checked for accessibilityLabel containsss one of the accesibilitylabels of the views on screen.. 🤔  Shouldnt be too hard.

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
designatednerd: [main comments addressed, I'll take a look in a bit at the layout]

marijnschilling: I didnt fix the layout, since its not related to this PR :) We should fix it in another PR though

Tests_UI/Robots/EarlGreyRobot.swift
```diff
@@ -131,3 +182,23 @@ extension EarlGreyRobot: MyProfileRobot { /*mix-in */ }
...
+extension EarlGreyRobot: HowDoesItWorkScreenRobot { /* mix-in */ }
+
+/**
+*  Matcher for UI element with the provided accessibility @c label.
+*
+*  @param label The accessibility label to be matched.
+*
+*  @return A matcher for the accessibility label of an accessible element.
+*/
+func matcher(forPlaceholder placeholder: String) -> GREYMatcher {
```
designatednerd: style thing: maybe `matcher(for: placeholder)`, but that might be personal preference.

Tests_UI/Robots/HowDoesItWorkScreenRobot.swift
designatednerd: ^^ let me know what ya think on this

PR #803, opened 2018-04-03
[IOS-583]: Thread muting

yuliaveres:
Now there is option to mute or unmute a thread. As soon as app goes to background we close socket in order to increase probability of the app being launched in background by the system if a remote notification is received. (According to the documentation, there is no guarantee that app will be launched, it is up to the system and how much of energy is used for the work which app does in background)


Toshi/Models/MessageFetcherJob.swift
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            if more {
+                return strongSelf.run()
+            }
+        }
+    }
+
+    private func parseMessagesResponse(responseObject: Any?) -> (envelopes: [OWSSignalServiceProtosEnvelope], more: Bool)? {
+        guard let responseObject = responseObject else {
```
Elland: Why 3 guards instead of one? Readability?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -227,8 +232,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, didReceiveRemoteNotification userInfo: [AnyHashable: Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
 
-        Navigator.tabbarController?.triggerWalletTabReloadIfNeeded(basedOn: userInfo)
-        completionHandler(.noData)
+        // We need to make sure completion handler is called within 30 seconds after notification is received
+        DispatchQueue.main.asyncAfter(seconds: 20, execute: {
```
Elland: This feels weird. Why an arbitrary timeout, instead of running the fetcher with a completion handler?

You can make the fetching quicker by bailing out early if there is no connection or there are no new messages. If the user has a connection, it should take maybe 1-2s to run, not 20. By forcing 20s all the time, you're signalling to the system that this app is always performing long operations, which might downgrade us to being called less and less often. 🤔

yuliaveres: yeah i know, but i do not have callback when the envelope is processed and local notif is about to be shown, i will see now what we can do. signal were doing exactly same.. i do not really like this though.. let me see

Toshi/Models/MessageFetcherJob.swift
```diff
@@ -15,6 +15,18 @@
 
 import Foundation
 
+enum MessageResponseKeys {
```
Elland: Any reason to declare this in global scope, instead of inside the fetcher class?

PR #805, opened 2018-04-05
[IOS-570, IOS-590, IOS-591]: Threads datasource updates

yuliaveres:
Now we use threads datasource for both main chats page and unaccepted chats page (Messages requests).

img_0132
img_0126
img_0130
img_0133
img_0128
img_0129


Toshi/Controllers/Recent/ChatsNavigationController.swift
```diff
@@ -15,10 +15,10 @@
 
 import UIKit
 
-class RecentNavigationController: UINavigationController {
+final class ChatsNavigationController: UINavigationController {
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
-        return .default
+        return (topViewController == viewControllers.first) ? .lightContent : .default
```
marijnschilling: You can add a let to make clear what this is testing:
```
let isShowingChatsViewController = topViewController == viewControllers.first
return isShowingChatsViewController ? .lightContent : .default
```

Toshi/Controllers/Recent/ChatsNavigationController.swift
```diff
@@ -28,49 +28,19 @@ class RecentNavigationController: UINavigationController {
     override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
         super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
 
-        tabBarItem = UITabBarItem(title: Localized.tab_bar_title_recent, image: ImageAsset.tab2, tag: 0)
+        tabBarItem = UITabBarItem(title: Localized.tab_bar_title_chats, image: ImageAsset.tab2, tag: 0)
         tabBarItem.titlePositionAdjustment.vertical = TabBarItemTitleOffset
     }
 
-    lazy var backgroundBlur: BlurView = {
```
marijnschilling: Oh wow, if we remove the back ground blur that means we can have a lot simpler setup of tableViews right? Since we dont need the content to stretch underneath the nav bar?

Toshi/Controllers/Recent/ChatsNavigationController.swift
```diff
@@ -119,4 +89,43 @@ class RecentNavigationController: UINavigationController {
 
         return super.popToViewController(viewController, animated: animated)
     }
+
+    override func popViewController(animated: Bool) -> UIViewController? {
+         UserDefaultsWrapper.selectedThreadAddress = nil
+
+        guard let colorChangingVC = previousViewController as? NavBarColorChanging else {
+            // Just call super and be done with it.
```
marijnschilling: Could you add to you comments an explanation of in which case we don't have to do what?

Toshi/Controllers/Recent/ChatsNavigationController.swift
```diff
@@ -119,4 +89,43 @@ class RecentNavigationController: UINavigationController {
...
+    }
+
+    private var previousViewController: UIViewController? {
+        guard viewControllers.count > 1 else {
+            return nil
+        }
+        return viewControllers[viewControllers.count - 2]
+    }
+
+    private func setNavigationBarColors(with colorChangingObject: NavBarColorChanging) {
```
marijnschilling: Is this to have the green navigation bar fade to the default color navigation bar when a controller is pushed?

Toshi/Controllers/Recent/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+import SweetFoundation
+import SweetUIKit
+
+final class ChatsViewController: SweetTableController {
+
+    private lazy var dataSource: ThreadsDataSource = {
```
marijnschilling: Why is it called threads? Isn't `chats` or `conversations` a better term? threads makes me immediately think of background threads and those kind of threads.

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -15,39 +15,95 @@
 
 import UIKit
 
-enum ThreadsContentSection: Int {
```
marijnschilling: Hmmm i think this file isnt added in the project file system, i cannot find it.

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -15,39 +15,95 @@
...
+            return Localized.recent_invite_a_friend
+        default:
+            return nil
+        }
+    }
+
+    var icon: UIImage? {
+        switch self {
+        case .inviteFriend:
+            return #imageLiteral(resourceName: "invite_friend")
```
marijnschilling: We have Ellen's new way of addressing images right, that one is merged
`ImageAsset.invite_friend` easy! And then its also not optional! 🎉 

https://github.com/toshiapp/toshi-ios-client/pull/795

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -75,18 +134,32 @@ final class ThreadsDataSource: NSObject {
         return Int(viewModel.acceptedThreadsMappings.numberOfItems(inSection: UInt(0)))
     }
 
+    var numberOfSections: Int {
```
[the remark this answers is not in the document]
marijnschilling: Yeah or even make three different static lets for the SectionAmounts 
```
static let numberOfSectionsForMainPage = 2
static let numberOfSectionsForMainPageWithMessageRequests = 2
static let numberOfSectionsForMessageRequests = 1
```
haha so long though! But a lot clearer in my opinion.

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -15,39 +15,95 @@
...
-            return Localized.messages_requests_title
+            return Theme.navigationBarColor
+        }
+    }
+
+    var navTitleColor: UIColor {
+        switch self {
+        case .chatsMainPage:
+            return Theme.lightTextColor
+        case .unacceptedThreadRequests:
```
marijnschilling: I would name this case equal to the actual titles of the view, so call it `messageRequests` instead of unnacceptedThread. Thats a lot easier for other devs and yourself in the future to figure out what this corresponds with.

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -15,39 +15,95 @@
...
-        case .acceptedThreads:
-            return Localized.recent_messages_section_header_title
+    var prefersLargeTitle: Bool {
+        return self == .unacceptedThreadRequests
+    }
+}
+
+enum ChatsSectionType {
+    case findPeople
+    case messagesRequests
```
marijnschilling: This can be called `messageRequests` (since the requests is indicating plurality)

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -75,18 +134,32 @@ final class ThreadsDataSource: NSObject {
         return Int(viewModel.acceptedThreadsMappings.numberOfItems(inSection: UInt(0)))
     }
 
+    var numberOfSections: Int {
+        switch target {
```
marijnschilling: Are the databases of the "main" chats and message requests really that inextricably linked?

I would higly recommend making two different datasource files for this, one for the Chat and one for the Message Requests. that will make the files way easier to understand and way more flexible to change. imagine we want to remove the messageRequests at some point when we have separate datasources this will we a matter of deleting one file 😗🎶 while now we will have to carefully dissect the two files, which will take a lot of time and is higlhy prone to errors! 

I think at this point it will be still quite easy to split them, the longer we wait the more entangled it will get. It's the same probably i have just ran into with the DappsDataSource taking care of as well the DappsFrontPage` tabelView as the `DappsSearchResultTableView`, we'll have to separate those to keep the code flexible and maintainable.

marijnschilling:
Ow I wanted to comment with my change request :) 

I mostly looked at the DataSource for now, I think it's good how you make the different ways of how it's applied in different context clear. But my main concern is the long comment I placed where I think it might even make it more clear when we just create two different data sources. Let me know what you think!

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
```diff
@@ -1,166 +0,0 @@
...
-// GNU General Public License for more details.
-//
-// You should have received a copy of the GNU General Public License
-// along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-import UIKit
-import SweetFoundation
-import SweetUIKit
-
-final class MessagesRequestsViewController: SweetTableController {
```
marijnschilling: Where did this controller go?

Toshi/Controllers/Recent/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        preferLargeTitleIfPossible(target.prefersLargeTitle)
+        tabBarController?.tabBar.isHidden = false
+
+        tableView.reloadData()
+
+        dismissIfNeeded(animated: false)
```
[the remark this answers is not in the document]
yuliaveres: the flow is -> Chats -> MessagesRequests -> Accept / Decline - at this point not a request anymore. If you approve / decline all requests and go back - there are no more requests, so we just need to go straight to the Chats

Toshi/Controllers/Recent/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        tableView.delegate = self
+
+        BasicTableViewCell.register(in: tableView)
+
+        tableView.tableFooterView = UIView(frame: .zero)
```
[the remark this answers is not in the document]
yuliaveres: yes, will do, that is for avoiding empty cells in the bottom of the tableView.

Toshi/Controllers/Recent/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+        tableView.reloadData()
+        dismissIfNeeded()
+    }
+}
+
+// MARK: - Profiles List Completion Output
+
+extension ChatsViewController: ProfilesListCompletionOutput {
+
+    func didFinish(_ controller: ProfilesViewController, selectedProfilesIds: [String]) {
```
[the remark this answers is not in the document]
yuliaveres: Profiles controller in a selection mode passes selected results to its delegate

Toshi/Controllers/Recent/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+        case .chat:
+            showThread(at: indexPath)
+        case .inviteFriend:
+            shareWithSystemSheet(item: Localized.sharing_action_item)
+        }
+    }
+
+    func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+
+        guard indexPath.section < dataSource.sections.count else { return false }
```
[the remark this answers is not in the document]
yuliaveres: in datasource and table view reload race conditions 😄

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -277,12 +350,179 @@ final class ThreadsDataSource: NSObject {
 
     private func loadMessages() {
         viewModel.uiDatabaseConnection.asyncRead { [weak self] transaction in
+
+            guard let strongSelf = self else { return }
+
             self?.viewModel.acceptedThreadsMappings.update(with: transaction)
```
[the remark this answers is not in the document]
yuliaveres: will double check

Toshi/Controllers/MessageRequests/MessageRequestsViewController.swift
yuliaveres: we use chats controller as we display same thing - threads. but unaccepted

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -277,12 +350,179 @@ final class ThreadsDataSource: NSObject {
...
+            (cell as? BasicTableViewCell)?.titleTextField.textColor = Theme.tintColor
+            configurator.configureCell(cell, with: cellData)
+        case .messageRequests:
+            cell = messagesRequestsCell(for: indexPath)
+            cell.accessoryType = .disclosureIndicator
+        case .chat:
+            if let thread = acceptedThread(at: indexPath.row, in: 0) {
+                let threadCellConfigurator = ThreadCellConfigurator(thread: thread)
+                let cellData = threadCellConfigurator.cellData
+                cell = tableView.dequeueReusableCell(withIdentifier: AvatarTitleSubtitleDetailsBadgeCell.reuseIdentifier, for: indexPath)
```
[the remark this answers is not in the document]
yuliaveres: As we have the whole system of cells we use in Toshi with the abstract cell class and all subclasses which differ in appearance, we can't use *SweetUIKit* method which refers to cell class only, we must use reuseIdentifier explicitly.

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -277,12 +350,179 @@ final class ThreadsDataSource: NSObject {
...
+        guard indexPath.row < section.items.count else { return UITableViewCell() }
+
+        let item = section.items[indexPath.row]
+
+        switch item {
+        case .findPeople, .inviteFriend:
+            let configurator = CellConfigurator()
+            let cellData = TableCellData(title: item.title, leftImage: item.icon)
+            cell = tableView.dequeueReusableCell(withIdentifier: configurator.cellIdentifier(for: cellData.components), for: indexPath)
+            (cell as? BasicTableViewCell)?.titleTextField.textColor = Theme.tintColor
```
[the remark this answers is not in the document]
yuliaveres: no I just need to access BasicTableviewCell property here 🙂

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -277,12 +350,179 @@ final class ThreadsDataSource: NSObject {
...
+        }
+
+        guard let cell = tableView?.dequeueReusableCell(withIdentifier: cellConfigurator.cellIdentifier(for: cellData.components), for: indexPath) else { return UITableViewCell() }
+        cellConfigurator.configureCell(cell, with: cellData)
+        cell.accessoryType = accessoryType
+
+        return cell
+    }
+
+    private func messageRequestsCell(for indexPath: IndexPath, tableView: UITableView) -> UITableViewCell {
```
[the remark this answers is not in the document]
yuliaveres: i need to rename one 🙈

Toshi/Controllers/Recent/ChatsViewController.swift
yuliaveres: i can name it `chats`, okay...

Toshi/Controllers/Recent/ThreadsDataSource.swift
yuliaveres: we talked about it- it is there. will make some changes to folder name as well

Toshi/Controllers/Recent/ThreadsDataSource.swift
yuliaveres: yes, of course they are. all `TSThread` objects are stored in same collection. There is no purpose and it would be odd to have separate file of same code which turns on a filtering by one property `isAccepted` for the same collection of `TSThreads`. It is clear and easy to  initialize the datasource with target type `messageRequestsPage` if we want to have message requests only, and ⭐️ .

marijnschilling: I think it would be more logical to have the DataSource handle just what data should be displayed, maybe you van have another object that manages this collection of TSThreads. I'm not sure how to approach this exactly but I had a discussion about it with @designatednerd when we discussed how the dappsDataSource is set up. I really want to avoid a situation like the DappsDataSource where search and getting the dapps frontpage are combined. Since we now run into flexibility issues with that. I think it would be more logical to have it separated and to have the DataSource really provide the unique tableView with its necessary data.

Toshi/Controllers/Chats/ChatsNavigationController.swift
```diff
@@ -15,10 +15,11 @@
 
 import UIKit
 
-class RecentNavigationController: UINavigationController {
+final class ChatsNavigationController: UINavigationController {
```
marijnschilling: 🎉  Yeah!

Toshi/Controllers/Chats/ChatsNavigationController.swift
```diff
@@ -94,20 +65,20 @@ class RecentNavigationController: UINavigationController {
         self.pushViewController(chatViewController, animated: animated)
     }
 
+    // MARK: - Nav Bar Color Handling
+
     override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        if let colorable = viewController as? NavBarColorChanging {
```
marijnschilling: Can you rename this var to `colorChangingViewController` i think it's named like that in other usages, and it makes it a bit more clear.

Toshi/Controllers/Chats/ChatsNavigationController.swift
```diff
@@ -94,20 +65,20 @@ class RecentNavigationController: UINavigationController {
...
+            setNavigationBarColors(with: colorable)
```
marijnschilling: And from what i expect this function to do wouldnt it make more sense to call it: `setNavigationBarColors(on: colorable)`

Toshi/Controllers/Chats/ChatsNavigationController.swift
```diff
@@ -119,4 +90,43 @@ class RecentNavigationController: UINavigationController {
 
         return super.popToViewController(viewController, animated: animated)
     }
+
+    override func popViewController(animated: Bool) -> UIViewController? {
+         UserDefaultsWrapper.selectedThreadAddress = nil
+
+        guard let colorChangingVC = previousViewController as? NavBarColorChanging else {
```
marijnschilling: Could you also change this to `colorChangingViewController`? Not that important but nice to have no abbreviations for consistency 🤷‍♂️

Toshi/Controllers/Chats/ChatsNavigationController.swift
```diff
@@ -119,4 +90,43 @@ class RecentNavigationController: UINavigationController {
...
+            // Just call super and be done with it if previous controller does not handle color changes for navigation bar.
```
marijnschilling: I still don't really get what happens here. So you check wether the ViewController that is being was a color changing one, and if so you set the colors again?

Toshi/Controllers/Recent/ThreadsDataSource.swift
yuliaveres: i do not agree it would be somehow more logical and there is no need to have separate file apart from this which will just filter collection of Threads.

Toshi/Controllers/Chats/ChatsNavigationController.swift
```diff
@@ -119,4 +90,43 @@ class RecentNavigationController: UINavigationController {
...
+    }
+
+    private var previousViewController: UIViewController? {
+        guard viewControllers.count > 1 else {
+            return nil
+        }
+        return viewControllers[viewControllers.count - 2]
+    }
+
+    private func setNavigationBarColors(with colorChangingObject: NavBarColorChanging) {
```
marijnschilling: Or is this function setting the current navigation bar colors to the colors of the viewController given in the parameter? No right because the Colors are degined in the `NavBarColorChanign` protocol?

Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+
+        dismissIfNeeded(animated: false)
+        
+        navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .add, target: self, action: #selector(didPressCompose(_:)))
+    }
+
+    func dismissIfNeeded(animated: Bool = true) {
+        guard target == .messageRequestsPage else { return }
+
+        if Navigator.topNonModalViewController == self && dataSource.unacceptedThreadsCount == 0 {
```
marijnschilling: And what does this check check? maybe here is a similar thing possible to create a `let` first that explains what you are checking for ✌️

Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+        return dataSource
+    }()
+
+    private var chatAPIClient: ChatAPIClient {
+        return ChatAPIClient.shared
+    }
+
+    let idAPIClient = IDAPIClient.shared
+
+    private var target: ChatsDataSourceTarget = .chatsMainPage
```
marijnschilling: We can leave this value unset because you set it in the init, makes more sense ✌️

marijnschilling: And add some whitespace between 35 and 36 (🤓 )

Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+
+    func tableView(_: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
+        guard indexPath.section < dataSource.sections.count else { return }
+        let chatsPageSection = dataSource.sections[indexPath.section]
+        let item = chatsPageSection.items[indexPath.row]
+
+        switch item {
+        case .findPeople:
```
marijnschilling: What I don't really get is how this ViewController can be used for the .chatsMainPage and the .messageRequests page, doesn't the messageRequest page have whole other items on it? That is only messageRequests and not all the others?

Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+        case .chat:
+            showThread(at: indexPath)
+        case .inviteFriend:
+            shareWithSystemSheet(item: Localized.sharing_action_item)
+        }
+    }
+
+    func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+
+        guard indexPath.section < dataSource.sections.count else { return false }
```
marijnschilling: Maybe you could put this in a little helper method so you can just say 
```
guard let item = item(for: indexPath) else { return false }
return item == .chat 
``` 

or something...

Toshi/Controllers/Recent/ThreadsDataSource.swift
marijnschilling: Oke, but maybe you have another idea how to break up this functionality in that case? It's not about my preferences it's more that i ran into a lot of probems with the DappsDataSource being one big dataSource that handles and the frontpage and the Search. So if you have any other solution how to prevent that kind of issues presenting itself here int he future I'm open to that ✌️  it should of course make semantic sense to you!

Toshi/Controllers/Chats/ChatsDataSource.swift
```diff
@@ -0,0 +1,533 @@
...
+}
+
+extension ChatsDataSource: UITableViewDataSource {
+
+    func numberOfSections(in tableView: UITableView) -> Int {
+        return numberOfSections
+    }
+
+    func tableView(_: UITableView, numberOfRowsInSection section: Int) -> Int {
+        guard section < sections.count else { return 0 }
```
designatednerd: Why is this guard necessary

Toshi/Controllers/Chats/ChatsNavigationController.swift
designatednerd: i think these naming conventions are ones Yulia grabbed from a previous implementation of mine, fwiw - also the reason that this is `with` instead of `on` is that `NavBarColorChanging` is a protocol, so it doesn't necessarily have to be an actual nav bar, it can also be any other kind of object. So you're setting things *with* the colorable

Toshi/Controllers/Chats/ChatsNavigationController.swift
designatednerd: so this is checking if the previous VC implements `navBarColorChanging` - if it does, it applies those colors. If it doesn't, it leaves things as they are. Is that clearer?

Toshi/Controllers/Chats/ChatsNavigationController.swift
designatednerd: it is setting the colors of the `NavBarColorChanging` object passed in with the parameter. It may or may not be a view controller.

Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+
+    private func showThread(at indexPath: IndexPath) {
+        guard let thread = threadToShow(at: indexPath) else { return }
+        let chatViewController = ChatViewController(thread: thread)
+        navigationController?.pushViewController(chatViewController, animated: true)
+    }
+
+    func updateContactIfNeeded(at indexPath: IndexPath) {
+        if let thread = dataSource.acceptedThread(at: indexPath.row, in: 0), let address = thread.contactIdentifier() {
+            DLog("Updating contact info for address: \(address).")
```
designatednerd: do these `DLog`s still need to be in here?

Toshi/Protocols/SystemSharing.swift
```diff
@@ -1,10 +1,17 @@
+// Copyright (c) 2018 Token Browser, Inc
 //
-//  SystemSharing.swift
-//  Debug
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
 //
-//  Created by Ellen Shapiro (Work) on 2/9/18.
```
designatednerd: 🤦‍♀️  sorry about that

Toshi/Controllers/Chats/ChatsNavigationController.swift
marijnschilling: Aah so it's setting the current nav colors equal to the `NavBarColorChanging` object?

yuliaveres:
@designatednerd do you have other concerns ?

designatednerd:
@yuliaveres nope good on my end 👍

Toshi/Controllers/Recent/ChatsViewController.swift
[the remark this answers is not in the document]
Elland: Right, I recall the hack now 👍🏽

Toshi/Controllers/Recent/ChatsViewController.swift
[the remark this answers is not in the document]
Elland: Makes sense. Mind adding a comment explaining? 👍

Toshi/Controllers/Recent/ChatsViewController.swift
```diff
@@ -0,0 +1,234 @@
...
+            if dataSource.hasUnacceptedThreads {
+                let messagesRequestsViewController = ChatsViewController(style: .grouped, target: .unacceptedThreadRequests)
+                navigationController?.pushViewController(messagesRequestsViewController, animated: true)
+            } else {
+                showThread(at: indexPath)
+            }
+        case .chat:
+            showThread(at: indexPath)
+        case .inviteFriend:
+            shareWithSystemSheet(item: Localized.sharing_action_item)
```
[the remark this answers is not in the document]
Elland: Yeah, reads like this is going to share directly, instead of presenting a view controller. 👍

Toshi/Controllers/Recent/ChatsViewController.swift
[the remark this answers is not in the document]
Elland: Gotta love UIKit :D

Toshi/Controllers/Recent/ThreadsDataSource.swift
```diff
@@ -277,12 +350,179 @@ final class ThreadsDataSource: NSObject {
...
         }
     }
+
+    private func messagesRequestsCell(for indexPath: IndexPath) -> UITableViewCell {
+        guard let firstUnacceptedThread = unacceptedThread(at: IndexPath(row: 0, section: 0)) else {
+            return UITableViewCell(frame: .zero)
+        }
+
+        let cellConfigurator = CellConfigurator()
+        var cellData: TableCellData
```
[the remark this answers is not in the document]
Elland: But it's only set once. It's not changed after being set. You can  do that with a let:

```swift
let cellData: CellData

if a {
    cellData = cellDataA
else {
    cellData = cellDataB
}
```

Toshi/Controllers/Recent/ThreadsDataSource.swift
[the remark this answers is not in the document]
Elland: or both 👯🏽‍♂️

Toshi/Controllers/Recent/ChatsViewController.swift
Elland: I agree with  the naming being sub-optimal, but the backing model is a `TSThread`, right? Maybe we can use better naming on the SSK rewrite. 👍

PR #808, opened 2018-04-09
[IOS-571] remove favorites controller and tab

marijnschilling:
img_f255b9834438-1

Removed the favorites functionality.

Left it in the ProfileDataSource, because its too complicated :)


Toshi/Controllers/TabBarController.swift
```diff
@@ -81,9 +79,6 @@ class TabBarController: UITabBarController, OfflineAlertDisplaying {
 
     @objc func setupControllers() {
         dappsViewController = DappsNavigationController(rootViewController: DappsViewController())
-        let datasource = ProfilesDataSource(type: .favorites)
```
yuliaveres: Do we need `ProfilesDataSource` without having Favourites functionality ?

yuliaveres: as far as I remember this was needed only for search and filtering in database among contacts (favourites)

Toshi/Controllers/TabBarController.swift
```diff
@@ -291,24 +282,6 @@ extension TabBarController: ScannerViewControllerDelegate {
             scannerController.startScanning()
         }
     }
-
-    private func viewContact(with contactName: String) {
-        idAPIClient.retrieveUser(username: contactName) { [weak self] contact in
```
yuliaveres: we do not support profile qr code scanning at all anymore ?

marijnschilling: Nope, we discussed this with marek in toshi channel as well :V: We should consider removing the `MY QR-code` screen form settings as well.

marijnschilling:
haha I solved it like this now, just initialise the ProfilesController with a type instead of the data source. makes way more sense. Just noted that we don't exclude already added group members when we search for group members to add (so when in .updateGroupChat state.

PR #815, opened 2018-04-10
Return when there's no sections

marijnschilling:
The app would crash when you opened a new account, this fixes it. Not sure if its the best fix though, so open for comments ✌️ 

It crashed because it would'nt have set the the sections yet, because that happens in loadMessages() but when the user is nil it shows the notification instead of loading the messages. 

from line 168 in `ChatsDataSource`:
```
        if TokenUser.current != nil {
            viewModel.setupForCurrentSession()
            loadMessages()
            registerNotifications()
        } else {
            NotificationCenter.default.addObserver(self, selector: #selector(chatDBCreated(_:)), name: .ChatDatabaseCreated, object: nil)
        }
```


Toshi/Controllers/Chats/ChatsDataSource.swift
```diff
@@ -516,6 +516,8 @@ extension ChatsDataSource: UITableViewDataSource {
     }
 
     func tableView(_: UITableView, numberOfRowsInSection section: Int) -> Int {
+        guard sections.count > 0 else { return 0 }
```
designatednerd: I think this got pulled because I asked Yulia what this was for - can you throw in a comment so we don't make that mistake again? :smile:

yuliaveres: and i forgot why ! comment is a good thing 🙈

PR #816, opened 2018-04-10
Use calculated status bar frame since its different on iPhoneX

marijnschilling:
and all is right with the world again 😅 

So i pin it to the top of the view so it will 'cover' (but then be behind) the status bar during the transition animation, that was what was causing that glitch in the animation. I just forgot to get the status bar height from the application instead of using the hardcoded 20px 🙈 

screenshot 2018-04-10 16 53 28


Toshi/Controllers/Chats/NewChatViewController.swift
```diff
@@ -102,10 +102,10 @@ final class NewChatViewController: UIViewController {
         automaticallyAdjustsScrollViewInsets = false
         definesPresentationContext = true
 
-        searchHeaderView.top(to: layoutGuide())
+        searchHeaderView.top(to: view)
         searchHeaderView.left(to: view)
         searchHeaderView.right(to: view)
-        searchHeaderView.height(56)
+        searchHeaderView.height(56 + UIApplication.shared.statusBarFrame.size.height)
```
designatednerd: the status bar height doesn't always correspond with the safe area insets though, especially if the status bar is in double-height mode on a non-X. Maybe something where you pin the bottom to the bottom of the safe area inset + 56?

Toshi/Views/PushedSearchHeaderView.swift
```diff
@@ -109,7 +109,7 @@ final class PushedSearchHeaderView: UIView {
 
     private func addSubviewsAndConstraints() {
         addSubview(stackView)
-        stackView.edges(to: self, insets: UIEdgeInsets(top: 20, left: 0, bottom: 0, right: 0))
+        stackView.edges(to: self, insets: UIEdgeInsets(top: UIApplication.shared.statusBarFrame.size.height, left: 0, bottom: 0, right: 0))
```
designatednerd: same concern about this being really susceptible to status bar/safe area layout inset fuckery since the status bar frame can change

Toshi/Controllers/Chats/NewChatViewController.swift
marijnschilling: But isn't it also good then when it is in double height mode? 

The safe area inset is only accessible after the view is visible 🙈

designatednerd: only if the user doesn't exit double height mode while looking at the screen (ex, have a phone call or an audio recording end in the background). 

and true re the `safeAreaLayoutGuide` but if you pin to the bottom of it, I *think* that should basically extend the search header view beyond it, no matter where it is - does that make sense?

marijnschilling:
Thanks @designatednerd! It worked, and its way  better this way:

screenshot 2018-04-10 17 33 49

screenshot 2018-04-10 17 31 18

PR #822, opened 2018-04-11
WIP: [IOS-589, IOS-575]: Profiles controller and its datasource

yuliaveres:
screenshot at apr 11 15-58-21


yuliaveres:
still working on improvements but ready to take a look at

Toshi/APIClients/idAPIClient.swift
```diff
@@ -656,4 +658,108 @@ final class IDAPIClient {
...
+                    return
+                }
+
+                frontPage.sections.forEach({ section in
+                    section.profiles.forEach { AvatarManager.shared.downloadAvatar(for: $0.avatar ?? "") }
+                })
+
+                sections = frontPage.sections
+            case .failure(_, _, let error):
+                DLog(error.localizedDescription)
```
designatednerd: does this `DLog` need to stay?

Toshi/APIClients/idAPIClient.swift
```diff
@@ -656,4 +658,108 @@ final class IDAPIClient {
...
+
+    /// Searches profiles of a given type and optional search text.
+    ///
+    /// - Parameters:
+    ///   - type: stringtype of searched objects - user, bot or group.
+    ///   - searchText: optional search text.
+    ///   - completion: The completion closure to execute when the request completes
+    ///                 - dapps: A list of dapps, or nil
+    ///                 - toshiError: A toshiError if any error was encountered, or nil
+    func searchProfilesOfType(_ type: String, for searchText: String? = nil, completion: @escaping SearchedProfilesResults) {
```
designatednerd: What happens here if someone passes in a type other than `user`, `bot`, or `group`? Should we limit it to only these types?

Toshi/Controllers/Chats/NewChatViewController.swift
```diff
@@ -115,16 +137,19 @@ final class NewChatViewController: UIViewController {
...
+
+    func searchHeaderViewDidUpdateSearchText(_ headerView: PushedSearchHeaderView, _ searchText: String) {
+        // TODO: Search in data source
     }
 
-    func searchHeaderWillEndEditing(_ headerView: PushedSearchHeaderView) {
-        dataSource.isSearching = false
+    func searchHeaderWillBeginEditing(_ headerView: PushedSearchHeaderView) {
+        // TODO: Might be use for table views visibility handling
+        // comes lated when search is being implemented
```
designatednerd: here and below: `lated` -> `later`

Toshi/Controllers/UsersBotsGroups/FindProfilesViewController.swift
```diff
@@ -0,0 +1,173 @@
...
+        view.tableHeaderView = tableHeaderView
+        view.alpha = 0
+
+        return view
+    }()
+
+    private lazy var tableHeaderView: SegmentedHeaderView = {
+        let walletItemTitles = [Localized.users_section_title, Localized.bots_section_title, Localized.groups_section_title]
+        let headerView = SegmentedHeaderView(segmentNames: walletItemTitles, delegate: self)
+        let frame = CGRect(origin: .zero, size: CGSize(width: view.bounds.width, height: 44))
```
designatednerd: for `44` i think there's a constant something like `defaultBarHeight` in the `Theme` file

Toshi/Controllers/UsersBotsGroups/FindProfilesViewController.swift
```diff
@@ -0,0 +1,173 @@
...
+        view.addSubview(searchResultsTableView)
+        view.addSubview(searchHeaderView)
+
+        edgesForExtendedLayout = []
+        automaticallyAdjustsScrollViewInsets = false
+        definesPresentationContext = true
+
+        searchHeaderView.top(to: layoutGuide())
+        searchHeaderView.left(to: view)
+        searchHeaderView.right(to: view)
```
designatednerd: i'm just realizing this and I think i've been doing what you've been doing here but: Should we be pinning left and right to the layout guide as well? That should prevent any weirdness in the future if for instance we enable rotation

Toshi/Controllers/UsersBotsGroups/ProfilesDataSource.swift
```diff
@@ -0,0 +1,129 @@
...
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
+        delegate?.didSelectRow(at: indexPath)
+    }
+}
+
+extension ProfilesDataSource: DappsSectionHeaderViewDelegate {
+    func dappsSectionHeaderViewDidReceiveActionButtonEvent(_ sectionHeaderView: DappsSectionHeaderView) {
+
```
designatednerd: `//TODO`?

Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
```diff
@@ -0,0 +1,133 @@
...
+            return Localized.bots_section_title
+        case .group:
+            return Localized.groups_section_title
+        }
+    }
+
+    var typeString: String {
+        switch self {
+        case .user:
+            return ProfileType.userTypeString
```
designatednerd: here i thnk you can just return the string directly instead of having the `private static let`s above since both serve to keep callers from accessing the underlying string and centralize access to it.

Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
```diff
@@ -0,0 +1,133 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+import UIKit
+
+enum ProfileType: Int {
```
designatednerd: I think it might make more sense to keep this `enum` in the same file as the `Profile` object rather than here. 

Also, why an `Int` enum rather than a `String` enum?

Toshi/Models/UsersBotsGroups/ProfilesFrontPage.swift
```diff
@@ -0,0 +1,188 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+/// Users, bots and groups front page
+final class ProfilesFrontPage: Codable {
```
designatednerd: Couple things: 
1) I'd recommend making these `struct` instead of `final class` since that allows automatic synthesis of initializers, which doesn't whine in swift 4.0.3 but whines like hell in Swift 4.1
2) I would recommend naming this file after the non-wrapper class in it - `Profile.swift` - `ProfilesFrontPage` isn't really what I would be looking for if I were looking for the Profile object.

Toshi/Models/UsersBotsGroups/ProfilesFrontPage.swift
```diff
@@ -0,0 +1,188 @@
...
+            _current = newValue
+            DispatchQueue.main.async {
+                NotificationCenter.default.post(name: .currentUserUpdated, object: nil)
+            }
+        }
+    }
+
+    var dictionary: [String: Any]? {
+        guard let data = try? JSONEncoder().encode(self) else { return nil }
+        return (try? JSONSerialization.jsonObject(with: data, options: .allowFragments)).flatMap { $0 as? [String: Any] }
```
designatednerd: does this not work with just `as? [String: Any]`?

Toshi/Models/UsersBotsGroups/SearchedProfilesData.swift
```diff
@@ -15,30 +15,21 @@
 
 import Foundation
 
-struct UserBotsGroupsQueryData {
+/// Searched sers, bots and groups response data
+final class SearchedProfilesData: Codable {
```
designatednerd: same advice re: making this a struct instead of a `final class`

Toshi/APIClients/idAPIClient.swift
yuliaveres: yeah .. i think so

Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
yuliaveres: it easy to convert segmented header selected index `Int` to `ProfileType`

Toshi/Controllers/UsersBotsGroups/ProfilesDataSource.swift
yuliaveres: is implemented now

PR #823, opened 2018-04-11
[IOS-597] implement pushed search header on new group controller

marijnschilling:
I aded the pushed search header to group chat 🎉 

But while doing that i realised a whole bunch of things can go now. We used to use the `ProfileviewController` for the favorites, for creating a new chat, for creating a new group and for editing group members. Now we only use it for the last two cases. So i cleaned up that whole controller and removed all unnecessary stuff as well. 
- [ ] removed the profileHeaderView that adds a searchBar to it in <iOS11 (so basically removed a whole bunch of `if iOS11 is available lalalala`)
- [ ] I added the `ProfilesAddedToGroupHeader` as a normal view (no more tableViewHeader bs)
- [ ] Removed the custom `BrowseSearchResultsTableView` and added this as _the_ tableView to the profile view, since the favorites is gone anyway only search table view is there.
- [ ] removed the last remnants of favorites in localized strings and other things that were hanging around.

Create new group:
new group

Add to group:
add to group


Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -18,14 +18,11 @@ import UIKit
 // MARK: - Profiles View Controller Type
 
 public enum ProfilesViewControllerType {
```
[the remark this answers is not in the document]
yuliaveres: for me it is kind of "picker", user picker .. the purpose of which is to select a few users from some list.. but i do not mind it to stay like this 🤷‍♂️

Toshi/Controllers/Chats/FindUsersBotsGroupsViewController.swift
```diff
@@ -93,6 +92,10 @@ final class FindUsersBotsGroupsViewController: UIViewController {
 }
 
 extension FindUsersBotsGroupsViewController: PushedSearchHeaderDelegate {
+    func searchHeaderDidEnterSearchTerm(_ searchText: String, _ headerView: PushedSearchHeaderView) {
+
```
yuliaveres: empty ?

Toshi/Controllers/Chats/NewChatViewController.swift
```diff
@@ -115,6 +115,10 @@ final class NewChatViewController: UIViewController {
 }
 
 extension NewChatViewController: PushedSearchHeaderDelegate {
+    func searchHeaderDidEnterSearchTerm(_ searchText: String, _ headerView: PushedSearchHeaderView) {
+
```
yuliaveres: empty or just leftover ?

Toshi/Views/PushedSearchHeaderView.swift
```diff
@@ -16,18 +16,26 @@
...
     func searchHeaderWillBeginEditing(_ headerView: PushedSearchHeaderView)
     func searchHeaderWillEndEditing(_ headerView: PushedSearchHeaderView)
 
-    func searchHeaderDidReceiveCancelEvent(_ headerView: PushedSearchHeaderView)
+    func searchHeaderDidReceiveRightButtonEvent(_ headerView: PushedSearchHeaderView)
     func searchHeaderViewDidReceiveBackEvent(_ headerView: PushedSearchHeaderView)
 }
 
+enum PushedSearchHeaderViewState: Int {
+    case select, normal
```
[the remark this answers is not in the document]
yuliaveres: does the title of the right button depend on this only ?
if so - lets make it as accessible property so controller can say `pushedHeader.rightButtonTitle = "whatever"` so it is much more flexible, won't need to add any more cases if a new title is needed, and won't need to think on naming of these cases

Toshi/Views/PushedSearchHeaderView.swift
```diff
@@ -138,14 +168,16 @@ final class PushedSearchHeaderView: UIView {
...
+        rightButton.isHidden = true
     }
 
     private func adjustToSearching(isSearching: Bool) {
         layoutIfNeeded()
-        cancelButton.isHidden = !isSearching
-        backButton.isHidden = isSearching
+        rightButton.isHidden = !isSearching
 
+        guard state == .normal else { return }
```
yuliaveres: also can go to property `hidesRightButtonOnSearch` which by default might be YES / true

Toshi/Views/PushedSearchHeaderView.swift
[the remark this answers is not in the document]
marijnschilling: No it also decides wether the button should be enabled

Toshi/Controllers/Chats/FindUsersBotsGroupsViewController.swift
marijnschilling: We will need to implement this later on, i can place a comment in it!` //TODO:`

Toshi/Views/PushedSearchHeaderView.swift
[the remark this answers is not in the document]
marijnschilling: But yeah i think it's a good point. If we would even need a third isolated state to use this view we can add an enum. But for now just vars is easier to make it more flexible ✌️

Toshi/Controllers/Profiles/ProfilesViewController.swift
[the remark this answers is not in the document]
designatednerd: I'm good with `GroupParticipantsViewController` but I think we can also postpone dealing with the naming for a bit

Toshi/Views/PushedSearchHeaderView.swift
```diff
@@ -138,14 +156,16 @@ final class PushedSearchHeaderView: UIView {
...
+        rightButton.isHidden = true
     }
 
     private func adjustToSearching(isSearching: Bool) {
         layoutIfNeeded()
-        cancelButton.isHidden = !isSearching
-        backButton.isHidden = isSearching
+        rightButton.isHidden = !isSearching
+
+        guard hidesRightButtonOnSearch else { return }
```
designatednerd: should this `guard` be above where `rightButton.isHidden` is set?

marijnschilling: oops, good catch! the place is right, the name was wrong. We're hiding the backButton on search in some cases, not the right button.

PR #824, opened 2018-04-11
Xcode 9.3 / Swift 4.1 / CocoaPods 1.5 / Carthage 0.29

designatednerd:
.... and a partiiiidge in a Pear Treeeeeee!

In this PR: 
- Update to swift 4.1 and fix what that done broke locally (mostly renaming `flatMap` -> `compactMap`
- Rebuild Carthage stuff with new version of Carthage and Swift 4.1, fix what that done broke (mostly in Earl Grey robot)
- Add `AwesomeCache` to Cartfile since that seems to have gotten lost somewhere along the way
- Change a bunch of `final class`es  to structs in terms of data being passed around - we were losing the ability to have default constructors and since everything in the updated structs uses `let`s, we don't really lose much by switching to copy-on-write instead of pass-by-reference anyway. 
- some adding of `self->` in Obj-C to note explicit capture semantics which we're now getting warnings about in 9.3
- Add README instructions for reinstalling Marathon off of the Swift 4.1 branch
- Update `swift-lint` and `fabric` pods using new version of CocoaPods
- Clean up some `swift-lint` warnings which popped up


designatednerd:
(note: This is going to remain failing until I can diagnose what the hell is going on with logic around JSON updates)

designatednerd:
Looks like there's some oddness with pulling `Floats` out of an `NSDictionary` in Swift 4.1 - probably another good reason for us to switch the `User` model to `Codable` sooner rather than later.

yuliaveres:
@designatednerd it is being started in my latest PR

yuliaveres:
it will require some refactor throughout the app, the new codable is `Profile` model

marijnschilling:
So is this till WIP or is it ready for review?

designatednerd:
@marijnschilling ready for review!

designatednerd:
Finally passing! Merging.

PR #825, opened 2018-04-12
[IOS-573] Search users bots groups view

marijnschilling:
Maybe its just done like this actually, the dataSource can be added easily ✌️ 

I added it to the NewGroupViewController and the FindUserBotsGroups one:
search


Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -353,6 +353,11 @@
 "chat_payment_recieved" = "Payment received";
 "chat_payment_sent" = "Payment sent";
 
+/* Search Chat */
+"search_chat_users_tab_name" = "Users";
+"search_chat_groups_tab_name" = "Bots";
```
yuliaveres: groups

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -353,6 +353,11 @@
...
+"search_chat_bots_tab_name" = "Groups";
```
yuliaveres: bots

marijnschilling: oops 🙈

marijnschilling:
img_10ce27032c77-1
All better now 😅

yuliaveres:
All good 👍

PR #828, opened 2018-04-16
[IOS-601] Description text cell

marijnschilling:
Made a new Basic cell subclass that also holds a description text.

I made the titleTextField a subclass of UITextField so it can hold a block to set the font dynamically when the user resizes the font. This way we can set the wanted font from outside the basic cell itself.

The only worry I have for this is that setting it this way might be a bit unclear to why this is necessary. So any tips on how to make it less confusing for developers using this is super welcome! 

here you see the cells displayed in the SearchDapps results. Just to test them out: 
img_85be9751a4a4-1


marijnschilling:
It's failing because of an error: 

`"Profile" has no initialisers`

Which is weird, cause locally it runs fine!

designatednerd:
do you have the xcode 9.3 changes pulled into this branch?

yuliaveres:
yeah you have to merge `feature/chat-redesign` in

marijnschilling:
yeah i did

PR #830, opened 2018-04-16
[IOS-607, IOS-611]: Profiles search on tiny delay and request cancelling

yuliaveres:
Make a search network request after 0.2 delay instead of immediately after each symbol is typed in;
cancel each previous request


Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
```diff
@@ -54,19 +56,25 @@ final class SearchProfilesDataSource: NSObject {
         queryData.searchText = text
```
Elland: Do we need the empty line above?

yuliaveres:
@Elland  shit not everything is pushed 😭  adding

Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
```diff
@@ -54,19 +56,25 @@ final class SearchProfilesDataSource: NSObject {
...
         queryData.type = type
 
-        idAPIClient.searchProfilesOfType(queryData.type, for: queryData.searchText) { [weak self] profiles, type, error in
+        reloadTimer?.invalidate()
```
designatednerd: reload timer is invalidated here

yuliaveres:
@designatednerd  any thoughts ?

PR #831, opened 2018-04-16
[IOS-603] Separators on basic cell

marijnschilling:
Added a funny way to show separators in the style marek made in sketch.

now the `FindProfilesController` looks like this:

img_f8d1ac458cff-1


Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -175,22 +175,57 @@ class BasicTableViewCell: UITableViewCell {
...
         fatalError("addSubviewsAndConstraints() should be overriden")
     }
 
+    func showSeparator(forLastCellInSection: Bool = false) {
+        separator.isHidden = false
+        if forLastCellInSection {
+            separatorLeftConstraint?.constant = 15
+            separatorRightConstraint?.constant = -15
+        } else {
+            separatorLeftConstraint?.constant = 80
```
designatednerd: are there other numbers that should add up to 80 that make this work? ie, margin + margin + size of icon?

PR #832, opened 2018-04-16
iOS-587 / iOS-609: Fixing some X layout issues

designatednerd:
fixed 587: 

simulator screen shot - iphone x - 2018-04-16 at 21 19 11

fixed 609: 

simulator screen shot - iphone x - 2018-04-16 at 21 25 42


Toshi/Controllers/Payments/PaymentConfirmationViewController.swift
```diff
@@ -277,6 +285,11 @@ final class PaymentConfirmationViewController: UIViewController {
 
             hasLaidOutBefore = true
 
+            if (view.frame.height - layoutGuide().layoutFrame.height) > 0 {
```
marijnschilling: What does this check actually do? I feel you might be able to put the comments in a var name for a bool here.
`let isThereAGapBetweenTheBottomLalallalala = (view.frame.height - layoutGuide().layoutFrame.height) > 0` But than something that makes more sense? Is this gap only on iPhoneX? If so that might be good to mention

designatednerd: in theory it's only on x but i'll do some renaming since I have to get the effing thing to build again

PR #833, opened 2018-04-17
[CORE-118]: Chat redesign

yuliaveres:


Toshi/Controllers/Chats/ChatsDataSource.swift
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    func updateNewThreadRecepientsIfNeeded(_ thread: TSThread) {
+        DispatchQueue.main.async {
+            if let contactId = thread.contactIdentifier() {
+                guard SessionManager.shared.contactsManager.tokenContact(forAddress: contactId) == nil else { return }
+
+                let contactsIds = SessionManager.shared.contactsManager.tokenContacts.map { $0.address }
+
+                IDAPIClient.shared.findContact(name: contactId, completion: { foundUser in
```
designatednerd: @yuliaveres can this be extracted to a protocol?

yuliaveres: not today probably ..

yuliaveres: also when i look into it - why would you extract it ? here is just updating group members list, and is there is any not stored in the db yet - it is done. happens only here.
maybe you meant chats fetch to be extracted into protocol ?

yuliaveres: @designatednerd  ^

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -0,0 +1,216 @@
...
+    private func saveSettings() {
+        Yap.sharedInstance.insert(object: userSettings, for: toshiId, in: TokenUser.localUserSettingsKey)
+    }
+
+    private mutating func adjustToLocalCurrency() {
+        updateLocalCurrencyLocaleCache()
+
+        ExchangeRateClient.updateRate({ _ in
+            NotificationCenter.default.post(name: .localCurrencyUpdated, object: nil)
+        })
```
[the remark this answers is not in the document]
yuliaveres: this is basically copied from previous TokenUser model, but i will extract update and notify separately at least

Toshi/Protocols/ProfilesProvider.swift
```diff
@@ -15,12 +15,12 @@
 
 import Foundation
 
-protocol ProfilesProvider: class {
```
designatednerd: 👍  on changes, file name needs to be changed too though

yuliaveres: 🤦‍♀️ !

Toshi/Controllers/Chats/ChatsDataSource.swift
designatednerd: i think more something around "If this user exists in the contacts database, return them, otherwise, fetch their info"

yuliaveres: it does exactly how the function is called `updateNewThreadRecepientsIfNeeded`, does not return no

designatednerd: we should probably discuss this more later, it's totally fine for now

PR #836, opened 2018-04-17
[IOS-602] [IOS-608] add tests for new chat

marijnschilling:
Test added for profile front page and different types of profile search.

Also added an extra test for creating the path used in search, I extracted a method for that in the `IDAPIClient`.

And i renamed the filename of `idAPIClient` to be equal to the classname `IDAPIClient`


marijnschilling:
woops, created from feature branch that is now closed. I created a new PR from master here: https://github.com/toshiapp/toshi-ios-client/pull/837

PR #837, opened 2018-04-17
[IOS-602] [IOS-608] tests for new chat 

marijnschilling:
I added the tests in another file then the previous tests for `IDAPIClient` because those were written in QuickSpec, and @designatednerd was not so happy about that. So cheers to a new beginning! (the old tests I have renamed to `IDAPIClientQuickSpecs`)

Test added for profile front page and different types of profile search.

Also added an extra test for creating the path used in search, I extracted a method for that in the IDAPIClient.

And i renamed the filename of idAPIClient to be equal to the classname IDAPIClient


Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: IDAPIClientQuickSpecs.self), mockFilename: "fetchUsersFromPaymentAddresses")
+                    subject = IDAPIClient(teapot: mockTeapot)
+
+                    let paymentAddress = "0x1ad0bb2d14595fa6ad885e53eaaa6c82339f9b98"
+
+                    waitUntil { done in
+                        subject.findUserWithPaymentAddress(paymentAddress) { user, _ in
+                            expect(user).toNot(beNil())
+
+                            expect(user?.name).to(equal("Marijn Schilling"))
```
designatednerd: worth unwrapping the user here so you don't have to have a bunch of optionals?

Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: IDAPIClientQuickSpecs.self), mockFilename: "user")
+                    subject = IDAPIClient(teapot: mockTeapot)
+
+                    let username = "marijnschilling"
+
+                    waitUntil { done in
+                        subject.retrieveUser(username: username) { user in
+                            expect(user).toNot(beNil())
+                            
+                            expect(user?.name).to(equal("Marijn Schilling"))
```
designatednerd: same q on unwrapping `user`

Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: IDAPIClientQuickSpecs.self), mockFilename: "contact")
+                    subject = IDAPIClient(teapot: mockTeapot)
+                    
+                    let username = "designatednerd"
+
+                    waitUntil { done in
+                        subject.findContact(name: username) { user in
+                            expect(user).toNot(beNil())
+                            
+                            expect(user?.name).to(equal("Ellen Shapiro"))
```
designatednerd: one more `user`

Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                        "0x1c0e100fcf093c64cdaa54562c84c8fa0171a38c",
+                        "0xeedde9afd8ac67e3564f69d0c1a144e9d7536af1"
+                    ]
+                    
+                    waitUntil { done in
+                        subject.fetchUsers(with: ids) { users, error in
+                            expect(users).toNot(beNil())
+                            expect(error).to(beNil())
+                            
+                            expect(users?.count).to(equal(3))
```
designatednerd: worth unwrapping `users`?

Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                            expect(success).to(beTrue())
+                            expect(error).to(beNil())
+                            
+                            done()
+                        }
+                    }
+                }
+
+                it("logs in") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: IDAPIClientQuickSpecs.self), mockFilename: "", statusCode: .noContent)
```
designatednerd: why is there no mock file for this one? also, what about non-admin login?

Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                it("fetches the timestamp") {
+                    let mockTeapot = MockTeapot(bundle: Bundle(for: IDAPIClientQuickSpecs.self), mockFilename: "timestamp", statusCode: .unauthorized)
+                    subject = IDAPIClient(teapot: mockTeapot)
+
+                    waitUntil { done in
+                        subject.fetchTimestamp { timestamp, error in
+                            expect(timestamp).to(beNil())
+                            expect(error).toNot(beNil())
+                            
+                            expect(error?.responseStatus).to(equal(401))
```
designatednerd: worth unwrapping `error`?

Tests/IDAPIClientQuickSpecs.swift
```diff
@@ -0,0 +1,563 @@
...
+                        fail("Could not create test image")
+                        
+                        return
+                    }
+                    waitUntil { done in
+                        subject.updateAvatar(testImage) { success, error in
+                            expect(success).to(beFalse())
+                            expect(error).toNot(beNil())
+                            
+                            expect(error?.responseStatus).to(equal(401))
```
designatednerd: unwrap `error` [here and through the rest of the file]?

Tests/IDAPIClientTests.swift
```diff
@@ -1,563 +1,192 @@
...
-                    waitUntil { done in
-                        subject.adminLogin(loginToken: token) { success, error in
-                            expect(success).to(beTrue())
-                            expect(error).to(beNil())
-                            
-                            done()
-                        }
-                    }
-                }
+class IDAPIClientTests: XCTestCase {
```
designatednerd: Much as I get annoyed by `QuickSpec`, if you find it more helpful to use, it's fine to keep using it. ! I'd rather have tests whose style annoys me then leaving things untested. Please don't just make a change because I'm a giant grump, make it because you find something helpful about using this kind of test instead!

Tests/IDAPIClientTests.swift
```diff
@@ -1,563 +1,192 @@
...
-                            expect(dapps).to(beNil())
-                            
-                            expect(error?.responseStatus).to(equal(401))
-                            expect(error?.type).to(equal(.invalidResponseStatus))
-                            
-                            done()
-                        }
-                    }
-                }
+            XCTAssertEqual(searchedProfilesResults.map { $0.type }, ["groupbot"])
```
designatednerd: and again, this can probably be pulled out using `first` since there's only one

Tests/IDAPIClientTests.swift
```diff
@@ -1,563 +1,192 @@
...
+
+            XCTAssertEqual(frontPageResult.map { $0.name }, ["Popular Groups", "Featured Bots", "Public Users"])
+
+            guard frontPageResult.count == 3 else {
+                XCTFail("json result should have 3 sections")
+                return
+            }
+
+            let popularGroups = frontPageResult[0]
+            XCTAssertEqual(popularGroups.profiles.map { $0.type }, ["groupbot"])
```
designatednerd: since you're only working with one item here, maybe pull this one out using `popularGroups.first`

Tests/IDAPIClientTests.swift
marijnschilling: Hmm yeah that's a good point! But i must say i just tried out QuickSpec because of the  "everyone is using it"-argument. And I think it's good to try out XCTest as well 🤷‍♂️  Good to see the downsides and benefits of both.

I think you already mentioned one of the biggest downsides, that apparently if you run one test "under water" it's still running the entire test class. 

I do like that QuickSpec forces you to be verbose about your tests.

And what i don't like is having two different test styles in our project 😢 ...consistency...!

Tests/IDAPIClientQuickSpecs.swift
marijnschilling: Haa this file of QuickSpec I actually only just renamed to `IDAPIClientQuickSpecs` instead of `IDAPIClientTests` 😬 So not sure if we should review on this yes or no 🤔 (I wrote this a gazillion years ago)

designatednerd: never to late to clean up 😄 

(Note: This does not apply to IRL cleaning)

Tests/IDAPIClientTests.swift
designatednerd: Yah, I think if you could just run one single QuickSpec test without running the whole bloody lot I'd be a lot more okay with it. The syntax still drives me insane, but it'd at least be more usable. 

And consistency is definitely helpful, but I'd rather have inconsistent and tested than consisted and not tested. 😄

Tests/IDAPIClientTests.swift
marijnschilling: Yeah I see that we can do that, but I'd like to keep it the same as in the other cases so it will be much easier to copy paste when it needs to be adjusted...

What do you think about that?

designatednerd: Hm. I see your point but it does feel to me like the mapping creates a lot of unnecessary noise when you could just do `profile.type` for these things that are single items. It's more about ease of reading - which will almost certainly happen way more often than adjusting the tests.

Tests/IDAPIClientQuickSpecs.swift
marijnschilling: I actually don't know! I don't really know what this admin login is doing for us in the app 🤔  Do you know more about that @yuliaveres ?

Tests/IDAPIClientTests.swift
marijnschilling: aha! Good point, I agree :)

Tests/IDAPIClientQuickSpecs.swift
yuliaveres: all that is expected from response - is success. this is login to admin service, not login of a user inside toshi

Toshi/APIClients/IDAPIClient.swift
```diff
@@ -789,4 +785,13 @@ final class IDAPIClient {
             }
         }
     }
+
+    func createPathFor(_ type: String, with searchText: String?) -> String {
```
yuliaveres: why not private ?

marijnschilling: Cause then I can't test it 😭

PR #838, opened 2018-04-18
[IOS-616]: Screen protection flow improvements

yuliaveres:
https://toshiapp.atlassian.net/browse/IOS-616


Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -263,14 +264,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     // Screen protection
 
-    private func activateScreenProtection() {
-
-        guard screenProtectionWindow == nil else { return }
-
+    private lazy var screenProtectionWindow: UIWindow = {
         let window = UIWindow()
         window.isHidden = true
         window.backgroundColor = .clear
```
Elland: if it's transparent, how does it work atm?

yuliaveres: it has blur effect on top, 

```
let effectView = UIVisualEffectView(effect: UIBlurEffect(style: .extraLight))
window.addSubview(effectView)
```

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -83,21 +82,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func applicationWillResignActive(_ application: UIApplication) {
 
+        DispatchQueue.main.async {
+            self.activateScreenProtection()
```
designatednerd: why does this need to get dispatched on the main queue? Won't this come in on the main queue?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -284,22 +282,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         effectView.bottomAnchor.constraint(equalTo: window.bottomAnchor).isActive = true
         effectView.rightAnchor.constraint(equalTo: window.rightAnchor).isActive = true
```
designatednerd: ^ worth converting some of these to use `TinyConstraints` while you're mucking around in here?

Toshi/Supporting Files/AppDelegate.swift
```diff
@@ -82,31 +102,26 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationWillResignActive(_ application: UIApplication) {
+        DispatchQueue.main.async {
+            self.activateScreenProtection()
+        }
```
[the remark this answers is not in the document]
yuliaveres: so we had issues before, sometimes race conditions were happening and the activation / deactivation just did not work actually. by scheduling this to next runloop we are sure is going to work 🤷‍♂️  what can i say else in this cases 🤷‍♂️  🙈

PR #839, opened 2018-04-19
[IOS-618]: Fix chats sections realod race condition crash

yuliaveres:
https://fabric.io/bakken--bck2/ios/apps/org.toshi.distribution/issues/5ad8766336c7b235274c35d1?time=last-seven-days


Toshi/Controllers/Chats/ChatsDataSource.swift
```diff
@@ -517,7 +517,9 @@ extension ChatsDataSource: UITableViewDataSource {
 
     func tableView(_: UITableView, numberOfRowsInSection section: Int) -> Int {
         // If there is no user session, sections collection is empty
-        guard sections.count > 0 else { return 0 }
+        // or if currently there are less sections and table view tried to reload after Yap insert notification,
```
marijnschilling: So when there's a Yap insert notification to insert a (for example) 3rd section, numberOfRowsInsection is called for this new 3rd section, while the `sections` aren't updated yet? Do I understand that correctly?

Wouldnt this mean that it will show an empty section (since it will return 0 in this case)? Is that the result that is needed? Or is this reload followed up by another reload where the local sections property _has_ been updated? Maybe we should mention that in the comment (see below)

And maybe we can rephrase the comment in `When there's a new section inserted via a Yap insert notification a reload will be triggered and this method will be called before our local sections property is updated. In this case we return 0 because...`

PR #840, opened 2018-04-19
[IOS-617] Use right cell types for search result cells

marijnschilling:
And fix some layout things while im at it.

So wit the way our `CellConfigurator` is set up we need to be careful with initialising CellData with optionals, since if these are nil it can result in a whole other celltype. A bit tricky, we might have to think a bout a way to make this less dependant on that... 🤔


Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
```diff
@@ -100,14 +100,17 @@ extension SearchProfilesDataSource: UITableViewDataSource {
         guard let typeProfiles = profilesMap[queryData.type] else { fatalError() }
         let profile = typeProfiles[indexPath.row]
 
-        let cellData = TableCellData(title: (profile.name ?? "").isEmpty ? profile.username : profile.name, subtitle: profile.displayUsername, leftImagePath: profile.avatar, description: profile.description)
+        let cellData = TableCellData(title: (profile.name ?? "").isEmpty ? profile.username : profile.name, subtitle: profile.displayUsername, leftImagePath: profile.avatar)
```
designatednerd: i think there's a `nameOrUsername` property that does this for you

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -42,6 +42,7 @@ class BasicTableViewCell: UITableViewCell {
     static let horizontalMargin: CGFloat = 16.0
     static let verticalMargin: CGFloat = 15.0
     static let interItemMargin: CGFloat = 10.0
+    static let largeInterItemMargin: CGFloat = 15.0
```
designatednerd: probably shouldn't happen in this PR, but we should probably align these values and the values in `Theme` in another PR. I can do that if you want @marijnschilling but it's all yours if you'd prefer

marijnschilling:
I will make the change!

Toshi/Views/Cells/BasicTableViewCell.swift
```diff
@@ -40,12 +40,13 @@ extension BasicCellActionDelegate {
...
-    static let interItemMargin: CGFloat = 10.0
+    static let verticalMargin: CGFloat  = .spacingx3
+    static let interItemMargin: CGFloat  = .mediumInterItemSpacing
+    static let largeInterItemMargin: CGFloat  = .spacingx3
     static let imageSize: CGFloat = 48.0
     static let doubleImageSize: CGFloat = 48.0
-    static let imageMargin: CGFloat = 10.0
-    static let smallVerticalMargin: CGFloat = 5.0
+    static let imageMargin: CGFloat  = .mediumInterItemSpacing
+    static let smallVerticalMargin: CGFloat  = .smallInterItemSpacing
```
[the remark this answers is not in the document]
marijnschilling: yuuuup copy past mistakey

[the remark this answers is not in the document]
designatednerd: 🍝

Toshi/Controllers/UsersBotsGroups/SearchProfilesDataSource.swift
```diff
@@ -100,14 +100,17 @@ extension SearchProfilesDataSource: UITableViewDataSource {
...
+        let cellData = TableCellData(title: profile.nameOrDisplayName, subtitle: profile.displayUsername, leftImagePath: profile.avatar)
         let configurator = CellConfigurator()
 
         let reuseIdentifier = configurator.cellIdentifier(for: cellData.components)
         guard let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath) as? BasicTableViewCell else { fatalError("Unexpected cell") }
 
         configurator.configureCell(cell, with: cellData)
         cell.accessoryType = .disclosureIndicator
+        cell.titleTextField.setDynamicFontBlock = {
+            cell.titleTextField.font = Theme.preferredSemibold()
```
yuliaveres: Cell or self is captured in each place you use `setDynamicFontBlock`. cell is never destroyed, we have memory issues when you scroll or enter search a few times.
Please, use `{ [weak object] in ... ` in each case throughout the app.

PR #841, opened 2018-04-19
Cell configurator for dapp cells

marijnschilling:
fixed bug IOS-605: https://toshiapp.atlassian.net/browse/IOS-605
By giving a minimum height to the desciptionlabel, so when there's no or a shorter descirption the cell will remain the same height.

Incoorporated the Rect image cell thingy in our CellConfigurator.

Also made the separators on the BasicTableViewCell more flexible.

Made the collectibles cell according to design with a new cell type `CollectibleCell`
here you see how the cells behave when different fields are missing:
img_e7ebc0b4109b-1


Toshi/Controllers/Wallet/CollectibleViewController.swift
```diff
@@ -22,7 +22,7 @@ final class CollectibleViewController: UIViewController {
         view.translatesAutoresizingMaskIntoConstraints = false
 
         view.backgroundColor = nil
-        view.register(RectImageTitleSubtitleTableViewCell.self)
+        BasicTableViewCell.register(in: view)
```
yuliaveres: Please, make sure Collectibles list looks according to the design.
img_0545

But should be:
img_0544

Toshi/Controllers/Dapps/DappsListViewController.swift
```diff
@@ -100,16 +99,12 @@ final class DappsListViewController: UITableViewController {
...
-        cell.titleLabel.text = dapp.displayTitle
-        cell.subtitleLabel.text = dapp.displayDetails
-        cell.imageViewPath = dapp.itemIconPath
-        cell.leftImageView.layer.cornerRadius = 5
-
-        cell.selectionStyle = .default
-        cell.accessoryType = .disclosureIndicator
-        cell.subtitleLabel.numberOfLines = 2
+        configurator.configureCell(cell, with: cellData)
+        cell.showSeparator(leftInset: 100, rightInset: 15)
```
designatednerd: worth updating `15` to `.marginx3`, here and in other places in the PR?

Toshi/Views/Cells/AvatarTitleDescriptionCell.swift
```diff
@@ -0,0 +1,66 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+final class AvatarTitleDescriptionCell: BasicTableViewCell {
+    static let rectImageHeight: CGFloat = 72.0
```
yuliaveres: is not it same as `BasicTableViewCell.largeImageSize ` ?

Toshi/Views/Cells/AvatarTitleDescriptionCell.swift
```diff
@@ -0,0 +1,66 @@
...
+        contentView.addSubview(descriptionLabel)
+        contentView.addSubview(titleTextField)
+
+        setupLeftImageView()
+        setupTitleTextField()
+        setupDescriptionLabel()
+    }
+
+    private func setupLeftImageView() {
+        leftImageView.size(CGSize(width: AvatarTitleDescriptionCell.rectImageHeight, height: AvatarTitleDescriptionCell.rectImageHeight))
```
yuliaveres: am I wrong or `AvatarTitleDescriptionCell ` and `CollectibleCell` differ only by having / or not having subtitle ?
in this case would not it be reasonable to delete `AvatarTitleDescriptionCell ` class and use `CollectibleCell ` in all cases having title, subtitle and description in a vertical resizable stack view (as title and subtitle were in `RectImageTitleSubtitleTableViewCell` ) ?   
( also,  is `AvatarTitleDescriptionCell ` used somewhere ? )

marijnschilling: This will just require me adding a new imageView and the new pathcomponents and image, which requires quite some unnecessary duplication... thats why i chose this option

Toshi/Views/Cells/AvatarTitleDescriptionCell.swift
marijnschilling: yup!

PR #842, opened 2018-04-20
Fix bottom margin on Payment Confirmation VC

designatednerd:
Also fixed a couple minor things I noticed in the process of trying to fix this one. 

Before: 
img_0142

After: 
simulator screen shot - iphone 8 plus - 2018-04-19 at 17 04 19


Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -196,6 +196,11 @@ extension ChatsViewController: UITableViewDelegate {
         }
     }
 
+    /// WORKAROUND: For some reason the swift 4.1 compiler is convinced that `canEditRowAt` should be this
+    /// method instead. Overridding and returning the documented default value to silence the warning.
```
yuliaveres: 🤦‍♀️

designatednerd: god, i know

PR #843, opened 2018-04-20
[IOS-606]: Move to v2 id service and from TokenUser to Codable Profile

yuliaveres:


Toshi/APIClients/idAPIClient.swift
```diff
@@ -156,20 +151,20 @@ final class IDAPIClient {
...
+        guard let user = Profile.current, user.paymentAddress != Cereal.shared.paymentAddress else {
             return
         }
 
-        var userDict = user.dict
-        userDict[TokenUser.Constants.paymentAddress] = Cereal.shared.paymentAddress
+        guard var userDict = user.dictionary else { return }
+        userDict[Constants.paymentAddress] = Cereal.shared.paymentAddress
 
         updateUser(userDict) { _, _ in }
```
designatednerd: Since we're mucking around in here - if we're not actually going to use this closure we should probably make it nullable

Toshi/Models/Converters/EthereumConverter.swift
```diff
@@ -124,8 +124,8 @@ struct EthereumConverter {
     /// - Parameter balance: the value in wei
     /// - Returns: the fiat currency value with redundant 3 letter code for clarity.
     static func fiatValueStringWithCode(forWei balance: NSDecimalNumber, exchangeRate: Decimal) -> String {
-        let locale = TokenUser.current?.cachedCurrencyLocale ?? Currency.forcedLocale
-        let localCurrency = TokenUser.current?.localCurrency ?? Currency.forcedLocale.currencyCode
+        let locale = Profile.current?.cachedCurrencyLocale ?? Currency.forcedLocale
+        let localCurrency = Profile.current?.localCurrency ?? Currency.forcedLocale.currencyCode
```
designatednerd: might be helpful to do this with a `guard let` to access the current profile

Toshi/Models/Users/ProfilesManager.swift
```diff
@@ -0,0 +1,117 @@
...
+        super.init()
+
+        fetchContactsFromDataBase()
+    }
+
+    func profile(for toshiId: String) -> Profile? {
+        return profiles.first(where: {  $0.toshiId == toshiId })
+    }
+
+    func refreshProfile(_ profile: Profile) {
```
designatednerd: i think it might be better to name this something like `updateProfile` since it adds/updates the given profile rather than doing anything that looks like refreshing it from the database

Toshi/Models/Users/ProfilesManager.swift
```diff
@@ -0,0 +1,117 @@
...
+        guard let existingProfileIndex = profiles.index(where: { $0.toshiId == profile.toshiId }) else {
+            profiles.append(profile)
+            return
+        }
+
+        profiles[existingProfileIndex] = profile
+    }
+
+    @objc public func refreshProfiles() {
```
designatednerd: i'd go with something like `clearProfiles` for this one for similar reasons

Toshi/Models/Users/ProfilesManager.swift
```diff
@@ -0,0 +1,117 @@
...
+        profiles = []
+    }
+
+    @objc func profilesIds() -> [String] {
+        return profiles.map { $0.toshiId }
+    }
+
+    func fetchContactsFromDataBase() {
```
designatednerd: total nit: `DataBase` -> `Database` since it's one word, not two

Toshi/Models/Users/ProfilesManager.swift
```diff
@@ -0,0 +1,117 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+
+@objc public class ProfilesManager: NSObject {
+
+    private(set) var profiles: [Profile] = []
+    private(set) var contactIdToSignalRecipientsMap: [String: [SignalRecipient]] = [:]
```
designatednerd: Since this is not java, I think `map` might be a bit of a misnomer for what this is

Toshi/Models/Users/ProfilesManager.swift
```diff
@@ -0,0 +1,117 @@
...
+    func fetchContactsFromDataBase() {
+
+        guard let contactsData = yap.retrieveObjects(in: storedContactKey) as? [Data] else {
+
+            return
+        }
+
+        profiles.removeAll()
+
+        for data in contactsData {
```
designatednerd: so this is going through and putting anything we have cached info for straight into memory. what happens if the user has a ton of contacts?

Toshi/Models/Users/TokenUser.swift
```diff
@@ -26,423 +26,29 @@ extension NSNotification.Name {
...
-    }
-
-    init?(coder aDecoder: NSCoder) {
-        if let apps = aDecoder.decodeObject(forKey: "objects") as? [TokenUser] {
-            self.objects = apps
-        }
-
-        super.init()
-    }
+struct Constants {
```
designatednerd: Since these are no longer a sub-object of something else, you may want to consider a different name since they're constants specifically related to something. 

Alternately, consider putting these and the free-floating lets in this file into some kind of enum or something where you can see what they're for, like `LegacyUser.Constants.name`

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -69,7 +69,7 @@ enum ProfileType: Int {
 }
 
 /// An individual Profile
-struct Profile: Codable {
+struct Profile: Codable, Hashable {
```
designatednerd: can you please pull the `Hashable` conformance into its own extension? Also, since you override `==`, this can also be declared to conform to `Equatable` (although that conformance should probably also be in its own extension)

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -102,6 +102,14 @@ struct Profile: Codable {
         reviewCount = "review_count"
     }
 
+    var localCurrency2: String {
```
designatednerd: why the `2`?

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -147,11 +159,84 @@ struct Profile: Codable {
         }
     }
 
+    static  func retrieveCurrentUserFromStore() -> Profile? {
```
designatednerd: extra space between `static` and `func`

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -147,11 +159,84 @@ struct Profile: Codable {
...
+        var profile: Profile?
+
+        // migrate old user storage
+        if _current == nil, let userData = (Yap.sharedInstance.retrieveObject(for: legacyStoredUserKey) as? Data) {
```
designatednerd: you may want to consider going a `guard` on `_current == nil` and bailing out on that so you don't have to keep checking for it

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -147,11 +159,84 @@ struct Profile: Codable {
...
+                assertionFailure("Can't decode retrieved current user data")
+            }
+
+            // migrations
+            var shouldSaveMigration = false
+            if userSettings[Constants.verified] == nil {
+                if json[Constants.verified] != nil {
+                    userSettings[Constants.verified] = json[Constants.verified]
+                } else {
+                    userSettings[Constants.verified] = 0
```
designatednerd: should this be `false` instead of `0`?

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -147,11 +159,84 @@ struct Profile: Codable {
...
+            var userSettings = Yap.sharedInstance.retrieveObject(for: Cereal.shared.address, in: localUserSettingsKey) as? [String: Any] ?? [:]
+
+            // Because of payment address migration, we have to override the stored payment address.
+            // Otherwise users will be sending payments to the wrong address.
+            if json[Constants.paymentAddress] as? String != Cereal.shared.paymentAddress {
+                json[Constants.paymentAddress] = Cereal.shared.paymentAddress
+
+                if let updatedData = try? JSONSerialization.data(withJSONObject: json, options: []) {
+                    userData = updatedData
+                }
```
designatednerd: should updating this also cause the changes to save?

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -147,11 +159,84 @@ struct Profile: Codable {
...
+    var userInfo: UserInfo {
+        return UserInfo(address: toshiId, paymentAddress: paymentAddress, avatarPath: avatar, name: nameOrDisplayName, username: displayUsername, isLocal: true)
+    }
+
+    var data: Data? {
+        return try? JSONEncoder().encode(self)
+    }
+
+    static var defaultCurrency: String {
+        return Locale.current.currencyCode ?? "USD"
```
designatednerd: what happens here if the locale currency is one we don't support?

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -161,18 +246,57 @@ struct Profile: Codable {
...
+
     static func retrieveCurrentUser() {
-        guard let user = TokenUser.retrieveCurrentUserFromStore() else { return }
-        let jsonDecoder = JSONDecoder()
-        guard let decodedProfile = try? jsonDecoder.decode(Profile.self, from: user.json) else { return }
-        current = decodedProfile
+        guard let user = Profile.retrieveCurrentUserFromStore() else { return }
+        current = user
+
+        NotificationCenter.default.post(name: .userCreated, object: nil)
```
designatednerd: why is this `userCreated` rather than something like `retrievedUser`

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -161,18 +246,57 @@ struct Profile: Codable {
...
+        current = profile
+
+        Yap.sharedInstance.setupForNewUser(with: profile.toshiId)
+
+        let newUserSettings: [String: Any] = [
+            Constants.localCurrency: Profile.defaultCurrency,
+            Constants.verified: 0
+        ]
+
+        current?.userSettings = Yap.sharedInstance.retrieveObject(for: Cereal.shared.address, in: localUserSettingsKey) as? [String: Any] ?? newUserSettings
```
designatednerd: might be worth unwrapping `current` before proceeding here

Toshi/Models/UsersBotsGroups/Profile.swift
```diff
@@ -161,18 +246,57 @@ struct Profile: Codable {
...
+        current?.saveSettings()
+        current?.adjustToLocalCurrency()
 
         NotificationCenter.default.post(name: .userCreated, object: nil)
```
designatednerd: this looks like the user could either be created or retrieved, but maybe i'm missing something here.

Toshi/Models/Yap.swift
```diff
@@ -187,7 +187,7 @@ final class Yap: NSObject, Singleton {
     }
 
     private func useBackedDBIfNeeded() {
```
designatednerd: old nit but while we're fixing things: `useBackedDB` -> `useBackupDB`

Toshi/Controllers/GroupChat/GroupViewController.swift
```diff
@@ -186,12 +186,13 @@ final class GroupViewController: UIViewController {
     }
 
     private func showUserInfo(with userId: String) {
-        guard let currentUser = TokenUser.current else { return }
-        var users = SessionManager.shared.contactsManager.tokenContacts.filter { $0.address == userId }
-        users.append(currentUser)
+        guard let currentUser = Profile.current else { return }
+        var profiles = SessionManager.shared.profilesManager.profiles.filter { $0.toshiId == userId }
```
designatednerd: This seems like something that might be better hidden under the cover of the profile manager - that way if we are able to do some upgrades we can fetch as needed from the database rather than filtering an in-memory list of profiles

Toshi/APIClients/idAPIClient.swift
```diff
@@ -255,7 +262,7 @@ final class IDAPIClient {
 
                     if let path = userDict["avatar"] as? String {
                         AvatarManager.shared.refreshAvatar(at: path)
-                        TokenUser.current?.update(avatar: avatar, avatarPath: path)
+                        //Profile.current?.update(avatar: avatar, avatarPath: path)
```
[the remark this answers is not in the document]
yuliaveres: just WIP, am on all of these now

Toshi/Controllers/TabBarController.swift
```diff
@@ -155,10 +155,10 @@ class TabBarController: UITabBarController, OfflineAlertDisplaying {
         if url.user == "username" {
             guard let username = url.host else { return }
 
-            idAPIClient.retrieveUser(username: username) { [weak self] contact in
-                guard let contact = contact else { return }
-
-                let contactController = ProfileViewController(profile: contact)
+            idAPIClient.retrieveUser(username: username) { [weak self] profile, _ in
+                guard let profile = profile else { return }
```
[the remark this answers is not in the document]
yuliaveres: okay ..

Toshi/Extensions/TSThread+Addition.swift
```diff
@@ -18,20 +18,21 @@ import Foundation
...
+        if let userData = retrievedData {
 
-            recipient = TokenUser(json: json, shouldSave: false)
+            do {
+                let jsonDecoder = JSONDecoder()
+                recipient = try jsonDecoder.decode(Profile.self, from: userData)
+            } catch { }
         } else {
-            recipient = SessionManager.shared.contactsManager.tokenContacts.first(where: { $0.address == recipientAddress })
+            recipient = SessionManager.shared.profilesManager.profiles.first(where: { $0.toshiId == recipientAddress })
```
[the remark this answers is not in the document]
yuliaveres: this does not pull out of db here, they are fetched from there once, at launch + if updated; - these are in memory collection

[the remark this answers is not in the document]
designatednerd: yeah i think i noticed that later in the review but couldn't update this comment since I'd left too many :P 

But it's definitely worth centralizing this in the manager so that if we do wind up with something where we *don't* have to pull everything into memory, we can do that transparently to this caller.

Toshi/Extensions/TSThread+Addition.swift
```diff
@@ -52,7 +53,7 @@ extension TSThread {
     func updateGroupMembers() {
         if let groupThread = self as? TSGroupThread {
 
-            let contactsIDs = SessionManager.shared.contactsManager.tokenContacts.map { $0.address }
+            let contactsIDs = SessionManager.shared.profilesManager.profiles.map { $0.toshiId }
```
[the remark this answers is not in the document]
yuliaveres: we need just IDs collection here .. what do you mean by "another way to access the profile via its ID" ? no need to access a whole profile here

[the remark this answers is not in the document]
designatednerd: Ah, I missed that. I was thinking about pulling a profile out of the database using its ID, but that's not what we're looking for here. 

However, this is another good candidate for putting the logic under the hood so in the future if we can say "hey database, fetch me all these things and give me back just their IDs", we do that under the hood.

Toshi/Controllers/Profiles/ProfilesViewController.swift
```diff
@@ -250,8 +252,8 @@ extension ProfilesViewController: UITableViewDataSource {
         }
 
         let tableData = TableCellData(title: profile.name,
-                subtitle: profile.isApp ? profile.descriptionForSearch : profile.username,
-                leftImagePath: profile.avatarPath,
+                subtitle: profile.type != ProfileType.user.typeString ? profile.description : profile.username,
```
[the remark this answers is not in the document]
yuliaveres: possible

Toshi/Models/Users/ProfilesManager.swift
yuliaveres: it is not an issue, and it is not something new introduced here.
no need to have even more unknown Yap issues by increasing DB access so much. It is worth considering as soon as we move to more reliable db. now i am going to leave the fetch in a way as it was since the beginning.

Toshi/Models/UsersBotsGroups/Profile.swift
yuliaveres: why retrieved instead of created - it exists for both cases. no need to change this name here

Toshi/Models/Users/ProfilesManager.swift
designatednerd: ok, makes sense - we'll revisit this decision if we're able to stop using Yap

Toshi/Models/UsersBotsGroups/Profile.swift
designatednerd: ^still wondering here

Toshi/Models/UsersBotsGroups/Profile.swift
designatednerd: Because to me i would expect different notifications for "created" (implying a new user who hasn't used the app before has been created) vs "retrieved" (implying a user who already existed was retrieved from the database)

Toshi/Models/Users/ProfilesManager.swift
Elland: when* 🤞🏽😍

Toshi/Controllers/TabBarController.swift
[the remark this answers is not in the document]
designatednerd: meaning instead of `guard let profile = profile` something like `deeplinkedProfile = profile`

Toshi/Models/UsersBotsGroups/Profile.swift
yuliaveres: we won;t have conversion rate in this case. we should implement logic of checking locale currency code agains known currencies list.

Toshi/Models/UsersBotsGroups/Profile.swift
yuliaveres: 🤷‍♂️

Toshi/Models/UsersBotsGroups/Profile.swift
designatednerd: Can you make a Jira ticket for that please? We'll deal with it in a different PR

Toshi/Models/UsersBotsGroups/Profile.swift
designatednerd: I guess a better way to come at the question is: What happens when `userCreated` fires?

yuliaveres: it is easy to find in the code. if someone has idea about renaming existing code - s/he can go ahead and submit a pr. this pr contains huge amount of work without naming changes.

designatednerd: fair enough, i'll make myself a note to take a look

yuliaveres:
@designatednerd do you have other thoughts on this ?

designatednerd:
Will take a look through 👍

Toshi/Controllers/Profile/ProfileViewController.swift
```diff
@@ -442,9 +438,9 @@ final class ProfileViewController: DisappearingNavBarViewController {
     
     private func configureForCurrentProfile() {
         // Set nils for empty strings to make the views collapse in the stack view
-        nameLabel.text = profile.name.isEmpty ? nil : profile.name
-        aboutContentLabel.text = profile.about.isEmpty ? nil : profile.about
-        locationContentLabel.text = profile.location.isEmpty ? nil : profile.location
+        nameLabel.text = profile.name?.isEmpty ?? true ? nil : profile.name
```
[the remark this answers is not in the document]
designatednerd: I saw you made a `contentsOrEmpty` helper method used in other places - wanna use it here?

Toshi/Controllers/Settings/Models/ProfileEditItem .swift
```diff
@@ -68,23 +68,23 @@ class ProfileEditItem {
             detailText = user.username
         case .displayName:
             titleText = Localized.edit_profile_display_name_label
-            detailText = user.name
+            detailText = user.nameOrDisplayName
         case .about:
             titleText = Localized.edit_profile_about_label
-            detailText = user.about
+            detailText = String.contentsOrEmpty(for: user.description)
```
designatednerd: If you want to display an empty string if `user.description` is `nil`, perfecto. If you want to not display anything and hide the detail text view, you should just be able to do `user.description` and have the `detailText` set as `nil`

Toshi/Controllers/TabBarController.swift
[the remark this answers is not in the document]
designatednerd: [this can stay]

Toshi/Extensions/String+Additions.swift
```diff
@@ -183,4 +183,8 @@ extension String {
 
         return firstChars + truncationString + lastChars
     }
+
+    static func contentsOrEmpty(for string: String?) -> String {
+        return string ?? ""
```
designatednerd: 🌠 woooo!

Toshi/Controllers/Settings/Models/ProfileEditItem .swift
yuliaveres: yes, designed to be an empty string if nothing is there.

PR #845, opened 2018-04-23
iOS-621: Use Secure Text Entry

designatednerd:
This disables the use of third-party keyboards and invalidates iOS-342, where the text field would get deselected when entering emojis, because it also disables the use of the emoji keyboard 👏 🔑 

On my phone, an app with third party keyboard and emoji: 

img_2833

And trying to use the sign in page after making this change: 

img_2831


marijnschilling:
yay! Good fix, let's point to some official docs from here just for reference ✌️

designatednerd:
From the [Custom Keyboard guide](https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/CustomKeyboard.html): 

>There are certain text input objects that your custom keyboard is not eligible to type into. First is any secure text input object. Such an object is defined by its secureTextEntry property being set to YES and is distinguished by presenting typed characters as dots.
>
>When a user taps in a secure text input object, the system temporarily replaces your custom keyboard with the system keyboard. When the user then taps in a nonsecure text input object, your keyboard automatically resumes.

PR #846, opened 2018-04-23
carthage update

designatednerd:
- EtherealCereal to 0.3
- Nimble to 7.1.1
- Quick to 1.3.0


designatednerd:
Just ran `carthage update` since we have a pretty significant bugfix in `EtherealCereal`. No other code changed.

designatednerd:
same thing as before with circle not being able to figure out where the iPhone 7 sim is. I kicked the build so hopefully that'll pass

designatednerd:
Gonna leave the resetting sims in there until the Circle folks figure out what's going on there

PR #847, opened 2018-04-23
IOS-586: ALLOW SHOUTY-CASE URL INPUT

designatednerd:
Note that when a redirect fires, it will eventually wind up lowercased. But until then your URL will SHOUT AS LOUD AS YOU WANT IT TO: 

simulator screen shot - iphone 8 plus - 2018-04-23 at 17 09 12


designatednerd:
Update: I was wrong!

PR #849, opened 2018-04-24
Mock collectibles

marijnschilling:
When in debug I create a mockTeapot and fetch the collectible and collectibles on that one. 

I wonder if this debug check i made is safe and clear enough, or should it be something that should be able to enable?

Right now the images don't load, I'm not sure why not. Might be the same reason that the MessageRequest images aren't loading, so will take that on on the similar issue.


Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -208,7 +208,14 @@ final class EthereumAPIClient {
     }
 
     func getCollectible(address: String = Cereal.shared.paymentAddress, contractAddress: String, completion: @escaping ((Collectible?, ToshiError?) -> Void)) {
-        self.activeTeapot.get("/v1/collectibles/\(address)/\(contractAddress)") { result in
+        var collectiblesTeapot = self.activeTeapot
+
+        // If we are on debug we will mock out the collectible data since they are only on production
```
designatednerd: what if we're debugging and we actually WANT production? Would it make sense to check both that and the bundle ID is not the prod bundle?

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -208,7 +208,14 @@ final class EthereumAPIClient {
...
+        #if DEBUG
+        collectiblesTeapot = MockTeapot(bundle: Bundle(for: EthereumAPIClient.self), mockFilename: "getACollectible")
```
designatednerd: can you please indent this one more tab for readability?

Toshi/APIClients/EthereumAPIClient.swift
designatednerd: [Nevermind, `#DEBUG` is an `other swift flags` var that's passed in so this comment is not relevant]

marijnschilling: Should i maybe explain this better in the comment?

designatednerd: Couldn't hurt

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -208,7 +208,14 @@ final class EthereumAPIClient {
     }
 
     func getCollectible(address: String = Cereal.shared.paymentAddress, contractAddress: String, completion: @escaping ((Collectible?, ToshiError?) -> Void)) {
-        self.activeTeapot.get("/v1/collectibles/\(address)/\(contractAddress)") { result in
+        var collectiblesTeapot = self.activeTeapot
+
+        // If we are on debug (specified in "other swift flags") we will mock out the collectiblesince they are only on production
```
designatednerd: [optional] `collectiblesince` -> `collectiblesSince`

PR #850, opened 2018-04-24
[IOS-588]: Store preferred switched network in user defaults instead of keychain

yuliaveres:
Issue caused the preference to affect different target app. There is no need to store current network in the keychain as well.


designatednerd:
@yuliaveres Please pull in recent `master` - there's now a workaround for this issue where Circle doesn't think the iPhone 7 exists. Still working on a permanent fix but at least this should get it up and running temporarily.

designatednerd:
@yuliaveres Indeed: 

screen shot 2018-04-25 at 9 36 59 am

designatednerd:
Tests pass locally for me, I'm fine if we merge this while Circle gets its shit together

PR #853, opened 2018-04-26
[IOS-599] keyboard adjust send token view

marijnschilling:
I put everything on a scrollView and implemented our `KeyboardAdjustable` protocol to adjust that scrollView.

keyboard-adjust


Toshi/Controllers/Wallet/SendToken/SendTokenViewConfigurator.swift
```diff
@@ -55,6 +55,12 @@ final class SendTokenViewConfigurator: NSObject {
         }
     }
 
+    lazy var scrollView: UIScrollView = {
+        let scrollView = UIScrollView()
```
designatednerd: It might be simpler just to do this as `lazy var scrollView = UIScrollView()` since you're not setting anything on this instance

Toshi/Controllers/Wallet/SendToken/SendTokenViewController.swift
```diff
@@ -138,3 +143,25 @@ extension SendTokenViewController: TokenSendConfirmationDelegate {
...
+// MARK: - Keyboard Adjustable
+
+extension SendTokenViewController: KeyboardAdjustable {
+
+    var keyboardWillShowSelector: Selector {
+        return #selector(keyboardShownNotificationReceived(_:))
+    }
+
+    var keyboardWillHideSelector: Selector {
+        return #selector(keyboardHiddenNotificationReceived(_:))
```
designatednerd: it looks like all these selectors are doing are passing things straight through to `keyboardWillShow` and `keyboardWillHide` - why not add those as the selectors instead?

marijnschilling: because it's in a protocol :D

marijnschilling: Or wait, thats terrible reasoning. I mean, it's a protocol that is used all over the app, that is set up like this i believe. So we cannot call the `keyboardWillShow(_ notification: NSNotification` directly cause we don't get a notification here... I'm not 100% on top of the inner workings of this protocol anymore. Might be worth looking at, but since it's used in a ot of places all over the app in this way, might not be for this PR.

designatednerd: Yeah i was having REAL weird issues with notifications in a protocol as well - let's make a ticket to come back to this

marijnschilling: Issue created: https://toshiapp.atlassian.net/browse/IOS-646

marijnschilling:
Tests fail, but i don't have time to look into that now 🙈

designatednerd:
Test failures are from looking at the wrong cached data when testing - pull in master and they should be fixed

PR #856, opened 2018-04-30
[IOS-644]: Fix crash on table view cell selection after profiles search returned an error

yuliaveres:
https://fabric.io/bakken--bck2/ios/apps/org.toshi.distribution/issues/5ae677fe638393737ab76631?time=last-seven-days

call completion and trigger table view reload if search profiles returned an error;
add percent encoding for search profile text;
update collectible tests to match existing mocked response.


Toshi/APIClients/IDAPIClient.swift
```diff
@@ -813,9 +813,12 @@ final class IDAPIClient {
 
     func createPathFor(_ type: String, with searchText: String?) -> String {
         var path = "/v2/search?type=\(type)"
-        if let query = searchText {
-            path.append("&query=\(query)")
-        }
+
+        guard var searchedText = searchText else { return path }
+
+        searchedText = searchedText.addingPercentEncoding(withAllowedCharacters: IDAPIClient.allowedSearchTermCharacters) ?? searchedText
```
designatednerd: oof, good catch

PR #857, opened 2018-04-30
[IOS-614]: Change avatars cache access API

yuliaveres:
Before:
img_0574
img_0575

Now:
img_0572
img_0573


Toshi/Avatars/AvatarManager.swift
```diff
@@ -152,16 +152,19 @@ final class AvatarManager: NSObject {
...
     ///   - found image or nil if not present
-    @objc func cachedAvatar(for key: String, completion: @escaping ((UIImage?) -> Void)) {
-        guard !key.hasAddressPrefix else {
-            guard let avatarPath = UserDefaults.standard.object(forKey: key) as? String else { return }
-            cachedAvatar(for: avatarPath, completion: completion)
 
-            return
+    @objc func cachedAvatar(for key: String) -> UIImage? {
+        guard !key.hasAddressPrefix else {
+            guard let avatarPath = UserDefaults.standard.object(forKey: key) as? String else { return nil }
```
designatednerd: probably a longer term question to examine here is why we're caching the paths in user defaults instead of just letting Haneke do this work for us, but does not need to be addressed here. 

Can you make a ticket for it though?

Toshi/Avatars/AvatarManager.swift
```diff
@@ -152,16 +152,19 @@ final class AvatarManager: NSObject {
...
+            return cachedAvatar(for: avatarPath)
         }
 
-        cache.fetch(key: key).onSuccess { image in
-            completion(image)
+        var image: UIImage?
+
+        cache.fetch(key: key).onSuccess { cachedImage in
```
designatednerd: is this a synchronous call?

yuliaveres: yep, that what i was surprised by.

Toshi/Avatars/AvatarManager.swift
yuliaveres: so the thing with different path is that we might want to retrieve an image by address / toshi id (`0xkhd...`) or by avatar path itself. The most frequent case is requesting by url path. (if avatar was changed - path was changed but address is going to be the same always.) But sometimes avatar is being requested by toshi id, i guess we should rely more on profile manager for this and retrieve profile by toshi id, then request an avatar by url. will add ticket.

PR #859, opened 2018-04-30
IOS-827 pt 1: Network Status Bar

designatednerd:
simulator screen shot - iphone x - 2018-04-30 at 12 21 13

Part 1 of (probably) 2: 

Here: 
- Updated `ActiveNetworkShowing` so it shows from the bottom of the view if you're on a non-main network, regardless of what build you're on. 
- Cleaned up a bunch of the logic around the protocol so as much code as possible is centralized
- Made `ActiveNetworkShowing` deal with logic around updating from notifications
- Anal-retentively organized the `xcodeproj` file
- Deleted a few things I found weren't being used while doing said organizing

Still to come: 
- Getting this banner to show up on way more bits of the app than the two places it already was.


Toshi/Controllers/AlertController/AlertController.swift
```diff
@@ -1,162 +0,0 @@
...
-
-    init(title: String, titleColor: UIColor = UIColor.lightGray, icon: UIImage? = nil, block: @escaping ActionBlock) {
-        self.title = title
-        self.titleColor = titleColor
-        self.icon = icon
-        self.block = block
-    }
-}
-
-class AlertController: ModalPresentable {
```
marijnschilling: 👋 bye bye! Nice cleanup ✌️

Toshi/Controllers/Chats/ChatsViewController.swift
```diff
@@ -68,7 +68,6 @@ final class ChatsViewController: SweetTableController {
         super.viewWillAppear(animated)
 
         preferLargeTitleIfPossible(target.prefersLargeTitle)
-        tabBarController?.tabBar.isHidden = false
```
marijnschilling: I wonder how you found out we can remove this, do you know what it did and why it's unnecessary now?

Toshi/Protocols/ActiveNetworkDisplaying.swift
```diff
@@ -20,35 +20,42 @@ import SweetUIKit
...
     func showActiveNetworkViewIfNeeded()
     func hideActiveNetworkViewIfNeeded()
 
+    func removeNotificationObservers()
+
     func requestLayoutUpdate()
 }
 
 extension ActiveNetworkDisplaying where Self: UIViewController {
-    func setupActiveNetworkView(hidden: Bool = false) {
```
marijnschilling: So, do we always show the active network view now? We used to have this since we'd only show the view if we weren't on main net.

Toshi/Protocols/ActiveNetworkDisplaying.swift
```diff
@@ -57,17 +64,48 @@ extension ActiveNetworkDisplaying where Self: UIViewController {
     }
 
     func hideActiveNetworkViewIfNeeded() {
-        guard !NetworkSwitcher.shared.isDefaultNetworkActive else { return }
-
-        DispatchQueue.main.async {
-            self.activeNetworkView.heightConstraint?.constant = 0
-            self.requestLayoutUpdate()
+        guard self.activeNetworkView.heightConstraint?.constant != 0 else {
```
marijnschilling: Just a little flowery comment here 🌸 ✨ 🤗  I would put this bool in a named let so you don't need the comment, so there's less chance for it to get outdated. so just `let isActiveNetworkViewHidden = blabla != 0` and then guard on that let.

Toshi/Protocols/ActiveNetworkDisplaying.swift
```diff
@@ -57,17 +64,48 @@ extension ActiveNetworkDisplaying where Self: UIViewController {
...
+        notificationObservers.append(observer)
+    }
+
+    func removeNotificationObservers() {
+        for observer in notificationObservers {
+            NotificationCenter.default.removeObserver(observer)
+        }
+    }
+
+    private func updateActiveNetworkView() {
```
marijnschilling: Great to pull this in here!

Toshi/Controllers/Chats/ChatsViewController.swift
designatednerd: It's because we've already set `hidesBottomBarOnPush` to true. I was trying to futz with the animation and then discovered that manually showing and hiding the bar is not necessary if you set that property.

Toshi/Protocols/ActiveNetworkDisplaying.swift
designatednerd: No - `updateActiveNetworkView`, which is called as part of `setupActiveNetworkView`, will automatically hide the network view if you're on mainnet and show it if you're not.

designatednerd:
Merging after slack 👍 on new changes from @marijnschilling

PR #860, opened 2018-04-30
IOS-368: Can't enter passphrase on iPhone SE with gigantic dynamic text on

designatednerd:
We were only looking for visible cells to get selected index paths rather than just accessing the selected index paths directly, so when on a tiny screen with giant font, the cells would stop being visible, and you couldn't type anything. 

Now: 
simulator screen shot - iphone se - 2018-04-30 at 13 34 24


designatednerd:
it appears the latter - they were showing up on things bigger than the SE, but not on the SE.

PR #861, opened 2018-04-30
Add Wallet section to settings view

marijnschilling:
Small PR for this little addition to settings:
img_b51afc7bbf05-1

Next step is to add the selection view ✌️ 

I used a custom cell in our configurator structure for this celltype. The `InputCell` that is (mis)used for the Balance will probably be removed anyways ✌️


Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -348,6 +367,9 @@ extension SettingsController: UITableViewDelegate {
             let profileVC = ProfileViewController(profile: current, readOnlyMode: false)
             
             self.navigationController?.pushViewController(profileVC, animated: true)
+        case .wallet:
+            print("wallet cell selected")
```
designatednerd: Can you do this as a `DLog` so we at least get some context about where and when this was called?

Toshi/Controllers/Settings/Wallet/WalletCell.swift
```diff
@@ -0,0 +1,42 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import UIKit
+
+final class WalletCell: BasicTableViewCell {
```
designatednerd: why do this as a specific cell rather than like a `Title Value cell` or something more generic?

Toshi/Controllers/Settings/SettingsController.swift
marijnschilling: oh yeah, this is just untill i'ma implementing it right in my next PR 😬  

Can't have switch case wihtout anything, and wanted to break this PR into this small piece ✌️

Toshi/Controllers/Settings/Wallet/WalletCell.swift
marijnschilling: Because I think we should only implement these generic cells where it's necessary (that will be when we actually need the cells in with those components in more than one place), it's a lot of work to add one of these to the CellConfigurator. 

I also created a new celltype for `CollectibleCell` like that. 

I think it will make our CellConfigurator system a bit more flexible. Because now it has happened that just subclasses of default UITableViewCells were used since it didnt fit in our CellConfigurator that easily. I think this is a good middleground where you can still make use of the benefits of the `BasicTableViewCell` but not necessarily with allll the hassle of creating a unique cell component combination. 

But i am in doubt wether this actual cell should be a `TitleValueCell` I could add the value as a component 🤔 

For the collectible this wasn't possible it would mean i had to add a `squareImage` and a `squareImagePath`and double up all the setup around those 🙈

PR #862, opened 2018-04-30
IOS-435: Passphrase view occasionally freaks out when ending the first with a three-letter word

designatednerd:
OK, I finally got some (occasionally) reproducible steps from @marijnschilling for this on iOS 11, and I *think* i got it.

Basically, what it appears is happening is that something's getting reloaded and causing accepted words to go back to what the user typed rather than the accepted match.

Secondarily, matches were not always checked before returning the cell if the user hadn't pasted a passphrase, which, as far as I can tell, is what was causing three-letter words to randomly start to error out and cause everything else to reload, fucking up everything.  

Also while i was in here, I renamed `caretView` and its associated bits to `cursorView` to more accurately reflect what it is and does because I wasn't sure if it had anything to do with this before that.


Toshi/Controllers/Chat/ChatMenuTableViewController.swift
```diff
@@ -40,7 +40,7 @@ final class ChatMenuTableViewController: UITableViewController {
         var sourceFrame = superview.convert(sourceView.frame, from: superview)
         
         if let cell = sourceView as? SignInCell {
-            let origin = cell.caretView.convert(cell.caretView.frame.origin, to: cell)
```
marijnschilling: I think it should be `caret` but cursor is better! 🥕

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -293,6 +295,10 @@ extension SignInViewController: UICollectionViewDataSource {
                 cell.setText(text, isFirstAndOnly: itemCount == 1)
             }
 
+            if !userDidPastePassphrase {
```
marijnschilling: Seems like we can better change this statement around
```
if userDidPastePassphrase { //do things with the cell }
return cell
```

to avoid bad placement of this return statement in the future.

Toshi/Controllers/Chat/ChatMenuTableViewController.swift
designatednerd: Maybe it's a 🇺🇸  English thing - to me a `caret` is something that looks like `^` or `>`, while a cursor is the blinky thing you type with. Since this is a blinky thing, `caret` threw me off because I was looking for a totally different view than was actually there.

Toshi/Controllers/Sign In/SignInViewController.swift
designatednerd: hmm... let me take another swipe at changing this to be a little easier to understand.

Toshi/Controllers/Chat/ChatMenuTableViewController.swift
designatednerd: also this is a good follow up to our discussion of the 800 words for 🥕  in 🇳🇱 at lunch 😆

designatednerd:
yah thanks for calling me on that - i was trying to keep the code footprint small to the detriment of actually making it clearer.

PR #636, opened 2017-12-08
Mark incoming message as read if it comes through the current chat.

Elland:
There must be a better way.

PR #864, opened 2018-04-30
Fix for incoming message not getting marked as read

designatednerd:
This PR isolates @Elland's fix from #636. 

I would like both @Elland and @yuliaveres's eyes on this before I merge it so I can make sure i'm not missing something goofy with Yap.


designatednerd:
If you could re-verify the change I would much appreciate it.

designatednerd:
Moving to another PR

PR #865, opened 2018-05-01
[IOS-626] Remove qrcode and balance from settings

marijnschilling:
They are bye bye:
img_89266dc2ecc2-1


Toshi/Controllers/Settings/Balance/BalanceController.swift
```diff
@@ -1,174 +0,0 @@
...
-
-    private lazy var tableView: UITableView = {
-        let view = UITableView(frame: .zero, style: .grouped)
-        view.backgroundColor = nil
-        view.dataSource = self
-        view.delegate = self
-        view.separatorStyle = .singleLine
-
-        view.register(UITableViewCell.self)
-        view.registerNib(InputCell.self)
```
designatednerd: ooh can we also get rid of `InputCell` and its `xib`?

marijnschilling: Noo, because we still use it in the ProfileCOntroller. It's actually a cell that allows the user to input the fields on the left. That's why it's great to be outa here, because here it was used as a static cell so no input at all. Now it's used how the creator meant it to be used 🙏

designatednerd: okely dokely

PR #866, opened 2018-05-01
IOS-627, part 2: Banners, Banners everywhere!

designatednerd:
This adds auto-hiding banners indicating what non-main network you're connected to for: 

- Everything in the wallet 
- Everything in the Dapps section
- The Web View
- Everything in the settings section except: 
    - The QR code controller that's going away
    - The balance controller that's going away
    - Controllers for storing your passphrase and your local currency since those preferences doesn't have anything to do with what network you're on.

Most of it looks like this when you're on a testnet: 
simulator screen shot - iphone se - 2018-05-01 at 15 07 01

~The only exception is the `WalletQRCodeViewController`, where it looks like this:~ [Removed]

I also converted a couple things from UITableViewControllers to UIViewControllers to facilitate this, and I got rid of the one storyboard we had in the process since it was more confusing to only have one rather than have some kind of mix.


petejkim:
Thanks! I think we can remove the banner for the barcode screen - the wallet address is going to be the same anyway.

Toshi/Controllers/Dapps/DappsListViewController.swift
```diff
@@ -31,6 +46,9 @@ final class DappsListViewController: UITableViewController {
 
     private var categoryId: Int?
 
+    lazy var activeNetworkView: ActiveNetworkView = defaultActiveNetworkView()
+    lazy var notificationObservers = [NSObjectProtocol]()
```
marijnschilling: Maybe this name should indicate that this is only the observers of the network switching? I can imagine it will get confusing if there's also other notifications in this VC.

Toshi/Controllers/Dapps/DappsListViewController.swift
```diff
@@ -70,29 +87,39 @@ final class DappsListViewController: UITableViewController {
...
     }
 
-    private func configureTableView() {
-        tableView.alwaysBounceVertical = true
-        tableView.showsVerticalScrollIndicator = true
-        tableView.estimatedRowHeight = 50
-        tableView.tableFooterView = UIView(frame: .zero)
-        tableView.separatorInset = UIEdgeInsets(top: 0, left: 100, bottom: 0, right: .defaultMargin)
+    deinit {
+        removeNotificationObservers()
```
marijnschilling: Maybe this name should indicate that this is only the observers of the network switching? I can imagine it will get confusing if there's also other notifications in this VC, you might think they are being removed by this function as well. (and yes the beginning of this comment I copied from the comment about the name for the array of those observers)

Toshi/Controllers/Wallet/CollectibleViewController.swift
```diff
@@ -13,6 +13,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import TinyConstraints
```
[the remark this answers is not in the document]
designatednerd: In theory, you're supposed to import a library into any file you use it in. I do get some occasional `Unknown Type Constrainable` warnings when it's not imported, probably because my computer is old and sad.

Toshi/Controllers/Wallet/CollectibleViewController.swift
```diff
@@ -58,7 +62,14 @@ final class CollectibleViewController: UIViewController {
 
     private func addSubviewsAndConstraints() {
         view.addSubview(tableView)
-        tableView.edges(to: view)
+
+        setupActiveNetworkView()
+
+        let guide = layoutGuide()
+        tableView.top(to: guide)
```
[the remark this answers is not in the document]
designatednerd: IIRC it doesn't work with layout guides but I'll give it another try.

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -118,21 +121,19 @@ class SettingsController: UIViewController {
...
+    lazy var activeNetworkView: ActiveNetworkView = defaultActiveNetworkView()
+    lazy var notificationObservers = [NSObjectProtocol]()
 
     override func viewDidLoad() {
         super.viewDidLoad()
 
         title = Localized.settings_navigation_title
 
-        view.addSubview(tableView)
-        tableView.topAnchor.constraint(equalTo: view.topAnchor).isActive = true
```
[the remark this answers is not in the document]
designatednerd: hey at least it was using anchors and `activate`

Toshi/Controllers/Dapps/DappsListViewController.swift
designatednerd: I left it general since this when it removes all the observers, it removes them without regard for what they are. I'd have to have it be a dictionary of `[Notification.Name: NSObjectProtocol]` in order to only remove specific observers. But I could also do that...

Toshi/Controllers/Dapps/DappsListViewController.swift
designatednerd: yeah true, that's a really good point.

Toshi/Controllers/DisappearingNavBarViewController.swift
```diff
@@ -28,11 +28,16 @@ import UIKit
 /// Note that all conformances are on the main class since methods in extensions cannot be overridden.
 class DisappearingNavBarViewController: UIViewController, DisappearingBackgroundNavBarDelegate, UIScrollViewDelegate, UIGestureRecognizerDelegate {
 
-    /// If disappearing is enabled at present.
+    /// If disappearing is enabled at present. Defaults to false.
```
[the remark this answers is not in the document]
designatednerd: Woof, good catch.

Toshi/Controllers/Dapps/DappsListViewController.swift
marijnschilling: Aha so it actually does waht it says, remove alllll the observers from the VC. 

I don't know how much hassle it is to create this in a dictionary, but it would be nicer to have this protocol only deal with the network notifications. But your call! ⚖️

Toshi/Controllers/Wallet/CollectibleViewController.swift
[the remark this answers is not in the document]
designatednerd: Nope, it does not. I'll file an enhancement ticket on TinyConstraints though, that seems like a reasonable thing to add.

Toshi/Controllers/Dapps/DappsListViewController.swift
designatednerd: I figured out a better way - will push shortly

designatednerd:
@marijnschilling Great suggestions, I think this makes the code a hell of a lot clearer.

designatednerd:
@petejkim removed, good point.

Toshi/Controllers/Dapps/DappsListViewController.swift
```diff
@@ -47,7 +47,7 @@ final class DappsListViewController: UIViewController {
     private var categoryId: Int?
 
     lazy var activeNetworkView: ActiveNetworkView = defaultActiveNetworkView()
-    lazy var notificationObservers = [NSObjectProtocol]()
+    var notificationObserver: NSObjectProtocol?
```
marijnschilling: Can this also be `activeNetworkObserver` 😬

designatednerd: ooh good point

PR #869, opened 2018-05-02
[IOS-649]: Enable network switching on all targets, show mainnet and ropsten for production

yuliaveres:


Toshi/Network/NetworkSwitcher.swift
```diff
@@ -106,7 +106,7 @@ final class NetworkSwitcher {
         #if DEBUG || TOSHIDEV
             return [.ropstenTestNetwork, .toshiTestNetwork]
         #else
-            return [.ropstenTestNetwork]
+            return [.mainNet, .ropstenTestNetwork]
```
designatednerd: a) do we also want to allow the toshi test network here? 
b) Is there anything that needs to change in terms of what's happening with the `TokenURLPaths` in addition to these changes to make this work?

yuliaveres: a) as you see - no
b) again - no until we have additional test networks, which will mean - just few new URLs

designatednerd: Right I saw that it's not there, I was wondering if that was an intentional omission

designatednerd:
I'm gonna merge this since Yulia's down in 🇦🇺 at the moment and done with her day.

yuliaveres:
Thanks @designatednerd

PR #872, opened 2018-05-03
[IOS-628] Wallet selection table view

marijnschilling:
- Setup table view and datasource for the wallet picker view
- Setup Cell for the wallet picker

- I used fake data for now, we need to implement the real wallet data later.

wallet selection


Toshi/Controllers/Settings/Wallet/WalletPickerDataSource.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        return cell
+    }
+}
+
+extension WalletPickerDataSource: UITableViewDelegate {
+
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+        indexPathForSelectedWallet = indexPath
```
yuliaveres: this also won't be needed as preferred wallet can be checked directly from user preferences storage when needed (when determining which index path should be selected), but i can also adjust it next

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -113,7 +113,7 @@ class SettingsController: UIViewController {
         return view
     }()
 
-    private var balance: NSDecimalNumber? {
+    private var selectedWallet: Wallet? {
```
yuliaveres: this won't be needed here as well right ?

Toshi/Controllers/Settings/Wallet/WalletPickerController.swift
```diff
@@ -0,0 +1,85 @@
...
+
+        guard isMovingFromParentViewController else { return }
+        guard let wallet = dataSource?.selectedWallet else { return }
+
+        delegate?.didSelectWallet(wallet)
+    }
+}
+
+extension WalletPickerController: NavBarColorChanging {
+    var navTintColor: UIColor? { return Theme.tintColor }
```
yuliaveres: why do you conform to `NavBarColorChanging ` here ? ..

Toshi/Controllers/Settings/SettingsNavigationController.swift
```diff
@@ -43,4 +43,51 @@ class SettingsNavigationController: UINavigationController {
             self.navigationBar.prefersLargeTitles = true
         }
     }
+
+    // MARK: - Nav Bar Color Handling
+
+    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
```
yuliaveres: i am missing something, but what is the purpose of all this "Nav Bar Color Handling" related code ? Is wallet list / picker going to be with green nav bar ?

yuliaveres: as currently it is plain, not sure if something is missing

marijnschilling: oh yeah, I thought it had to be green! But checked with robert and it doesn't (it was only in andorid design) Will remove it.

Toshi/Controllers/Settings/Wallet/WalletPickerDataSource.swift
marijnschilling: I leave it for now, cause it just shows how it would work. You can just replace it with the real way to check which wallet is selected.

yuliaveres: sure, the only questions I had are navigation bar style related , this things can (should, as I would base my next work on those) be merged as the are.

PR #873, opened 2018-05-03
Spring Cleaning: Update to new SwiftLint standards

designatednerd:
Updated Toshi to use B&B's official SwiftLint file, and fixed a ton of warnings that came up after the switch.

Did a bit of refactoring in places where it seemed like it could reduce either reliance on implicitly unwrapped optionals or methods with gigantic cyclomatic complexity. Tried to limit myself to only things that could be really easily extracted. 

Also found one file through warnings that's no longer used, and deleted it and its corresponding nib.


Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -371,141 +371,160 @@ extension SOFAWebController: WKScriptMessageHandler {
...
+            parameters[PaymentParameters.from] = from
+        }
+        if let to = transaction[PaymentParameters.to] {
+            parameters[PaymentParameters.to] = to
+        }
+        if let value = transaction[PaymentParameters.value] {
+            parameters[PaymentParameters.value] = value
+        } else {
+            parameters[PaymentParameters.value] = "0x0"
+        }
```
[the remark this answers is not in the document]
designatednerd: It'd only pass through whatever null values it receives, but I'm perfectly fine to leave it like this if you'd prefer.

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -362,6 +362,8 @@ extension SOFAWebController: WKScriptMessageHandler {
         }
     }
 
+    //swiftlint:disable cyclomatic_complexity
+    // TODO: Yeesh, make this less complicated.
```
marijnschilling: Woow! This function eh!

Toshi/Controllers/Wallet/SendToken/SendTokenViewController.swift
```diff
@@ -30,6 +30,7 @@ final class SendTokenViewController: UIViewController {
     var scrollViewBottomInset: CGFloat = 0.0
     var scrollView: UIScrollView { return configurator.scrollView }
 
+    // swiftlint:disable:next implicitly_unwrapped_optional - this is set up in viewDidLoad.
```
marijnschilling: We need to enable it again! or is this a special way of only disabling it for one line?

Toshi/SOFAWebview/SOFAWebController.swift
```diff
@@ -373,143 +371,155 @@ extension SOFAWebController: WKScriptMessageHandler {
             let payload = "{\\\"error\\\": null, \\\"result\\\": [\\\"" + Cereal.shared.paymentAddress + "\\\"]}"
             jsCallback(callbackId: callbackId, payload: payload)
         case .signPersonalMessage:
-            guard let messageBody = message.body as? [String: Any], let msgParams = messageBody["msgParams"] as? [String: Any], let messageEncodedString = msgParams["data"] as? String else {
-                jsCallback(callbackId: callbackId, payload: "{\\\"error\\\": \\\"Invalid Message Body\\\"}")
-                return
-            }
+            signPersonalMessage(from: message, callbackId: callbackId)
```
marijnschilling: This is a great improvement!

.swiftlint.yml
```diff
@@ -1,24 +1,39 @@
...
+  - redundant_nil_coalescing # nil coalescing operator is only evaluated if the lhs is nil, coalescing operator with nil as rhs is redundant
+  - switch_case_on_newline # Cases inside a switch should always be on a newline
+  - discouraged_object_literal # Prefer initializers over object literals (or use generated code)
+
 trailing_whitespace:
   ignores_empty_lines: true
+private_outlet:
+  allow_private_set: true
+cyclomatic_complexity:
+  ignores_case_statements: true
```
marijnschilling: Could you explain what this does exactly?

Toshi/Controllers/Wallet/SendToken/SendTokenViewController.swift
designatednerd: yeah `swiftlint:disable` disables until re-enabled. `swiftlint:disable:next`, `swiftlint:disable:this` and `swiftlint:disable:previous` only disable it for the specified line.

.swiftlint.yml
designatednerd: Yes! So when determining what the cyclomatic complexity of a method is (ie, how many possible paths are there for the method to take), this tells SwiftLint that it should ignore case statements when there's a `switch`, since the tradeoff in complexity is generally worth the safety it provides.

marijnschilling: aaah oke! Clear. but i was mostly aiming at you explaining it in the file itself with a comment 😬 (just as we do for the other rules for swiftlint)

designatednerd: Can do!

PR #874, opened 2018-05-03
[IOS-654] Settings structure

marijnschilling:
Made all the changes in one commit, I was in a flow! 🙈 

Points to #872 because it builds upon some changes made in that branch, i wanted to avoid conflictus mergus. Though it's maybe not the most logical way now. I suggest if you have time to review it now, I will just leave it open til #872 is merged and then point it to the base feature branch and merge it after. 

Took some refactoring in it too, cleaning up the cell for row.

img_387d2f49d809-1


Toshi/Controllers/Settings/SettingsNavigationController.swift
```diff
@@ -43,4 +43,51 @@ class SettingsNavigationController: UINavigationController {
             self.navigationBar.prefersLargeTitles = true
         }
     }
+
+    // MARK: - Nav Bar Color Handling
```
designatednerd: would it make sense to make an intermediate superclass that handles all this crap, then make that the superclass of Settings Nav controller and the other one that deals with changing colors? Might make it easier if we wind up adding it to even more shit in the future

Toshi/Controllers/Settings/Wallet/WalletPickerDataSource.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        self.tableView.dataSource = self
+        self.tableView.delegate = self
+
+        //TODO: implement getting the actual wallets of the user
+        self.wallets = [
+            Wallet(name: "Wallet1", address: "0xf1c76a75d8b3175fr8", imagePath: "https://bakkenbaeck.com/images/team/marijn.096ca0b8ab.jpg"),
+            Wallet(name: "Wallet2", address: "0xf3a65c12d8d3175fr8", imagePath: "https://bakkenbaeck.com/images/team/ellen.2454e06760.jpg"),
+            Wallet(name: "Wallet3", address: "0xf7b76a75d8b3175fr8", imagePath: "https://bakkenbaeck.com/images/team/yulia.dfc4e6cba7.jpg"),
+            Wallet(name: "Wallet4", address: "0xf353ca75d8b3175fr8", imagePath: "https://bakkenbaeck.com/images/team/mark.f73a421a3e.jpg")]
```
designatednerd: if you're gonna do these on separate lines, it usually helps to have the `]` on a separate line too so that if you need to add or delete something another, only the line that you've added shows up in the changes

Toshi/Resources/Base.lproj/Localizable.strings
```diff
@@ -95,19 +95,21 @@
 "transaction_error_message" = "Error completing transaction";
 
 /* Settings */
-"settings_navigation_title" = "Me";
+"settings_navigation_title" = "Settings";
```
designatednerd: 👍 thanks for finding this, it was making me nuts

designatednerd:
re the failure: You can delete any parts of `MyProfileRobot` that aren't currently used

PR #877, opened 2018-05-04
Remove redundant check if the login button is visible

marijnschilling:
before: 

screenshot 2018-05-04 12 05 34

after: 
screenshot 2018-05-04 12 02 06

I don't know why this test was failing though. But removing the check wether the login button view was visible made it succeed again. I think it's quite double to check wether a view is visible and then check wether that view is enabled. A view thats not visible can't be enabled. But please check my logic here :)


Tests_UI/Robots/SignInRobot.swift
```diff
@@ -142,7 +142,6 @@ extension SignInRobot {
 
     @discardableResult
     func validateSignInEnabled(file: StaticString = #file, line: UInt = #line) -> SignInRobot {
-        confirmViewVisibleWith(accessibilityLabel: SignInScreenButton.signIn.accessibilityLabel, file: file, line: line)
```
designatednerd: Hm...I wonder if it's failing "visibile" because it's behind the keyboard or scrolled offscreen or something. Enabled doesn't necessarily mean onscreen. but I think you're correct, this should only be testing enabled, not visible + enabled.

marijnschilling: yeah! I think so, it could be scrolled of screen. but in that case you just scroll it on screen and select it. But really happy I know now why it was failing. 🙏

PR #878, opened 2018-05-04
New settings

marijnschilling:
Made all the changes in one commit, I was in a flow! 🙈 

Took some refactoring in it too, cleaning up the cell for row.

img_387d2f49d809-1


Tests_UI/Robots/MyProfileRobot.swift
```diff
@@ -28,14 +28,13 @@ enum MyProfileCell {
...
     localCurrency,
-    qrCode,
     signOut,
     storePassphrase
     
     var accessibilityLabel: String {
         switch self {
         case .advanced:
-            return Localized.settings_cell_advanced
+            return Localized.settings_header_advanced
```
designatednerd: do we still need to be looking for `advanced` since it's no longer a cell?

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -202,59 +198,90 @@ class SettingsController: UIViewController {
...
+    private func cellForProfileSectionAt(indexPath: IndexPath) -> UITableViewCell {
+        guard let cell = tableView.dequeue(SettingsProfileCell.self, for: indexPath) as? SettingsProfileCell else { return UITableViewCell() }
+        guard let currentUserProfile = Profile.current else { return UITableViewCell()}
+
         cell.displayNameLabel.text = currentUserProfile.name
         cell.usernameLabel.text = currentUserProfile.displayUsername
-        
-        guard let avatarPath = currentUserProfile.avatar else { return }
+
+        guard let avatarPath = currentUserProfile.avatar else { return UITableViewCell() }
```
designatednerd: why return a fresh UITableviewCell here instead of just `cell`?

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -202,59 +198,90 @@ class SettingsController: UIViewController {
...
         switch item {
-        case .profile:
-            cell = tableView.dequeue(SettingsProfileCell.self, for: indexPath)
         case .wallet:
-           cell = setupWalletCell(for: indexPath)
+            //TODO: Implement real selected wallet cell name
+            walletCellConfigurator.configureCell(cell, withTitle: Localized.settings_cell_wallet, value: "Wallet 1")
+        case .network:
+            walletCellConfigurator.configureCell(cell, withTitle: Localized.settings_cell_network, value: NetworkSwitcher.shared.activeNetwork.label)
         default:
```
designatednerd: do we still need a default case here?

Tests_UI/Robots/MyProfileRobot.swift
designatednerd: oh derp this also can all be deleted since we're not really doing this testing yet

Toshi/Controllers/Settings/SettingsController.swift
marijnschilling: Copy paste mistake 😬

Toshi/Controllers/Settings/SettingsController.swift
marijnschilling: Well yeah because `SettingsItem` contains all items in all sections.

designatednerd: hmmmmm - maybe that points to breaking that up into enums per section instead?

Elland: If you're rewriting this tableview stuff, might as well go ahead and remove all this hard-coded stuff and use a backing model instead?

designatednerd: We should consider simplifying this but I don't know that putting it in a model is going to make it all that much simpler than just using exhaustive enums per section. 

Either way: This can wait until another PR, the rest of this looks good so let's get it merged.

marijnschilling: Do you have an example of that i think it'd be a good improvement.

designatednerd: Yah i'll throw a playground into the slack channel once I have this stupid nonsense with push figured out

marijnschilling: Next time 😅

PR #881, opened 2018-05-04
Fix push registration

designatednerd:
So part of this was a dumb move on my end: I was using the `cereal.address` instead of the `cereal.paymentAddress`. 

The other part is that order we're sending parameters to the server with apparently matters. I had to update the constructor and order of everything for `APIPushInfo` to match the order previously, because even when the address and registration ID were identical, if they weren't sent in the proper order, the signature verification would fail. 

So yeah, fun one to look out for.


tristan:
@designatednerd 
> order we're sending parameters to the server with apparently matters

The order you send to the server doesn't matter, what matters is that the data you use to generate the `hashedPayload` must be exactly the same as what is sent in the body of the request.

the real issue here is that your new serialiser is not serialising the json in the same way as the teapot serialiser. So your fix here is only really fixing the symptoms for this one case, not the underlying issue, which could cause problems system-wide.

designatednerd:
@tristan Well, this at least explains why this was the only endpoint affected - this is the only signed endpoint that uses `Codable` and its companion `JSONEncoder` to create the data, and then directly used the data to create the signature. Everything else that's signed is just passing a dictionary directly. 

Damned weird though - I was handing the encoded JSON data directly to Teapot, so I'll need to dig into what the fuck Teapot is doing with the data that caused the params to get reversed.

Elland:
De/re-serialising the data under the hood should not matter, and it certainly did not matter before. I'm wondering now why you're encoding the data manually with `JSONEncoder` instead of passing it as an array or dictionary directly. 🤔

Toshi/Models/Users/APIPushInfo.swift
```diff
@@ -17,13 +17,14 @@ import Foundation
 
 struct APIPushInfo: Codable {
 
-    let address: String
+    // NOTE: Order matters on these - if the JSON is not sent in the registration_id -> address order, the signature verification fails.
```
Elland: Mind updating this comment? 👍

designatednerd: Yeah I'm going to update all of this when the teapot stuff gets merged

PR #883, opened 2018-05-07
[IOS-652, IOS-639, IOS-640]: Multiple wallets generation, preferred wallet storage

yuliaveres:
img_0596


Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -439,7 +440,8 @@ final class EthereumAPIClient {
         let address = cereal.address
         let path = "/v1/apn/deregister"
 
-        let params = ["registration_id": appDelegate.token, "address": cereal.paymentAddress]
+        // No need to specify addressed here, as we want to deregister all of them
+        let params = ["registration_id": appDelegate.token]
```
designatednerd: Might be worth pulling in changes from master here - you'd need to make `address` an optional on `APIPushInfo`, but it'll save some merge headaches in the long run.

Toshi/Utilities/UserDefaultsWrapper.swift
```diff
@@ -190,6 +191,15 @@ class UserDefaultsWrapper: NSObject {
             setValue(newValue, for: .activeNetwork)
         }
     }
+
+    static var activeWalletPath: Int? {
+        get {
+            return int(for: .activeWalletPath)
+        }
+        set {
+            setValue(newValue, for: .activeWalletPath)
```
designatednerd: in theory, there shouldn't be anything an attacker can do with just the wallet path since the actual passphrase is stored in the keychain, correct?

designatednerd:
Looking solid! just a couple questions

Toshi/APIClients/EthereumAPIClient.swift
yuliaveres: yes, that one is merged now! will do

Toshi/Utilities/UserDefaultsWrapper.swift
yuliaveres: correct..

Toshi/Models/Wallet/Wallet.swift
```diff
@@ -0,0 +1,128 @@
...
+        let walletKeychain: BTCKeychain = Wallet.walletKeychain(from: mnemonic, lastPath: path)
+        let walletPrivateKey = walletKeychain.key.privateKey.hexadecimalString()
+        self.cereal = EtherealCereal(privateKey: walletPrivateKey)
+    }
+
+    func activate() {
+        UserDefaultsWrapper.activeWalletPath = Int(path)
+    }
+
+    static func generate(mnemonic: BTCMnemonic) {
```
marijnschilling: I'm wondering is it possible to test this function? The wallets stay the same every time right?

yuliaveres: they are same for the passphrase, yes

PR #885, opened 2018-05-07
[CORE-131]: Network Switching

designatednerd:
This is a PR of the feature branch we've been using for Network Switching to `master`. Should have an extra set of eyes before the merge.


marijnschilling:
img_bcce8420de46-1
What i understood in the meeting yesterday is that we shouldn't show the network bar when we're on the default network of the current target. But here i'm on debug and on Toshi Test Network i do get te banner. Is this a bug?

designatednerd:
@marijnschilling Yep this is what #888 fixes

designatednerd:
Moved to a separate PR.

PR #889, opened 2018-05-08
WIP: Investigate possible frequency of `Cereal` not being able to load from yap

designatednerd:


designatednerd:
Fixed via #891

PR #890, opened 2018-05-08
[CORE-130]: Feature hd wallet

marijnschilling:
This is a PR of the feature branch we've been using for HD-wallet to master. Should have an extra set of eyes before the merge.


marijnschilling:
Confirmed that this bug is fixed after the master was merged: https://toshiapp.atlassian.net/browse/IOS-656

Toshi/Models/Wallet/Wallet.swift
```diff
@@ -0,0 +1,128 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import Foundation
+import EtherealCereal
+import HDWallet
+import Haneke
+import BlockiesSwift
+
+struct Wallet {
```
designatednerd: I'm concerned that there aren't any tests on this. Particularly given that this deals with users' money, I don't believe we should merge this without at least some tests validating expected and/or any known error states.

Toshi/Models/Wallet/Wallet.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    static var walletsAddresses: [String] {
+        return items.map { $0.address }
+    }
+
+    static var activeWallet: Wallet {
+        return items.first(where: { $0.path == activeWalletPath }) ?? Wallet.items.first!
+    }
+
+    static private var identiconGenerationQueue: OperationQueue = {
```
designatednerd: I think it would be helpful to pull this identicon stuff out into its own class or struct. Both from a testing perspective (reducing potential stuff we need to test on the wallet, making it easier to specifically test identicon generation if we decide we should), and a reusability perspective (what if we want to generate identicons someplace else?), it seems like a reasonable place to separate things out. 

Doesn't have to be before merge, but if it helps in getting things tested on the wallet, great.

Toshi.xcodeproj/project.pbxproj
```diff
@@ -1016,6 +1017,7 @@
 		C1128E2CDD482DB9BE1BF5A3 /* libPods-CocoaPods-Distribution.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 7E67E3F2BFB7B0171BF61B30 /* libPods-CocoaPods-Distribution.a */; };
 		C904ABD5324415F45702AB48 /* libPods-CocoaPods-Distribution.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 2446336EA68730ACD1CE100D /* libPods-CocoaPods-Distribution.a */; };
 		D197B006BD046DC2E6B46351 /* String+nsRange.swift in Sources */ = {isa = PBXBuildFile; fileRef = D197B695935159A20363BBF9 /* String+nsRange.swift */; };
+		D197B02250886AD34FFCD093 /* SecuriryCellConfigurator.swift in Sources */ = {isa = PBXBuildFile; fileRef = D197BDE56C0F64F372F9B728 /* SecuriryCellConfigurator.swift */; };
```
designatednerd: it appears that while the class here is named correctly, the file did not get renamed when a spelling error was discovered.

Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -346,25 +318,19 @@ extension SettingsController: UITableViewDelegate {
...
+            self.navigationController?.pushViewController(WalletPickerController(), animated: true)
+        case .network:
+            navigationController?.pushViewController(NetworkSettingsController(), animated: true)
         case .localCurrency:
             self.navigationController?.pushViewController(CurrencyPicker(), animated: true)
-        case .advanced:
-            self.pushViewController("AdvancedSettings")
+        case .about:
+            //TODO: implement about ViewController when design is ready
+            DLog("push about ViewController")
```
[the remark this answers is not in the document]
marijnschilling: I'll check! Otherwise we'll leave out the cell. I'll pick this up!

Toshi/Models/Wallet/Wallet.swift
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            identiconGenerationQueue.addOperation(operation)
+        }
+
+        items = generatedWalletItems
+    }
+
+    private static func walletKeychain(from mnemonic: BTCMnemonic, lastPath: UInt32) -> BTCKeychain {
+        // wallet path: 44H/60H/0H/0 and then 0 again. Metamask root path, first key.
```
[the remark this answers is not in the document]
marijnschilling: @yuliaveres Could you pick this up?

Toshi/Models/Wallet/Wallet.swift
marijnschilling: I'll pick up writing tests for this. Def's shouldn't merge this without tests! 🙏

Toshi/Models/Wallet/Wallet.swift
marijnschilling: I'll pick this up when writing tests. I'll let @yuliaveres know at the end of the day if i need any help with this :)

Toshi/Models/Wallet/Wallet.swift
marijnschilling: Never mind, @yuliaveres is already working on this

yuliaveres:
Moving to a different PR.

PR #891, opened 2018-05-08
Hotfix for Cereal issues

designatednerd:
This version: 

- Disables account creation if you're not connected to the internet or if the creation process otherwise fails, preventing you from winding up in a state where you could have an account with no user. 
- ADDED: Refactors the `Cereal` class so that the empty initializer is no longer available. The shared instance is initialized under the hood only with a retrieved instance - it must otherwise be explicitly initialized by registering or logging in. If you attempt to access the shared instance without checking that it exists, the app will crash. 
- Will log and crash if for some reason the user has created and stored a passphrase in Yap, but we are unable to access it. This should prevent a user from getting into a weird state where a failure by Yap causes a new passphrase to be generated, causing them to potentially lose access to their account if they haven't secured their passphrase.
- Updates Teapot so we actually know when requests fail due to network failures.
- Adds the skeleton of some alert convenience updates, but only implements it in the places we're showing alerts related to this problem. 

Version 1.3.6 should be cut, tagged, and released from this branch, then we should update based on `master` so these fixes can be merged back in to the continuing code.


Toshi/Models/Cereal.swift
```diff
@@ -216,5 +222,6 @@ class Cereal: NSObject {
 
     @objc private func userCreated(_ notification: Notification) {
         Yap.sharedInstance.insert(object: mnemonic.words.joined(separator: " "), for: Cereal.privateKeyStorageKey)
```
marijnschilling: Use is also created on line 71 in Cereal. We also need to set this value to true there.

designatednerd: Yes! The refactoring I just did centralized shoving this in the database to this method - adding it in the other initializer wasn't working anyway, since the database hadn't been initialized. It was only once this method was hit that it was actually persisted 😛

Toshi/Models/Cereal.swift
```diff
@@ -84,40 +145,24 @@ class Cereal: NSObject {
...
-            guard result == 0 else {
-                CrashlyticsLogger.log("Failed to generate random entropy data")
-                fatalError("Failed to randomly generate and copy bytes for entropy generation. SecRandomCopyBytes error code: (\(result)).")
-            }
+    private override init() {
+        fatalError("Don't use the empty initializer!")
+    }
 
-            mnemonic = BTCMnemonic(entropy: entropy, password: nil, wordListType: BTCMnemonicWordListType.english)!
+    static func generateEntropy() -> Data {
```
marijnschilling: Can this be private as well? Or is it public for testing purposes?

Toshi/Models/Cereal.swift
```diff
@@ -20,9 +20,39 @@ import HDWallet
 /// An EtherealCereal wrapper. Generates the address and public key for a given private key. Signs messages.
 class Cereal: NSObject {
 
-    @objc static var shared: Cereal = Cereal()
+    /// The underlying shared Cereal. Will be nil at first if there is no stored cereal,
+    /// at which point it should be set once the user is created.
+    private static var _shared: Cereal? = {
+        guard let passphrase = Cereal.storedPassphrase() else {
+            if UserDefaultsWrapper.hasStoredPassphraseInYap {
```
marijnschilling: This is a great check, so if we couldn't found the stored passphrase but the password has been saved it will throw an fatal error. So this will pop up when we're dealing with race conditions from the Yap database not having stored the password yet when we try to receive it.

Tests/CerealTests.swift
```diff
@@ -91,4 +92,40 @@ class CerealSignEthereumTransactionWithWalletTests: XCTestCase {
...
+        XCTAssertNotNil(Cereal(entropy: Cereal.generateEntropy()), "Could not generate cereal from entropy")
+    }
+
+    func testGeneratingFromEmptyDataFails() {
+        XCTAssertNil(Cereal(entropy: Data()), "Should not be able to generate cereal from empty data")
+    }
+
+    func testGeneratingFromValidWordList() {
+        guard let fromWordList = Cereal(words: words) else {
+            XCTFail("Could not generate cereal from vali word list")
```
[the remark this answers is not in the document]
designatednerd: 🤦‍♀️

Toshi/Models/Cereal.swift
designatednerd: Public for testing purposes

Toshi/Controllers/Sign In/SignInViewController.swift
```diff
@@ -105,12 +105,14 @@ final class SignInViewController: UIViewController {
...
             switch result {
+            case .notConnected:
+                strongSelf.showErrorOKAlert(message: Localized.alert_no_internet_message)
             case .signUpWithPassphrase:
                 let alertController = UIAlertController(title: Localized.sign_up_with_passphrase_alert_title, message: Localized.sign_up_with_passphrase_alert_message, preferredStyle: .alert)
                 alertController.addAction(UIAlertAction(title: Localized.cancel_action_title, style: .cancel, handler: nil))
                 alertController.addAction(UIAlertAction(title: Localized.sign_up_with_passphrase_accept_action_title, style: .default, handler: { _ in
                     guard let validCereal = Cereal(words: passphrase) else { return }
-                    Cereal.shared = validCereal
+                    Cereal.setSharedCereal(validCereal)
```
yuliaveres: this is probably also too early to set `shared`, as user might not accept terms / or change his mind and create a new account with different seed, while "new" will be created with this one.

designatednerd: This wouldn't set the cereal unless they actually performed the action being added though, which is accepting the terms

Toshi/Controllers/Sign In/SplashViewController.swift
```diff
@@ -224,6 +238,12 @@ final class SplashViewController: UIViewController {
     }
 }
 
+extension SplashViewController: ActivityIndicating {
+    var activityIndicator: UIActivityIndicatorView {
+        return activityView
+    }
+}
```
designatednerd: you actually can just make the `lazy var` be named `activityIndicator`- then you don't have to add more than the mix-in declaration of conformance to `ActivityIndicating`.

Toshi/Supporting Files/SessionManager.swift
```diff
@@ -92,6 +93,16 @@ final class SessionManager {
...
 
+            guard let status = Navigator.tabbarController?.reachabilityManager.currentReachabilityStatus else {
+                // Can't check status but just to be safe:
+                return
+            }
+
+            if status == .notReachable {
+                completion(.notConnected)
+                return
+            }
```
designatednerd: would it be worth checking that there actually was an error (or at least no profile) before performing these checks?

Toshi/Controllers/Sign In/SignInViewController.swift
yuliaveres: @designatednerd i guess my explanation is bad. This alert is about making sure the use wants to create a new Toshi account with the given seed s/he inputted on Sign in View. 
"Create new account action"
img_0598
cereal was set here, but after user clicks this action, s/he is thrown back to splash when "Accept terms" alert is shown. And there s/he should accept terms and only after user is created, cereal should be set.

I have tested this flow, and Agreed to crate a new toshi user with the seed I provided on Sign in flow. Then I was thrown to Splash, but I have not accepted terms, or just decided to create a new Toshi account with different seed, so I tapped "Create new account" on Splash (which should generate random 12 words for me). But since cereal was set before, it was used afterwards and new account did not have new 12 words generated, but same which I have entered in Sign in view.

designatednerd: Ahaaaaaaaaa, that makes more sense. Thank you very much for the explanation!

marijnschilling:
Doing Q&A with list of tests below. Let me know if there's other things i should test. I just tested on my device right now.
- Sign in ✅ 
- Sign in while offline ✅
- Sign in with wrong passphrase ❌ (see below)
- Sign in with unregistered passphrase ✅
- Sign in with unregistered passphrase - cancel on accept terms - create new account ✅

- Create new account ✅
- Create new account while offline ✅

Everything works except signing in with a wrong passphrase (i.e: valid words just not a valid combination) Doing that dismisses the sign-up view and brings me to this view:
screenshot 2018-05-09 10 26 07
But i don't know if that was introduced by this fix, it isn't blocking this hotfix.

designatednerd:
woot, tests passing now! Gonna merge this into master.

PR #892, opened 2018-05-09
[IOS-651]: Use additional test networks

yuliaveres:
Now we do not allow network switching on Dev or Debug, just on Production, due to PN registration certificates related reason. New networks are added here - Kovan and Rinkeby; Ropsten network path being adjusted according to the build.


Toshi/Controllers/Settings/SettingsController.swift
```diff
@@ -36,7 +36,11 @@ class SettingsController: UIViewController {
             case .security:
                 return [.security]
             case .settings:
+                #if DEBUG || TOSHIDEV
+                return [.localCurrency, .signOut]
+                #else
                 return [.localCurrency, .advanced, .signOut]
+                #endif
```
designatednerd: a) Can you indent the `return` statements here since they're within an if/else block please? 
b) can you clarify a bit why push notifications aren't working for other networks if we're on debug, which seems like why we're disabling access to the advanced section here? 
c) How will this be affected in the future when we add the wallet stuff to the advanced section?

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -26,43 +26,61 @@ enum NetworkInfo {
...
         static let MainNet = "https://ethereum.service.toshi.org"
-        static let RopstenTestNetwork = "https://ethereum.development.service.toshi.org"
         static let ToshiTestNetwork = "https://ethereum.internal.service.toshi.org"
+        static let Kovan = "https://toshi-eth-service-kovan.herokuapp.com"
+        static let Rinkeby = "https://toshi-eth-service-rinkeby.herokuapp.com"
+        #if TOSHIDEV
+        static let Ropsten = "https://ethereum.development.service.toshi.org"
+        #else
+        static let Ropsten = "https://toshi-eth-service-ropsten.herokuapp.com"
+        #endif
```
designatednerd: a) Same request on indentation
b) Should this also check for `DEBUG`, or does the `TOSHIDEV` flag also cover the debug build?
c) Why does this change Ropsten instead of the toshi test network? It seems like it's a different address for that rather than for the Ropsten network, but maybe I'm missing something happening under the hood
d) This could also just be moved into the `path` property of Network instead of being a separate struct if you'd like

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -26,43 +26,61 @@ enum NetworkInfo {
     struct Label {
 
         static let MainNet = Localized.mainnet_title
-        static let RopstenTestNetwork = Localized.ropsten_test_network_title
         static let ToshiTestNetwork = Localized.toshi_test_network_title
+        static let Kovan = Localized.kovan_network_title
+        static let Rinkeby = Localized.rinkeby_network_title
+        static let Ropsten = Localized.ropsten_network_title
```
designatednerd: you could remove this struct and just return all these localized strings from the `label` property on the `Network` enum if you wanted to simplify it.

Toshi/Network/NetworkSwitcher.swift
```diff
@@ -103,10 +121,12 @@ final class NetworkSwitcher {
...
     var availableNetworks: [Network] {
-        #if DEBUG || TOSHIDEV
-            return [.ropstenTestNetwork, .toshiTestNetwork]
+        #if DEBUG
+        return [.toshiTestNetwork]
+        #elseif TOSHIDEV
+        return [.ropsten]
         #else
-            return [.mainNet, .ropstenTestNetwork]
+        return [.mainNet, .ropsten, .rinkeby, .kovan]
```
designatednerd: same request on indentation

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -399,7 +399,7 @@ final class EthereumAPIClient {
                                                     completion: ((_ success: Bool, _ message: String?) -> Void)? = nil) {
         var path = "/v1/apn/register"
         if !register {
-            var path = "/v1/apn/deregister"
+            path = "/v1/apn/deregister"
```
designatednerd: ...how did this ever compile....

Toshi/Controllers/Settings/SettingsController.swift
designatednerd: [@tristan clarified b) for me but other questions still apply]

yuliaveres: We will still be able to have whatever in advanced section, but move this limitation to cell level (network cell)...

Toshi/Network/NetworkSwitcher.swift
yuliaveres: b) - Ropsten is not used / available on Debug
c) - so the only network path that varies depending on Build is Ropsten, On dev builds - it is old one, on Production - it is new one `https://toshi-eth-service-ropsten.herokuapp.com`

PR #893, opened 2018-05-09
Hotterfix

designatednerd:
This PR is branched off of the submitted build. 

This is what was causing our tests to fail - hitting stuff in the profile before we had a shared cereal. I *think* we might be safe as if installing a brand new build and then signing in or creating a user works on the submitted build. It might just be a testing issue.


designatednerd:
Closing this as unnecessary - only issue was with tests, nothing else hits the profile without having gotten to the appropriate point. Will go with fixes on the primary `hotfix` branch.

PR #897, opened 2018-05-10
[IOS-659] Implement add token service methods

marijnschilling:
Added the two new API Calls and added tests for them

ADD CUSTOM ERC20 TOKEN:

> POST `/v1/token`
> Adds a custom erc20 token to the list of tokens, and/or forces a token to always show in the token balance list even if it has a 0 balance. Uses Toshi-ID authentication headers to proove ownership of an ethereum address.
> 
> name, symbol and decimals are optional, and the service will attempt to get the value from the contract metadata if they aren’t given.

SPECIFIC ERC20 TOKEN:

> GET /v1/token/{token_address}
> 
> Returns name, symbol and decimals for the given erc20 token

- I will add tests for the expected headers as well as soon as [bakkenbaeck/Teapot#33](url) is merged.
- I'm wondering if making the address optional in `CustomToken` struct is the good way to go. When adding a token this field is _not_ optional. But when getting specific token we use the address to return the three other fields. So to create the same `CustomToken` object from those results I had to make the address optional. I'm wondering if I should just add the address where the request was made with to the CustomToken tehre to make it non-optional.

### DIFF
--- a/Tests/Mocks/AddToken/0x4d8fc1453a0f359e99c9675954e656d80d996fbf.json
+++ b/Tests/Mocks/AddToken/0x4d8fc1453a0f359e99c9675954e656d80d996fbf.json
@@ -1,0 +1,5 @@
+{
+  "symbol": "BEE",
+  "name": "Bee",
+  "decimals": 18
+}
--- a/Toshi/APIClients/EthereumAPIClient.swift
+++ b/Toshi/APIClients/EthereumAPIClient.swift
@@ -354,12 +354,16 @@ final class EthereumAPIClient {
 
             let customTokenData = CustomToken(contractAddress: address, name: name, symbol: symbol, decimals: decimals)
             guard let jsonData = customTokenData.toOptionalJSONData() else {
-                completion(false, nil)
+                DispatchQueue.main.async {
+                    completion(false, nil)
+                }
                 return
             }
 
             guard let headers = try? HeaderGenerator.createHeaders(timestamp: timestamp, path: path, payloadData: jsonData) else {
-                completion(false, nil)
+                DispatchQueue.main.async {
+                    completion(false, nil)
+                }
                 return
             }
 
--- a/Toshi/APIClients/EthereumAPIClient.swift
+++ b/Toshi/APIClients/EthereumAPIClient.swift
@@ -347,13 +347,13 @@ final class EthereumAPIClient {
         }
     }
 
-    func addToken(with address: String, completion: @escaping ((_ success: Bool, _ error: ToshiError?) -> Void)) {
+    func addToken(with address: String, name: String? = nil, symbol: String? = nil, decimals: Int? = nil, completion: @escaping ((_ success: Bool, _ error: ToshiError?) -> Void)) {
         timestamp(mainTeapot) { timestamp, _ in
             guard let timestamp = timestamp else { return }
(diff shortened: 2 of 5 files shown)


designatednerd:
Ooh I bet the time stamp is looking for the headers even though it’s an overridden endpoint - I will try to fix this tomorrow

Tests/Mocks/AddToken/0x4d8fc1453a0f359e99c9675954e656d80d996fbf.json
```diff
@@ -0,0 +1,5 @@
+{
+  "symbol": "BEE",
```
designatednerd: 🐝

Toshi/APIClients/EthereumAPIClient.swift
```diff
@@ -347,6 +347,85 @@ final class EthereumAPIClient {
...
     }
 
+    func addToken(with address: String, name: String? = nil, symbol: String? = nil, decimals: Int? = nil, completion: @escaping ((_ success: Bool, _ error: ToshiError?) -> Void)) {
+        timestamp(mainTeapot) { timestamp, _ in
+            guard let timestamp = timestamp else { return }
+            let path = "/v1/token"
+
+            let customTokenData = CustomToken(contractAddress: address, name: name, symbol: symbol, decimals: decimals)
+            guard let jsonData = customTokenData.toOptionalJSONData() else {
+                completion(false, nil)
```
designatednerd: you probably need to fire these completion blocks off on the main thread

yuliaveres:
@marijnschilling could you please make sure this works ? 
https://bakkenbaeck.slack.com/archives/C0PHFR45V/p1526287942000254

yuliaveres:
especially, `POST` test token with its address `0x4306b90c513d133d66f6925a4fd3ee675ce73c67` (it is shown here as well: https://token-eth-faucet-development.herokuapp.com) to check if our implementation does what is needed.

Toshi/Controllers/Wallet/Models/Token.swift
```diff
@@ -139,3 +139,20 @@ extension Token: WalletItem {
         return symbol
     }
 }
+
+// MARK: - Custom Token
+
+struct CustomToken: Codable {
```
yuliaveres: I would not create a third way to represent a token, we already have `Token`, `EtherToken` and here `CustomToken` which are all of same purpose and fields (some might be optional). 
While implementing wallet items / tokens persistence, I tackled an issue with EtherToken, after being fetched from cache, being represented as Token; thus a lot of things were broken due to differences in their implementation. I got rid of EtherToken and moved logic relating to it to Token (https://github.com/toshiapp/toshi-ios-client/pull/901). Please, could you adjust and use Token as well ?

yuliaveres: (discussed)

marijnschilling: I don't think i can do the same for CustomToken, i would have to make all Token related fields optional. Which would mean that if the type is not handled well things will be very much broken as well. So it won't solve the problem. I'm going to discuss a better solution, but for now leave customToken as a type.

marijnschilling:
Moved over to the private rep.

PR #898, opened 2018-05-11
Separate Carthage Updates from feature branch

designatednerd:
Trying to make #890 a little less gigantic


designatednerd:
Since this only deals with dependencies that are definitely getting added in another PR and it passes all tests, and the other Toshi iOS folks are off today, I'm gonna go ahead and merge this.

PR #899, opened 2018-05-11
Image Picking Centralization + Other Alerting Improvements

designatednerd:
In this PR: 

- Image picking has been (mostly) centralized to a protocol. This also allowed me to centralize not showing the option for a camera if a device doesn't have one (ex, is a simulator or an old-ass iPod Touch)
- Extension to show alerts got a bunch of love: 
    - Convenience extension method to show an action sheet
    - Test helpers extension got collapsed into the alerting extension
    - Actual alert construction was pulled into an extension on `UIAlertController`
- A convenience extension method has been added on `UIAlertAction` to quickly create a cancel action (in addition to the existing `.okAction()`), a default-style or a destructive-style action. Why `.default` isn't actually provided as the default value for `style` is beyond me.


designatednerd:
OK this is now fully ready for review. Cleaned up a couple things and added better handling of weakself, but otherwise not too bad.

designatednerd:
Moved to a different PR.

PR #900, opened 2018-05-14
[IOS-661]: Wallet tests and identicon logic extraction

yuliaveres:


Toshi/Models/Wallet/Wallet.swift
```diff
@@ -102,9 +101,7 @@ struct Wallet {
 
             let operation = BlockOperation()
             operation.addExecutionBlock {
-
-                let blockies = Blockies(seed: wallet.address)
-                if let identicon = blockies.createImage() {
+                if let identicon = Identicon.generate(for: wallet.address) {
```
marijnschilling: Do we need to catch an error when the identicon generation fails?

Toshi/Avatars/Identicon.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// Generates identicon for given address.
+    ///
+    /// - Parameters:
+    ///   - seed: The seed to be used.
+    ///   - size: The number of blocks per side for this image. Defaults to 8.
+    ///   - scale: The number of pixels per block. Defaults to 6.
+    ///   - color: The foreground color. Defaults to nil, then random is used indside Blockies.
+    ///   - bgColor: The background color. Defaults to nil, then random is used indside Blockies.
+    ///   - spotColor: A color which forms mouths and eyes. Defaults to nil, then random is used indside Blockies.
+    static func generate(for address: String, size: Int = defaultBlocksSize, scale: Int = defaultScale, color: UIColor? = nil, bgColor: UIColor? = nil, spotColor: UIColor? = nil) -> UIImage? {
```
marijnschilling: I think a basic tests for when we put in a valid address it should return a valid result could help us detect regression on this method. 

We could even consider testing this method by comparing the data of the result, but not sure if that would be valuable. You can decide!

Tests/WalletTests.swift
```diff
@@ -0,0 +1,54 @@
...
+                "0xa40cfbfc8534ffc84e20a7d8bbc3729b26a35f6f",
+                "0xb191a13bfe648b61002f2e2135867015b71816a6",
+                "0x593814d3309e2df31d112824f0bb5aa7cb0d7d47",
+                "0xb14c391e2bf19e5a26941617ab546fa620a4f163",
+                "0x4c1c56443abfe6dd33de31daaf0a6e929dbc4971"]
+    }()
+
+    private lazy var testCereal: Cereal = {
+        guard let cereal = Cereal(words: ["abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "about"]) else {
+            fatalError("failed to create cereal")
```
marijnschilling: I think for tests it's clearer to throw a XCTAssertFalse("failed to create cereal") when this happens, this way we get a clear message when and why the test failed instead of breaking the tests that are being run.

Not sure if this means you need to move this to a local var in the testMethod or if It might also work when you do the assertion in this lazy var though, but maybe test if that works!

Tests/WalletTests.swift
```diff
@@ -0,0 +1,54 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+@testable import Toshi
+import XCTest
+import BlockiesSwift
```
marijnschilling: Why is this import necessary here?

Tests/WalletTests.swift
marijnschilling: Ha the idea is that we would circumvent the fatalError this way, since then the tests will crash and fail instead of just fail with an error. 

ut i saw this being done in other placed of the app as well, so it's really not vital.

Tests/WalletTests.swift
```diff
@@ -0,0 +1,54 @@
...
+    }()
+
+    private lazy var testCereal: Cereal = {
+        guard let cereal = Cereal(words: ["abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "abandon", "about"]) else {
+            fatalError("failed to create cereal")
+        }
+        return cereal
+    }()
+
+    override func setUp() {
```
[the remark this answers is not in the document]
designatednerd: [addressed]

Toshi/Models/Wallet/Wallet.swift
designatednerd: [addressed]

Toshi/Avatars/Identicon.swift
designatednerd: [addressed]

PR #901, opened 2018-05-15
[IOS-643]: Persist tokens on disk

yuliaveres:
Move Ether token into Token due to caching specifics. Add new `balance` field to Token, now `value` is considered deprecated and will be removed soon.


Toshi/Controllers/Wallet/Models/Token.swift
```diff
@@ -18,90 +18,85 @@ import Foundation
...
 
     lazy var displayValueString: String = {
-        return self.value.toDisplayValue(with: self.decimals)
+        guard self.isEtherToken, let valueInWei = wei else { return self.balance.toDisplayValue(with: self.decimals) }
+        return EthereumConverter.ethereumValueString(forWei: valueInWei, withSymbol: false, fractionDigits: 6)
     }()
 
     var isEtherToken: Bool {
-        return symbol == "ETH"
+        return symbol == Token.EthSymbol
```
marijnschilling: Isn's it also exclusively an etherToken when the wei value in non nil? And they other way around? When the wei value is nil, it's not an ether token?

marijnschilling:
I made a feature branch `feature/custom-token` could you set that as base branch?

Toshi/Controllers/Wallet/Models/Token.swift
yuliaveres: kind of. but this is better way to distinguish, by unique symbol

PR #904, opened 2018-05-18
Amplitude for event tracking

johnnycoinbase:
### *What changed? Why?*
Amplitude integration for event tracking. An `Analytics` class has been added to abstract away the implementation details and to allow for easier swapping of event tracking platforms if needed.
Notice that there are production and development keys to ensure that we don't pollute production data during CI/test/development.

### *How has it been tested?*
Viewed the events being tracked on the Amplitude dashboard.


Toshi/APIClients/APIKeys/APIKeys.swift
```diff
@@ -35,11 +36,19 @@ struct APIKeys {
...
-        }
-
-        CrashlyticsClient.start(with: fabricKey)
+        if let fabricKey = APIKeys.key(named: APIKeys.Fabric) {
+			CrashlyticsClient.start(with: fabricKey)
+		} else {
+			DLog("Can't load Fabric API Key")
+		}
+
+		let amplitudeKey = Bundle.main.bundleIdentifier == "org.toshi.distribution" ?
```
petejkim: instead of hardcoding both dev and prod keys in the code and using bundleIdentifier select a key at runtime, perhaps it's better to put `$(AMPLITUDE_API_KEY)` in Info.plist and have the var defined for each build configuration?

petejkim:
Is it not necessary to `identify` user? by some unique identifier? e.g. primary wallet address?

Toshi/APIClients/APIKeys/APIKeys.swift
johnnycoinbase: The keys live in a folder that is git-ignored so as to not expose them publicly so it can't be put in the project's info.plist. I can put a comment here to explain that.

petejkim: I see, I thought `AmplitudeProductionAPIKey` and `AmplitudeDevelopmentAPIKey` were keys themselves and the values were just placeholders, but those are _keys_ as in keyval keys for `APIKeys`.

Issue #852, opened 2018-04-25
Black sceen when opening app


Idealist42:
When I open the app I get a black screen and nothing loads.

If I force quit the app and re-open it, it launches.

heumn:
Does this happen the first time you opened the app? Every time? Can you reproduce it in some ways?

Idealist42:
It was reproducible on my iPhone 6.  

It would happen every time I opened the app, except for the first time opening after force quitting.

I just got a new iPhone 8+ today and have only had the issue once since setting up the new phone.  I've been trying to reproduce it on the new phone, but it stopped happening as soon as I opened this bug report.

designatednerd:
@Idealist42 Can you confirm what version of the app this was happening on with your 6?

Idealist42:
I was using the newest version.  There were no updates available.

Idealist42:
Follow up on this issue...  I didn't open the app for a few days on my new phone, and I just tried opening it.  Getting the black screen again now on the new device.

designatednerd:
Thank you for the heads up - we'll try to get a look at it.

PR #905, opened 2018-05-21
Fixes for IOS-633: Coming back from background to black screen

designatednerd:
Reported on GH as #852 as well. 

I was never able to reproduce this issue, but from the looks of the code the fact that there were two competing windows - the main window and the protection window - appears to be the root of the problem. 

I switched everything up to using a particular type of view controller and presenting/dismissing it from the application window's root view controller. Everything seems to be working, including after force-quitting the application, but I'd particularly appreciate a look from @Elland since I know he mentioned having the same issue as the reporter of #852.


designatednerd:
Moving to a different PR.

